PR #3, opened 2023-03-21
Added documentation

manojmalik20:
#### Summary
Added documentation


jespino:
I'm doing some changes in the configuration, and in the setup, but I think I can do the changes to the documentation too. I think this is great, so let's merge it, and then I can do the extra changes in the config :)

PR #17, opened 2023-03-23
Added unit test cases for message_hooks.go file

ayusht2810:
#### Summary
Added unit test cases for some functions of message_hooks.go file:
- `MessageWillBePosted`
- `ReactionHasBeenAdded`

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/16

### DIFF
--- a/server/message_hooks_test.go
+++ b/server/message_hooks_test.go
@@ -1,0 +1,177 @@
+package main
+
+import (
+	"errors"
+	"testing"
+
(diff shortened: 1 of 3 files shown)


jespino (approved):
Great work, I'm missing some checks around error logs, but I think that is fine because they are swallow by the "newTestPlugin" function. I think I going to remove the generic mock for `LogError` there, and only keep the `LogDebug`.

PR #47, opened 2023-04-03
Add unit test cases for plugin.go file

ayusht2810:
#### Summary
Added unit test cases for some functions of plugin.go file:

- `ConnectTeamsAppClient`
- `Start`
- `GeneratePluginSecrets`

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/43

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -167,9 +167,11 @@ func (p *Plugin) startSubscriptions(ctx context.Context) {
 	maxRetries := 20
 	for {
 		resp, _ := http.Post(p.GetURL()+"/changes?validationToken=test-alive", "text/html", bytes.NewReader([]byte{}))
-		resp.Body.Close()
-		if resp.StatusCode == 200 {
-			break
+		if resp != nil {
+			resp.Body.Close()
+			if resp.StatusCode == 200 {
+				break
+			}
 		}
 
 		counter++
--- a/server/plugin_test.go
+++ b/server/plugin_test.go
@@ -391,10 +391,8 @@ func TestConnectTeamsAppClient(t *testing.T) {
 			ExpectedError: "unable to connect to the app client",
 		},
 		{
-			Name: "ConnectTeamsAppClient: Valid",
-			SetupAPI: func(api *plugintest.API) {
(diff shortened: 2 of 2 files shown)


server/plugin_test.go
```diff
@@ -373,3 +372,111 @@ func TestSyncUsers(t *testing.T) {
...
+}
+
+func TestStart(t *testing.T) {
+	for _, test := range []struct {
+		Name        string
+		SetupAPI    func(*plugintest.API)
+		SetupStore  func(*storemocks.Store)
+		SetupClient func(*mocks.Client)
+	}{
+		{
```
jespino, line 425: Is there any reason to not have a success case here?

jespino (changes requested):
A couple of details, but otherwise looks good to me.

server/plugin_test.go
ayusht2810: a function `startSubscriptions` is called in this function which calls `p.clusterMutex.Lock()` which returns a `nil pointer dereference` error, due to which we are not able to add a success case here

jespino: probably because the cluster mutex that you are creating is an empty cluster mutex. You can create a fake cluster mutex with something like `cluster.NewMutex(p.API, clusterMutexKey)`. That way, it should work fine.

jespino (changes requested):
Another improvement request on 1 test.

server/plugin_test.go
ayusht2810: updated the test case

server/plugin.go
```diff
@@ -167,9 +167,11 @@ func (p *Plugin) startSubscriptions(ctx context.Context) {
 	maxRetries := 20
 	for {
 		resp, _ := http.Post(p.GetURL()+"/changes?validationToken=test-alive", "text/html", bytes.NewReader([]byte{}))
-		resp.Body.Close()
-		if resp.StatusCode == 200 {
-			break
+		if resp != nil {
```
jespino, line 170: I love it

jespino (approved):

PR #50, opened 2023-04-03
Add unit tests for crypt.go file

ayusht2810:
#### Summary
Add unit test cases for crypt.go file

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/49

### DIFF
--- a/server/store/crypt_test.go
+++ b/server/store/crypt_test.go
@@ -1,0 +1,89 @@
+package store
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPad(t *testing.T) {
+	assert := assert.New(t)
+	resp := pad(make([]byte, 3))
+	assert.Equal([]byte{0x0, 0x0, 0x0, 0xd, 0xd, 0xd, 0xd, 0xd, 0xd, 0xd, 0xd, 0xd, 0xd, 0xd, 0xd, 0xd}, resp)
+}
+
+func TestUnpad(t *testing.T) {
+	assert := assert.New(t)
+	resp, err := unpad(make([]byte, 1))
+	assert.Nil(err)
+	assert.Equal([]uint8([]byte{0x0}), resp)
+}
+
+func TestEncrypt(t *testing.T) {
+	for _, test := range []struct {
+		Name          string
+		Key           []byte
+		ExpectedError string
+	}{
+		{
+			Name:          "Encrypt: Invalid key",
+			Key:           make([]byte, 1),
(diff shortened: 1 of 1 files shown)


server/store/crypt_test.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func TestUnpad(t *testing.T) {
+	assert := assert.New(t)
+	resp, err := unpad(make([]byte, 1))
+	assert.Nil(err)
+	assert.Equal([]byte{0x0}, resp)
+}
+
+func TestEncrypt(t *testing.T) {
```
jespino, line 22: This is not testing the result.

Also I would like to get a test that proof the symmetry between the two functions something that verify that encrypts and decrypts works well together.

jespino (changes requested):
Still needed to verify the encrypt output.

Also needed a test that verify that encrypt and decrypt functions work well with each other.

server/store/crypt_test.go
ayusht2810: updated the test case. Let me know if any further changes need to be done

jespino (changes requested):
Suggested an improvement in the approach of testing it.

server/store/crypt_test.go
ayusht2810: @jespino updated the test case. Let me know if any further changes need to be done

server/store/crypt_test.go
```diff
@@ -57,3 +57,72 @@ func TestEncrypt(t *testing.T) {
...
+	} {
+		t.Run(test.Name, func(t *testing.T) {
+			assert := assert.New(t)
+			resp, err := encrypt(test.Key, "mockData")
+			if test.ExpectedError != "" {
+				assert.Contains(err.Error(), test.ExpectedError)
+				assert.Equal("", resp)
+			} else {
+				assert.Nil(err)
+				assert.NotEqual("", resp)
```
jespino, line 85: Here is one of my main concerns. This test is saying anything that the `encrypt` function returns that is not empty, is a valid result for encrypting "mockData" string. so if encrypts always returns "fake", it will pass. If I change the implementation of encrypt, as far as it return a non empty string, it will pass. The same for the decrypt method. Returning not empty string is not a valid enough check.

jespino (changes requested):
I would like to see explicit checks about the results of the encrypt and decrypt, not only "not empty"

server/store/crypt_test.go
jespino: ok, that makes sense.

PR #56, opened 2023-04-04
Add unit test cases for client.go file

ayusht2810:
#### Summary
Added unit test cases for some functions of client.go file:

- `convertToMessage`
- `GetResourceIds`

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/55

### DIFF
--- a/server/msteams/client_test.go
+++ b/server/msteams/client_test.go
@@ -20,65 +20,93 @@ func TestConvertToMessage(t *testing.T) {
 	attachmentContentType := "mockAttachmentContentType"
 	attachmentName := "mockAttachmentName"
 	attachmentURL := "mockAttachmentURL"
-	assert := assert.New(t)
-	resp := convertToMessage(&msgraph.ChatMessage{
-		From: &msgraph.IdentitySet{
-			User: &msgraph.Identity{
-				ID:          &teamsUserID,
-				DisplayName: &teamsUserDisplayName,
-			},
-		},
-		ReplyToID: &teamsReplyID,
-		Body: &msgraph.ItemBody{
-			Content: &content,
-		},
-		Subject:              &subject,
-		LastModifiedDateTime: &time.Time{},
-		Attachments: []msgraph.ChatMessageAttachment{
-			{
-				ContentType: &attachmentContentType,
-				Content:     &attachmentContent,
-				Name:        &attachmentName,
-				ContentURL:  &attachmentURL,
-			},
-		},
-		Reactions: []msgraph.ChatMessageReaction{
-			{
-				ReactionType: &reactionType,
-				User: &msgraph.IdentitySet{
+	for _, test := range []struct {
+		Name           string
+		ChatMessage    *msgraph.ChatMessage
+		ExpectedResult Message
+	}{
+		{
+			Name: "ConvertToMessage: With data filled",
+			ChatMessage: &msgraph.ChatMessage{
+				From: &msgraph.IdentitySet{
 					User: &msgraph.Identity{
-						ID: &teamsUserID,
+						ID:          &teamsUserID,
+						DisplayName: &teamsUserDisplayName,
+					},
+				},
+				ReplyToID: &teamsReplyID,
+				Body: &msgraph.ItemBody{
+					Content: &content,
+				},
+				Subject:              &subject,
+				LastModifiedDateTime: &time.Time{},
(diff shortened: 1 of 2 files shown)


server/msteams/client_test.go
```diff
@@ -0,0 +1,115 @@
...
+func TestGetResourceIds(t *testing.T) {
+	for _, test := range []struct {
+		Name           string
+		Resource       string
+		ExpectedResult ActivityIds
+	}{
+		{
+			Name:     "GetResourceIds: With prefix chats(",
+			Resource: "chats(chat123-ID/mockChannel456-ID",
+			ExpectedResult: ActivityIds{
```
jespino, line 93: This expected result is non-sense, the resource doesn't have that shape. If you are testing the happy path the input should be the right one. If you are testing a non-happy path, the error should be captured somehow. Probably the function should be failing somehow with that input.

server/msteams/client_test.go
```diff
@@ -0,0 +1,115 @@
...
+	teamsReplyID := "mockTeamsReplyID"
+	content := "mockContent"
+	subject := "mockSubject"
+	reactionType := "mockReactionType"
+	attachmentContent := "mockAttachmentContent"
+	attachmentContentType := "mockAttachmentContentType"
+	attachmentName := "mockAttachmentName"
+	attachmentURL := "mockAttachmentURL"
+	assert := assert.New(t)
+	resp := convertToMessage(&msgraph.ChatMessage{
```
jespino, line 24: I would like to see another testcase where the data passed is almost empty (a lot of nils) to see that the application doesn't panic and is following the other paths in the ifs.

server/msteams/client_test.go
```diff
@@ -0,0 +1,115 @@
...
+			Resource: "chats(chat123-ID/mockChannel456-ID",
+			ExpectedResult: ActivityIds{
+				ChatID:    "hat123-",
+				MessageID: "l456-",
+			},
+		},
+		{
+			Name:     "GetResourceIds: Without prefix chats(",
+			Resource: "mockTeam123-ID/mockChannel456-ID/mockMessage789-ID/mockReply910-ID",
+			ExpectedResult: ActivityIds{
```
jespino, line 101: Same here, the resource format is not the expected one

jespino: See here for more details: https://learn.microsoft.com/en-us/graph/teams-changenotifications-chatmessage#notification-payloads

jespino (changes requested):
Some improvements needed

server/msteams/client_test.go
```diff
@@ -0,0 +1,115 @@
...
+		ChatID:    "mockChatID",
+	}, *resp)
+}
+
+func TestGetResourceIds(t *testing.T) {
+	for _, test := range []struct {
+		Name           string
+		Resource       string
+		ExpectedResult ActivityIds
+	}{
```
jespino, line 89: I would like to have other cases here, like "big nonsense string", empty string, or very tiny string (for example 2 characters), or maybe a string that contains 2 `/` together to confuse a bit the "parser".

ayusht2810:
@jespino fixed review comments. Please re-review and let me know if anything else needs to be changed.

server/msteams/client_test.go
```diff
@@ -0,0 +1,161 @@
...
+			Name: "GetResourceIds: Empty resource",
+		},
+		{
+			Name:     "GetResourceIds: Resource with small length",
+			Resource: "ID",
+		},
+		{
+			Name:     "GetResourceIds: Resource with large length",
+			Resource: "very-long-teams-ID-with-very-long-chat-ID",
+			ExpectedResult: ActivityIds{
```
jespino, line 150: I think this expected result makes no sense, I would expect that this doesn't work, probably returning an empty result, like other cases. I prefer a failing tests that makes sense, than a not failing test that follows the current wrong behavior.

ayusht2810: @jespino  should I change the function implementation, as the current implementation will not return an empty result or should we remove this test case?

jespino: Yes, that is the idea. If the behavior is not correct, we should fix the behavior, not change the test. I'm ok if you make the changes for the test to fail and I can jump in later and fix the implementation. However you prefer. But I don't want "test that verify the current behavior", I want "test that verify the correct behavior".

jespino (approved):
LGTM

PR #58, opened 2023-04-05
Add unit test cases for store.go file

ayusht2810:
#### Summary
Added unit test cases for some functions of store.go file:

- `GetAvatarCache`
- `SetAvatarCache`
- `CheckEnabledTeamByTeamID`

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/57

### DIFF
--- a/server/store/store_test.go
+++ b/server/store/store_test.go
@@ -1,0 +1,148 @@
+package store
+
+import (
+	"testing"
+
+	"github.com/mattermost/mattermost-plugin-msteams-sync/server/testutils"
+	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/mattermost/mattermost-server/v6/plugin/plugintest"
+	"github.com/stretchr/testify/assert"
+)
+
+func setupTestStore(api *plugintest.API) (*SQLStore, *plugintest.API) {
+	store := &SQLStore{}
+	store.api = api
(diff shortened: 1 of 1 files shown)


server/store/store_test.go
```diff
@@ -0,0 +1,148 @@
...
+			ExpectedResult: true,
+		},
+		{
+			Name: "CheckEnabledTeamByTeamID: Unable to get the team",
+			SetupAPI: func(api *plugintest.API) {
+				api.On("GetTeam", "mockTeamID").Return(nil, testutils.GetInternalServerAppError("unable to get the team"))
+			},
+			EnabledTeams: func() []string {
+				return []string{"mockTeamsTeam"}
+			},
```
jespino, line 112: I would prefer to see here the `ExpecteResult` field, because even if that is working because the zero value of `bool` type, I think makes sense to clearly show the expectation there, because the expectation in this case has a clear meaning.

jespino: And of course the same for all the test cases here.

jespino (changes requested):
A tiny improvement, but otherwise looks good to me.

ayusht2810:
@jespino fixed the review comments. Please re-review

jespino (approved):
LGTM

PR #82, opened 2023-04-13
Create synthetic users on channel messages when it is needed

jespino:


jespino:
I still need to manually test this, because my local environment is not working.

jespino:
I tested it manually and it is working :)

PR #88, opened 2023-04-18
[MI-3007] Fixed the issue of multiple MM users connecting to a single Teams user

manojmalik20:
#### Summary
Modified the primary key in the users table to a composite primary key
Modified the unit tests of store.go to fix them
Added a new function DeleteUserInfo in the store
Modified the disconnect user and bot slash commands to delete the infos from DB
Modified the unit tests for command.go to fix them

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/87


jespino (approved):
LGTM. Thanks!

Issue #91, opened 2023-04-20
Update syncing of users on the basis of their emails


ayusht2810:
- Sync users of MS in Mattermost only if they are not already present in the Mattermost
- Create new users in Mattermost keeping their emails the same as that of MS

ayusht2810:
fixed in PR https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/92. closing this issue

PR #92, opened 2023-04-21
Update syncing of users on the basis of their emails

ayusht2810:
#### Summary
Updated the logic of syncing users from Microsoft Teams to Mattemost

#### Ticket Link
#91


server/msteams/client.go
```diff
@@ -1013,6 +1015,12 @@ func (tc *ClientImpl) ListUsers() ([]User, error) {
 			DisplayName: *u.DisplayName,
 			ID:          *u.ID,
 		}
+
+		if u.Mail != nil {
+			users[i].Mail = strings.ToLower(*u.Mail)
+		} else {
```
jespino, line 1021: I would check here `u.UserPrincipalName != nil` to avoid to panic in case there is no `UserPrincipalName`

server/plugin.go
```diff
@@ -395,3 +400,30 @@ func generateSecret() (string, error) {
 	s = s[:32]
 	return s, nil
 }
+
+func generateRandomPassword() string {
```
jespino, line 404: What is the reasoning behind using this `generateRandomPassword` instead of a random ID?

server/msteams/client.go
```diff
@@ -43,8 +43,10 @@ type Chat struct {
 }
 
 type User struct {
-	DisplayName string
-	ID          string
+	DisplayName       string
+	ID                string
+	Mail              string
+	UserPrincipalName string
```
jespino, line 49: You are not using this new field that you are created, I think is not needed.

jespino (changes requested):
The main problem with this PR is that I'm using the email field to match the users. I'm generating a fake email based on the userID at MSTeams, because that allows me to query the users easily. Also, I don't want to sent Mattermost emails to synthetic users.

What is exactly the problem that we are trying to solve here?

server/plugin.go
ayusht2810: on using this feature I was getting an error something like this `createUserOrGuest: Login failed because of invalid password.` in my logs. Password requires these validations whereas the `NewId()` method generates a string without any special characters and no surety on the addition of capital and small letters.
image

ayusht2810:
There were mainly two reasons behind this issue:

- If a user already exists on MS Teams and Mattermost both, we were creating a new user with a fake email leading to a user duplicacy.
- For resolving the issue https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/65, we were thinking of adding missing users in the mattermost channel and vice-versa, while linking the channels. To compare the users of teams and mattermost, so that we only add missing users we needed some common field between them, which can be email in this case.

jespino:
@ayusht2810 do you mind to merge the main branch into this branch?

ayusht2810:
@jespino fixed review comments and update some other code as well. Please review

jespino (approved):
LGTM, only missing the merge from main to use the new client sdk

ayusht2810:
@jespino resolved the conflicts

PR #97, opened 2023-04-25
[MI-3027] Fix issue 15 of handling lifecycle subscriptionRemoved event

ayusht2810:
#### Summary
Fixed issue of handling lifecycle "subscriptionRemoved" event

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/15


server/api.go
```diff
@@ -138,8 +138,17 @@ func (a *API) processLifecycle(w http.ResponseWriter, req *http.Request) {
...
+			if err != nil {
+				a.p.API.LogError("Unable to subscribe to channels", "error", err)
+			}
+
+			_, err = a.p.msteamsAppClient.SubscribeToChats(a.p.GetURL()+"/", a.p.configuration.WebhookSecret, !a.p.configuration.EvaluationAPI)
+			if err != nil {
+				a.p.API.LogError("Unable to subscribe to chats", "error", err)
+			}
 		}
-		// TODO: handle "missed" lifecycle event to resync
```
jespino, line 142: we are not handling yet the `missed` event, so I would prefer to keep this comment here to remember that.

jespino (changes requested):
Tiny change, but otherwise looks good to me.

server/api.go
ayusht2810: the resources we are using don't support `missed` lifecycle notification event
you can read more about it here: https://learn.microsoft.com/en-us/graph/webhooks-lifecycle?tabs=http#resources-that-support-lifecycle-notifications

ayusht2810:
@jespino replied to the comment

jespino (approved):
LGTM

Issue #89, opened 2023-04-19
Even after regenrating the webhook secret in plugin setting in MM, still the MM channel is in sync with the Teams.


AayushChaudhary0001:
If the user regenerates the webhook secret after connecting the user and linking the channel, the linked channel in MM is still found in Sync with Teams. All the messages delivered on MM are visible on Teams and vice versa.
Instead, the expected behaviour should be that all the channels linked should get unlinked after regenerating the webhook secret.

manojmalik20:
In the latest pull, this issue was not reproducible always. This was a bit flaky behaviour, so I changed the way the configuration values are being used and that resolved the issue.

manojmalik20:
There's a change in the expected behavior. When a webhook secret is changed, the subscriptions still keep working and they are also updated on MS Graph with the latest updated webhook secret.

PR #101, opened 2023-04-27
[MI-3030] Changed the way of using plugin's configuration settings to fix issue #89

manojmalik20:
#### Summary
Changed the way of using plugin's configuration settings everywhere

#### Ticket Link
#89


jespino (approved):
Good catch. Thanks!

PR #105, opened 2023-04-28
Adding the new monitoring system for subscriptions

jespino:
This PR introduces the concept of monitor system. The idea is the monitor system can grow in the future to monitor other health problems of the system and expose that to the user if it is needed.

The system implements 3 kind of checks, but only 1 is used for now. The checks ar for "global subscriptions", that are the current existing ones. Then we have also the "channel" subscriptions and the "user" subscriptions, that are going to be implemented in the future and can be monitor with this same system.


jespino:
@manojmalik20 do you mind to review this PR, I want to have a second pair of eyes on it.

server/msteams/client.go
```diff
@@ -463,7 +478,13 @@ func (tc *ClientImpl) SubscribeToChannels(baseURL string, webhookSecret string,
 	return tc.subscribe(baseURL, webhookSecret, resource, changeType)
 }
 
-func (tc *ClientImpl) SubscribeToChats(baseURL string, webhookSecret string, pay bool) (string, error) {
+func (tc *ClientImpl) SubscribeToChannel(teamID, channelID, baseURL string, webhookSecret string) (*Subscription, error) {
```
manojmalik20, line 481:
```suggestion
func (tc *ClientImpl) SubscribeToChannel(teamID, channelID, baseURL, webhookSecret string) (*Subscription, error) {
```

server/msteams/client.go
```diff
@@ -463,7 +478,13 @@ func (tc *ClientImpl) SubscribeToChannels(baseURL string, webhookSecret string,
...
+	resource := fmt.Sprintf("/teams/%s/channels/%s/messages", teamID, channelID)
+	changeType := "created,deleted,updated"
+	return tc.subscribe(baseURL, webhookSecret, resource, changeType)
+}
+
+func (tc *ClientImpl) SubscribeToChats(baseURL string, webhookSecret string, pay bool) (*Subscription, error) {
```
manojmalik20, line 487:
```suggestion
func (tc *ClientImpl) SubscribeToChats(baseURL, webhookSecret string, pay bool) (*Subscription, error) {
```

server/msteams/client.go
```diff
@@ -472,12 +493,29 @@ func (tc *ClientImpl) SubscribeToChats(baseURL string, webhookSecret string, pay
 	return tc.subscribe(baseURL, webhookSecret, resource, changeType)
 }
 
-func (tc *ClientImpl) RefreshSubscription(subscriptionID string) error {
+func (tc *ClientImpl) SubscribeToUserChats(userID, baseURL string, webhookSecret string, pay bool) (*Subscription, error) {
```
manojmalik20, line 496:
```suggestion
func (tc *ClientImpl) SubscribeToUserChats(userID, baseURL, webhookSecret string, pay bool) (*Subscription, error) {
```

server/msteams/client.go
```diff
@@ -463,7 +478,13 @@ func (tc *ClientImpl) SubscribeToChannels(baseURL string, webhookSecret string,
 	return tc.subscribe(baseURL, webhookSecret, resource, changeType)
 }
 
-func (tc *ClientImpl) SubscribeToChats(baseURL string, webhookSecret string, pay bool) (string, error) {
+func (tc *ClientImpl) SubscribeToChannel(teamID, channelID, baseURL string, webhookSecret string) (*Subscription, error) {
+	resource := fmt.Sprintf("/teams/%s/channels/%s/messages", teamID, channelID)
```
manojmalik20, line 482: Doesn't this resource require the license model?

server/msteams/client.go
```diff
@@ -472,12 +493,29 @@ func (tc *ClientImpl) SubscribeToChats(baseURL string, webhookSecret string, pay
 	return tc.subscribe(baseURL, webhookSecret, resource, changeType)
 }
 
-func (tc *ClientImpl) RefreshSubscription(subscriptionID string) error {
+func (tc *ClientImpl) SubscribeToUserChats(userID, baseURL string, webhookSecret string, pay bool) (*Subscription, error) {
+	resource := fmt.Sprintf("/users/%s/chats/getAllMessages", userID)
+	if pay {
+		resource = "chats/getAllMessages?model=B"
+	}
```
manojmalik20, line 500: This doesn't seem right. If we're using the metered API and the resource "user chat change notifications" doesn't support license modeling, then we shouldn't change the resource to "chats/getAllMessages" I think.

server/store/store.go
```diff
@@ -153,6 +162,10 @@ func (s *SQLStore) addColumn(tableName, columnName, columnDefinition string) err
 }
 
 func (s *SQLStore) Init() error {
+	if err := s.createTable("msteamssync_subscriptions", "subscriptionID VARCHAR(255) PRIMARY KEY, type VARCHAR(255), msTeamsTeamID VARCHAR(255), msTeamsChannelID VARCHAR(255), msTeamsUserID VARCHAR(255), secret VARCHAR(255), expiresOn BIGINT"); err != nil {
```
manojmalik20, line 165: Has this been tested with both Postgres and MySQL?

server/store/store_test.go
```diff
@@ -541,3 +549,245 @@ func testGetTokenForMSTeamsUserForInvalidID(t *testing.T, store *SQLStore, _ *pl
...
+		err := store.SaveGlobalSubscription(storemodels.GlobalSubscription{SubscriptionID: "test", Type: "allChannels", Secret: "secret", ExpiresOn: time.Now().Add(100 * time.Minute)})
+		require.NoError(t, err)
+		defer func() { _ = store.DeleteSubscription("test") }()
+
+		subscriptions, err := store.ListGlobalSubscriptionsToCheck()
+		require.NoError(t, err)
+		assert.Empty(t, subscriptions)
+	})
+
+	t.Run("one-subscription-almost-expired-and-other-dont", func(t *testing.T) {
```
manojmalik20, line 570:
```suggestion
	t.Run("one-subscription-almost-expired-and-other-didn't", func(t *testing.T) {
```

server/store/store_test.go
```diff
@@ -541,3 +549,245 @@ func testGetTokenForMSTeamsUserForInvalidID(t *testing.T, store *SQLStore, _ *pl
...
+			ids = append(ids, s.SubscriptionID)
+		}
+		assert.NotContains(t, ids, "test1")
+		assert.NotContains(t, ids, "test2")
+		assert.NotContains(t, ids, "test3")
+		assert.Contains(t, ids, "test4")
+		assert.Contains(t, ids, "test5")
+		assert.Contains(t, ids, "test6")
+	})
+}
```
manojmalik20, line 704:
```suggestion
}

```

server/monitor/subscriptions.go
```diff
@@ -0,0 +1,132 @@
...
+	for _, subscription := range subscriptions {
+		if time.Until(subscription.ExpiresOn) < (15 * time.Second) {
+			err := m.recreateGlobalSubscription(subscription.SubscriptionID, subscription.Type, subscription.Secret)
+			if err != nil {
+				m.api.LogError("Unable to recreate global subscription properly", "error", err)
+			}
+		} else {
+			err := m.refreshSubscription(subscription.SubscriptionID)
+			if err != nil {
+				m.api.LogError("Unable to refresh global subscription properly", "error", err)
```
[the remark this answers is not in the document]
jespino: It is not strictly needed because the LogError will handle that anyway. But I think we don't have a clear policy around that. Maybe we should have one.

server/msteams/client.go
jespino: Nope, actually, it is not supported, AFAIK

server/msteams/client.go
jespino: good catch. Thanks

server/store/store.go
jespino: I'm running it through the tests, so the table creation should work in both or fail in the CI.

manojmalik20 (approved):

PR #107, opened 2023-05-01
Use Site URL for frontend operations

gijsdb:
#### Summary
This PR sets the site URL for frontend operations

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/104

### DIFF
--- a/webapp/src/selectors/index.tsx
+++ b/webapp/src/selectors/index.tsx
@@ -1,6 +1,6 @@
 import {getConfig} from 'mattermost-redux/selectors/entities/general';
 
-export const getServerRoute = (state) => {
+export const getServerRoute = (state: any) => {
     const config = getConfig(state);
 
     let basePath = '';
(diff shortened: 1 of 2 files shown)


gijsdb:
@mickmister I have mimicked the example you provided. Let me know if that's what you envisioned.

mickmister (approved):
LGTM :+1:

webapp/src/selectors/index.tsx
```diff
@@ -0,0 +1,16 @@
+import {getConfig} from 'mattermost-redux/selectors/entities/general';
+
+export const getServerRoute = (state) => {
```
jespino, line 3: This is failing in the linter because you need to pass a type for the state, which can be `any`, but you have to provide that explicitly.

gijsdb: @jespino My apologies I should have caught that. Pushed a fix :)

jespino:
Thanks @gijsdb! :tada:

PR #108, opened 2023-05-02
Fix issue 65 of handling missing users in the channel

ayusht2810:
#### Summary

- Add missing users in Mattermost and MS Teams while linking the channels.
- Handle channel join/leave events on MS Teams whenever a user is removed from the Mattermost channel.
- Handle channel join/leave events on Mattermost whenever a user is removed from the MS Teams channel.

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/65

[screen-capture (26).webm](https://user-images.githubusercontent.com/100013900/235637218-67cd75e7-1873-448c-aadc-ff61059b1975.webm)


jespino:
@ayusht2810 do you mind merging the last main branch into this PR?

ayusht2810:
@jespino updated the PR

server/command.go
```diff
@@ -181,6 +187,112 @@ func (p *Plugin) executeLinkCommand(args *model.CommandArgs, parameters []string
 		return p.cmdError(args.UserId, args.ChannelId, "Unable to create new link.")
 	}
 
+	if p.getConfiguration().SyncUsers > 0 {
+		batchSize := 50
+		go func() {
```
jespino, line 192: For my taste this section of the function is complex enough to have its own function, actually, even if you move that to another function, it ends up being complex enough to split it again. Lets start splitting it in some chunks that makes sense, and lets see if that improve the readability or if from there emerges any interesting abstraction.

jespino (changes requested):
In general I like the code, except for the complex goroutine call for the syncing. Other than that, I think it looks ok.

Anyway, I need to think about it before merging it, because we are introducing the maintenance of join/leaving channels logic and that can have other implications, I need to understand better the problem, the solution and the consequences before merging this.

ayusht2810:
@jespino updated the code. please re-review

server/command.go
```diff
@@ -320,3 +339,125 @@ func (p *Plugin) executeDisconnectBotCommand(args *model.CommandArgs) (*model.Co
...
+		mmUsersMap[user.Email] = user
+	}
+
+	return mmUsers, mmUsersMap
+}
+
+func (p *Plugin) addMattermostUser(wg *sync.WaitGroup, teamID, channelID string, msChannelMembers []msteams.User, mmUsersMap map[string]*model.User, batchSize int) {
+	for i := 0; i < (len(msChannelMembers)/batchSize)+1; i++ {
+		wg.Add(1)
+		go func(i int) {
```
jespino, line 384: We are creating here an unbounded number of goroutines. Not necessarily a lot, but still, something without any control here. A lot of goroutines hammering the database without control. We have to find a better way to improve that. Probably something that give us the list of users that needs to be created, the list of users that are not in the team, and the list of users that are in the team but not in the channel. And then do the operations on that users, instead of hammering the API all the time with unneeded requests.

server/command.go
```diff
@@ -320,3 +339,125 @@ func (p *Plugin) executeDisconnectBotCommand(args *model.CommandArgs) (*model.Co
...
+				}
+			}
+		}(i)
+	}
+}
+
+func (p *Plugin) addMSTeamsUser(wg *sync.WaitGroup, client msteams.Client, teamID, channelID string, mmUsers []*model.User, msUsersMap map[string]msteams.User, batchSize int) {
+	for i := 0; i < (len(mmUsers)/batchSize)+1; i++ {
+		wg.Add(1)
+		go func(i int) {
```
jespino, line 436: Same here, unbounded number of goroutines :(

jespino: I this case is even worse, we hare hammering the msteams API like crazy. In this case we should do something like batching, at least.

jespino (changes requested):
Still, I don't have a clear vision about the channel memberships, but anyway there are things that needs to be fix before this gets merged.

lieut-data:
Closing this out for now and may revisit the underlying problem in a different way.

PR #109, opened 2023-05-02
[MI-3041] Created a slash command for showing the linked channels

manojmalik20:
#### Summary
- Added a new store function for getting all the links from the database with a max limit
- Created a new slash command handler for showing a table with all the linked channels
- Added the logic to show autocomplete for the new slash command to only system admins
- Added the logic to fetch connected Teams team and channel info in multiple goroutines

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/96

#### Screenshots
image


server/command.go
```diff
@@ -240,6 +249,168 @@ func (p *Plugin) executeShowCommand(args *model.CommandArgs) (*model.CommandResp
...
+	go p.SendLinksWithDetails(args.UserId, args.ChannelId, links)
+	return &model.CommandResponse{}, nil
+}
+
+func (p *Plugin) SendLinksWithDetails(userID, channelID string, links []*storemodels.ChannelLink) {
+	var sb strings.Builder
+	sb.WriteString("| Mattermost Team | Mattermost Channel | MS Teams Team | MS Teams Channel | \n| :------|:--------|:-------|:-----------|")
+	gotErrors := false
+	wg := sync.WaitGroup{}
+	batchSize := 10
```
Kshitij-Katiyar, line 272: Do we need an exported const for this as this batchSize in future this could be used in some other place ?

server/command.go
```diff
@@ -240,6 +249,168 @@ func (p *Plugin) executeShowCommand(args *model.CommandArgs) (*model.CommandResp
 	return &model.CommandResponse{}, nil
 }
 
+func (p *Plugin) executeShowLinksCommand(args *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
+	links, err := p.store.GetLinks()
+	if err != nil {
+		return p.cmdError(args.UserId, args.ChannelId, "Something went wrong.")
```
avas27JTG, line 255: should not we also log actual `err`

server/command.go
```diff
@@ -240,6 +249,168 @@ func (p *Plugin) executeShowCommand(args *model.CommandArgs) (*model.CommandResp
...
+			msTeamsChannelIDsVsNames[link.MSTeamsChannel],
+		)
+
+		if row != "\n|||||" {
+			sb.WriteString(row)
+		}
+	}
+
+	if gotErrors {
+		sb.WriteString("\nThere were some errors while fetching information about some links. Please check the server logs.")
```
avas27JTG, line 350:
```suggestion
		sb.WriteString("\nThere were some errors while fetching information. Please check the server logs.")
```

server/command.go
```diff
@@ -240,6 +249,168 @@ func (p *Plugin) executeShowCommand(args *model.CommandArgs) (*model.CommandResp
...
+
+	for _, msTeamsTeam := range msTeamsTeams {
+		msTeamsTeamIDsVsNames[msTeamsTeam.ID] = msTeamsTeam.DisplayName
+	}
+
+	return false
+}
+
+func (p *Plugin) GetMSTeamsChannelDetailsForAllTeams(msTeamsTeamIDsVsChannelsQuery, msTeamsChannelIDsVsNames map[string]string, wg *sync.WaitGroup) bool {
+	gotErrors := false
```
avas27JTG, line 384:
```suggestion
	errorsFound := false
```
this sound more technical to me if changing, change at all such places

server/command.go
```diff
@@ -240,6 +249,168 @@ func (p *Plugin) executeShowCommand(args *model.CommandArgs) (*model.CommandResp
...
+		}
+
+		go func(teamID, channelsQuery string) {
+			if wg != nil {
+				defer wg.Done()
+			}
+
+			channels, err := p.msteamsAppClient.GetChannelsInATeam(teamID, channelsQuery+")")
+			if err != nil {
+				p.API.LogDebug("Unable to get the MS Teams channel information for a team", "TeamID", teamID, "Error", err.Error())
```
avas27JTG, line 397:
```suggestion
				p.API.LogDebug("Unable to get the MS Teams channel information for the team", "TeamID", teamID, "Error", err.Error())
```
`teamID` is specific here

server/command.go
```diff
@@ -240,6 +249,168 @@ func (p *Plugin) executeShowCommand(args *model.CommandArgs) (*model.CommandResp
...
+	}
+
+	p.sendBotEphemeralPost(args.UserId, args.ChannelId, "Please wait while your request is being processed.")
+	go p.SendLinksWithDetails(args.UserId, args.ChannelId, links)
+	return &model.CommandResponse{}, nil
+}
+
+func (p *Plugin) SendLinksWithDetails(userID, channelID string, links []*storemodels.ChannelLink) {
+	var sb strings.Builder
+	sb.WriteString("| Mattermost Team | Mattermost Channel | MS Teams Team | MS Teams Channel | \n| :------|:--------|:-------|:-----------|")
```
avas27JTG, line 269: create a constant

server/command_test.go
```diff
@@ -238,6 +239,132 @@ func TestExecuteShowCommand(t *testing.T) {
...
+				api.On("SendEphemeralPost", testutils.GetUserID(), &model.Post{
+					UserId:    "bot-user-id",
+					ChannelId: testutils.GetChannelID(),
+					Message:   "Please wait while your request is being processed.",
+				}).Return(testutils.GetPost(testutils.GetChannelID(), testutils.GetUserID())).Once()
+
+				api.On("SendEphemeralPost", testutils.GetUserID(), &model.Post{
+					UserId:    "bot-user-id",
+					ChannelId: testutils.GetChannelID(),
+					Message:   "| Mattermost Team | Mattermost Channel | MS Teams Team | MS Teams Channel | \n| :------|:--------|:-------|:-----------|\n|Test MM team|Test MM channel|Test MS team|Test MS channel|",
```
avas27JTG, line 272: can create a constant for this table header layout so that same can be used at all such places also if in the future if we need to modify anything will need to change at only one place

server/command_test.go
```diff
@@ -238,6 +239,132 @@ func TestExecuteShowCommand(t *testing.T) {
...
+				ChannelId: testutils.GetChannelID(),
+			},
+			setupAPI: func(api *plugintest.API) {
+				api.On("GetTeam", testutils.GetTeamID()).Return(&model.Team{DisplayName: "Test MM team"}, nil).Times(1)
+				api.On("GetChannel", testutils.GetChannelID()).Return(&model.Channel{DisplayName: "Test MM channel"}, nil).Times(1)
+
+				api.On("SendEphemeralPost", testutils.GetUserID(), &model.Post{
+					UserId:    "bot-user-id",
+					ChannelId: testutils.GetChannelID(),
+					Message:   "Please wait while your request is being processed.",
```
Kshitij-Katiyar, line 266: Do we have a constant for this

server/command_test.go
```diff
@@ -238,6 +239,132 @@ func TestExecuteShowCommand(t *testing.T) {
...
+		},
+	} {
+		t.Run(testCase.description, func(t *testing.T) {
+			testCase.setupAPI(mockAPI)
+			p.SetAPI(mockAPI)
+			defer mockAPI.AssertExpectations(t)
+
+			testCase.setupStore(p.store.(*mockStore.Store))
+			testCase.setupClient(p.msteamsAppClient.(*mockClient.Client))
+			_, _ = p.executeShowLinksCommand(testCase.args)
```
Kshitij-Katiyar, line 362: check this error as well

manojmalik20: There's no use because we are already checking the error returned with the function "SendEphemeralPost" and this function always returns the error as "nil".

server/command.go
manojmalik20: I don't think so.

server/command_test.go
manojmalik20: I don't think there's any need to make this constant as this will not be used anywhere else except the unit tests.

Kshitij-Katiyar (approved):

server/command_test.go
avas27JTG: I mean creating this part as constant `"| Mattermost Team | Mattermost Channel | MS Teams Team | MS Teams Channel | \n| :------|:--------|:-------|:-----------|\n"` so that we can concatenate this to the rest of table content wherever needed like test cases and the ephemeral post

avas27JTG (approved):

server/command_test.go
manojmalik20: I understood you. This string will be used only in this function and only once, that's why I don't think there's a need to make this constant. For unit tests also, I don't think we should use a constant. The reason is whenever someone changes this string, the unit tests will fail and I want them to fail so that we can be aware of the changes. If we use a constant, then this will not be possible.

server/command.go
```diff
@@ -240,6 +249,173 @@ func (p *Plugin) executeShowCommand(args *model.CommandArgs) (*model.CommandResp
...
+	batchSize := 10
+	mattermostChannelIDsVsNames := make(map[string]string)
+	mattermostTeamIDsVsNames := make(map[string]string)
+
+	msTeamsTeamIDsVsNames := make(map[string]string)
+	msTeamsChannelIDsVsNames := make(map[string]string)
+	msTeamsTeamIDsVsChannelsQuery := make(map[string]string)
+
+	// Process all the links in batches using goroutines
+	for idx := 0; idx < len(links); idx += batchSize {
```
jespino, line 286: This is generating 1 goroutine per batch of 10 elements, that means that if you have 10.000 links, you are spawing 1000 goroutines that are going to hammer our database with 20 requests each. That means, 20.000 DB request in parallel. That is crazy. I would try to move this directly to the store, creating a new store method like "LinksWithNames", that returns a new links structures that contains the channel name and team name using a database Join. We can even do that for all the regular queries of links, probably is not going to have a performance impact, and we can change it later if that is the case.

That way we can get rid of all the store queries here, and the extra complexity around goroutines and syncrhonization.

jespino (changes requested):
Still some improvements that needs to be done.

server/command.go
```diff
@@ -274,20 +275,18 @@ func (p *Plugin) SendLinksWithDetails(userID, channelID string, links []*storemo
...
-	mattermostTeamIDsVsNames := make(map[string]string)
+	batchSize := math.Sqrt(float64(len(links)))
 
 	msTeamsTeamIDsVsNames := make(map[string]string)
 	msTeamsChannelIDsVsNames := make(map[string]string)
 	msTeamsTeamIDsVsChannelsQuery := make(map[string]string)
 
 	// Process all the links in batches using goroutines
-	for idx := 0; idx < len(links); idx += batchSize {
+	for idx := 0; idx < len(links); idx += int(batchSize) {
```
jespino, line 302: This is still an unbounded number of goroutines. But I guess being limited to this places is good enough.

jespino: Actually... I'm thinking. There is no longer IO inside this loop, I don't think we need batching or goroutines here. I would get rid of all that unneeded complexity and do it in a single loop.

jespino (changes requested):
Another iteration on the simplification of this, but other than that looks good to me.

server/command.go
```diff
@@ -292,56 +291,44 @@ func (p *Plugin) SendLinksWithDetails(userID, channelID string, links []*storemo
 	sb.WriteString("| Mattermost Team | Mattermost Channel | MS Teams Team | MS Teams Channel | \n| :------|:--------|:-------|:-----------|")
 	errorsFound := false
 	wg := sync.WaitGroup{}
-	batchSize := math.Max(math.Sqrt(float64(len(links))), 1000)
 
-	var msTeamsTeamIDsVsNames sync.Map
+	msTeamsTeamIDsVsNames := make(map[string]string)
 	var msTeamsChannelIDsVsNames sync.Map
```
jespino, line 296: Also we no longer need here the `sync.Map`, it is going to be simpler and more efficient to use a regular map.

jespino (changes requested):
Another tiny change needed, but we are almost there.

jespino (approved):
LGTM

PR #111, opened 2023-05-08
[MI-3058] Fix issue 95 of syncing mentions in the channels and chat

ayusht2810:
#### Summary
Fixed the issue of syncing user mentions and special mentions from MS teams to Mattermost and from Mattermost to MS teams.

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/95


server/msteams/client.go
```diff
@@ -1120,3 +1164,99 @@ func GetAuthURL(redirectURL string, tenantID string, clientID string, clientSecr
...
+				conversation.SetId(&channelID)
+			}
+
+			conversation.SetDisplayName(&mentionedText)
+
+			conversationIdentityType := models.CHANNEL_TEAMWORKCONVERSATIONIDENTITYTYPE
+			conversation.SetConversationIdentityType(&conversationIdentityType)
+			mentioned.SetConversation(conversation)
+		} else {
+			mmUser, err := tc.plugin.GetUserByUsername(username)
```
jespino, line 1222: I have concerns about having to pass the whole API and the whole store to address this. That leads me to think that  it is not responsability of the client calculate the list of mentions. Maybe it should be done outside somehow.

jespino (changes requested):
The main problem that i see in this code is we are introducing a big coupling here unless you have a clear vision that this should be part of the client, I would move that outside, probably as part of the package that calls it.

If you feel really strong about having it here, at least, I would encapsulate that in a function that you pass to the client, instead of passing the whole API and the whole store.

ayusht2810:
@jespino fixed the review comment. Please re-review

jespino (approved):
LGTM

PR #122, opened 2023-05-16
[MI-3084] Fix issue 116 of long usernames causing error in syncing users

ayusht2810:
#### Summary
Fixed issue 116 of long usernames causing errors while syncing users by truncating the username to a certain length.

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/116


ayusht2810:
@jespino we have truncated the slug of the display name of the MS Teams user to 27 characters. Then we added `_` and then the MS Teams user ID which is 36 characters which sum up to 64 characters. Can you please elaborate the issue in this approach?

jespino:
@ayusht2810 the problem is, we no longer wants to have long usernames, what means, we no longer wand to append the user ID at the end of the username, what automatically implies that we need a way to ensure the user does not collide with any existing user in the system. That means, if the user with that username already exists, we generate a new username with the same previous username, but with a "-1", "-2", "-3" suffix.

For example, now "Jesus Espino" with UserID "xxxxxxxxxxxx", would create an account "jesus-espino-xxxxxxxxxxxx", what I want is to change that to create something like "msteams:jesus-espino", if "msteams:jesus-espino" already exists, create a "msteams:jesus-espino-1", if that exists, "msteams:jesus-espino-2" and so on.

server/handlers/getters.go
```diff
@@ -125,8 +124,7 @@ func (ah *ActivityHandler) getOrCreateSyntheticUser(userID, displayName string)
 		memberUUID := uuid.Parse(userID)
 		encoding := base32.NewEncoding("ybndrfg8ejkmcpqxot1uwisza345h769").WithPadding(base32.NoPadding)
 		shortUserID := encoding.EncodeToString(memberUUID)
-		msDisplayName := slug.Make(userDisplayName)
-		username := "msteams:" + msDisplayName
+		username := "msteams_" + slug.Make(userDisplayName)
```
jespino, line 127: One cool think about the remote users (what we are using for synthetic users) is that we have the possibility of adding `:`, specifically for this cases, so I would use `msteams:` instead of `msteams_`.

server/plugin.go
```diff
@@ -412,37 +412,26 @@ func (p *Plugin) syncUsers() {
...
-			if appErr != nil {
-				if strings.Contains(appErr.Error(), "account with that username already exists") {
-					value, sErr := p.API.KVGet(msDisplayName)
-					if sErr != nil {
-						p.API.LogError("Unable to get value from KV store", "error", sErr.Error())
+			var newUser *model.User
+			for {
+				newUser, appErr = p.API.CreateUser(newMMUser)
+				if appErr != nil {
+					newUsername := p.CheckAndGetUsername(appErr, newMMUser.Username, userSuffixID)
```
jespino, line 419: This approach of passing an error to handle it, I don't think is the right approach, I think here I prefer a bit of repetition than this kind of encapsulation.

I would copy the content of the function directly here in the places where we are calling this function. If you want something to encapsulate, I would suggest to encapsulate something like `ensureCreateUserWithSuffix` or something like that. That returns the new user and an error in case the reason to fail is not the duplicated user.

You are swallowing here any error reason without failing. For example, my expected behavior is, if I can't create a user because, for example, a temporary problem accessing the database, I don't want you to create a different user with the next suffix available, I want you to fail and give that error.

jespino:
@ayusht2810 reviewed again, do the changes, please

server/handlers/getters.go
ayusht2810: image
Adding `:` to the username was not allowed through UI. Also tagging users with `:` in username were not getting tagged. So updated `:` to `_`.
image

ayusht2810:
@jespino removed the function and replied to the comment. please re-review

server/handlers/getters.go
jespino: It should be a bug, but, yes, lets follow your original approach here, we can decide to change it later. @wiggin77 Do you agree that this is a bug in mattermost side, right? Not being able to mention remote users with `:` in their username, or is on purpose?

server/message_hooks.go
```diff
@@ -34,11 +34,7 @@ func (p *Plugin) MessageWillBePosted(_ *plugin.Context, post *model.Post) (*mode
 					return post, ""
 				}
 
-				if user.RemoteId == nil {
-					continue
-				}
-
-				if isMSTeamsUser(*user.RemoteId, user.Username) {
+				if user.RemoteId != nil && strings.Contains(user.Username, "msteams") {
```
jespino, line 37: This is slightly imprecise, I would use `strings.HasPrefix` instead.

jespino (approved):
LGTM, I tiny comment but we can fix it after the merge

server/handlers/getters.go
wiggin77: I'm checking this now, but I suspect it's because the system expects `username:remotename`

wiggin77: remote-user-mention

wiggin77: @jespino mentions for users with a `username:remotename` are working. See screen above.

server/message_hooks.go
wiggin77: IMO you should be looking for `msteams:` in case an account gets created with `msteams` in the name.  Users cannot create usernames with colons so it is safe to assume `msteams:` won't appear.  Also, it was previously decided by PM that username would appear before remote name.

wiggin77: cc @jespino

PR #126, opened 2023-05-18
[MI-3092] Fixed the issue of Excel files not being sent from MM to MS Teams

manojmalik20:
#### Summary
Fixed the issue of Excel files not being sent from Mattemost to MS Teams

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/121

### DIFF
--- a/server/msteams/client.go
+++ b/server/msteams/client.go
@@ -22,6 +22,7 @@ import (
 	"github.com/microsoftgraph/msgraph-beta-sdk-go/drives"
 	"github.com/microsoftgraph/msgraph-beta-sdk-go/groups"
 	"github.com/microsoftgraph/msgraph-beta-sdk-go/models"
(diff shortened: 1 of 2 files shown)


server/msteams/client.go
```diff
@@ -296,6 +335,23 @@ func (tc *ClientImpl) SendMessageWithAttachments(teamID, channelID, parentID, me
 		attachment := models.NewChatMessageAttachment()
 		attachment.SetId(&att.ID)
 		attachment.SetContentType(&contentType)
+
+		fileNameAndExtension := strings.Split(att.Name, ".")
```
jespino, line 339: Instead of this I would use directly this: https://pkg.go.dev/path/filepath#Ext

jespino (changes requested):
Some changes requested

jespino (approved):
LGTM. Anyway you are including not related changes in this PR. I prefer to separate that kind of things whenever it is possible.

PR #140, opened 2023-05-23
[MI-3100] Fix issue 132 of mentions not working properly for real use…

ayusht2810:
…rs who have not connected even once

#### Summary
Fixed issue 132 of mentions not working properly for real users who have not connected even once.

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/132

### DIFF
--- a/server/handlers/converters.go
+++ b/server/handlers/converters.go
@@ -67,16 +67,32 @@ func (ah *ActivityHandler) handleMentions(msg *msteams.Message) string {
 	for _, mention := range msg.Mentions {
 		mmMention := ""
 		if mention.UserID != "" {
+			var mmUser *model.User
+			var getErr *model.AppError
 			mmUserID, err := ah.plugin.GetStore().TeamsToMattermostUserID(mention.UserID)
 			if err != nil {
-				ah.plugin.GetAPI().LogDebug("Unable to get mm UserID", "Error", err.Error())
-				continue
-			}
+				msUser, mErr := ah.plugin.GetClientForApp().GetUser(mention.UserID)
+				if mErr != nil {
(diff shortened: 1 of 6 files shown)


server/handlers/converters.go
```diff
@@ -67,16 +67,32 @@ func (ah *ActivityHandler) handleMentions(msg *msteams.Message) string {
...
 			if err != nil {
-				ah.plugin.GetAPI().LogDebug("Unable to get mm UserID", "Error", err.Error())
-				continue
-			}
-
-			mmUser, getErr := ah.plugin.GetAPI().GetUser(mmUserID)
-			if getErr != nil {
-				ah.plugin.GetAPI().LogDebug("Unable to get mm user details", "Error", getErr.Error())
-				continue
+				msUser, mErr := ah.plugin.GetClientForApp().GetUser(mention.UserID)
```
jespino, line 74: If we are going to identify the user by the email, we should have a way to store that email information directly in the local store and use that whenever is needed. If not this is going to trigger a lot of API calls that are not needed. In the sample places where we store the MSTeams User ID we should store the MSTeams Email, and the add another method to get the User Email based on the mattermost userID, for example.

jespino (changes requested):
We need to rethink it a bit. We need to store the email because we are using it now as an ID.

server/handlers/converters.go
ayusht2810: @jespino Storing the email information in the store will not help in reducing the API calls as these are the cases when the user details are not present in the store itself. So even if we store the email information in the store, we cannot get user details from the store.

lieut-data:
Let's revisit this as we evaluate during bug bashes.

PR #141, opened 2023-05-24
[MI-3113] Fix email case sensitivity issue

ayusht2810:
#### Summary
Fix the issue of the user not being able to connect due to an email case sensitivity issue

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -350,7 +350,7 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if mmUser.Id != a.p.GetBotUserID() && strings.ToLower(msteamsUser.Mail) != mmUser.Email {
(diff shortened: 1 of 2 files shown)


server/api.go
```diff
@@ -350,7 +350,7 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if mmUser.Id != a.p.GetBotUserID() && msteamsUser.Mail != mmUser.Email {
```
jespino, line 353: This change shouldn't be needed because the client layer should be taking care of that entirely.

jespino (changes requested):
A tiny change needed, but otherwise looks good to me.

ayusht2810:
@jespino updated the PR. Please re-review

jespino (approved):
LGTM

PR #144, opened 2023-05-25
[MI-3123] Add and update logs in the sync user function

ayusht2810:
#### Summary
Add and update logs present in the sync user function.

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -374,14 +374,15 @@ func (p *Plugin) syncUsersPeriodically(ctx context.Context, minutes int) {
 }
 
 func (p *Plugin) syncUsers() {
+	p.API.LogDebug("Starting sync user job")
 	msUsers, err := p.msteamsAppClient.ListUsers()
 	if err != nil {
-		p.API.LogError("Unable to sync users", "error", err)
+		p.API.LogError("Unable to list MS Teams users during sync user job", "error", err.Error())
 		return
 	}
 	mmUsers, appErr := p.API.GetUsers(&model.UserGetOptions{Active: true, Page: 0, PerPage: math.MaxInt32})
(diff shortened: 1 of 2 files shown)


server/plugin.go
```diff
@@ -422,7 +425,7 @@ func (p *Plugin) syncUsers() {
 						continue
 					}
 
-					p.API.LogError("Unable to sync user", "error", appErr.Error())
+					p.API.LogError("Unable to create new mm user during sync job", "email", msUser.Mail, "error", appErr.Error())
```
manojmalik20, line 428:
```suggestion
					p.API.LogError("Unable to create new MM user during sync job", "email", msUser.Mail, "error", appErr.Error())
```

manojmalik20 (approved):

server/plugin_test.go
```diff
@@ -347,7 +347,7 @@ func TestSyncUsers(t *testing.T) {
 		{
 			Name: "SyncUsers: Unable to store the user info",
 			SetupAPI: func(api *plugintest.API) {
-				api.On("LogError", "Unable to sync user", "error", mock.Anything).Times(1)
+				api.On("LogError", "Unable to set user info during sync user job", "email", "test@test.com", "error", mock.Anything).Times(1)
```
manojmalik20, line 350: Ideally, we shouldn't use mock.Anything here because we can control the error being logged through our mocks.

manojmalik20 (approved):

server/plugin_test.go
ayusht2810: we need to add `: , unable to store the user info` something like this in mock. I think it is fine in this way

jespino (approved):

Issue #152, opened 2023-05-30
The plugin is giving "Payment required" error on updating a post


manojmalik20:
Whenver we update a post in Mattermost, the plugin tries to update the corresponding post in MS Teams, but we get the following error while doing that - 
```
Error updating the post                       caller="app/plugin_api.go:979" plugin_id=com.mattermost.msteams-sync error="code: PaymentRequired, message: Evaluation mode capacity has been exceeded. Use a valid billing model. Visit https://docs.microsoft.com/en-us/graph/teams-licenses for more details."
```
This was tested on Mattermost version 7.7.4 and the latest `main` branch of the plugin.

PR #154, opened 2023-05-31
[MI-3132] Fixed the issue of infinite loop when editing messages and a potential bug related to no. of channel members

manojmalik20:
#### Summary
- Added the logic to continue the flow if we get PaymentRequired error from Graph API when updating a message
- Increased the no. of channel members that we are fetching from the Mattermost API
- Modified the unit tests

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/151

### DIFF
--- a/server/message_hooks.go
+++ b/server/message_hooks.go
@@ -587,6 +587,8 @@ func (p *Plugin) Update(teamID, channelID string, user *model.User, newPost, old
 
 	if err = client.UpdateMessage(teamID, channelID, parentID, postInfo.MSTeamsID, content, mentions); err != nil {
 		p.API.LogWarn("Error updating the post", "error", err)
+		// If the error is regarding payment required for metered APIs, ignore it and continue because
+		// the post is updated regardless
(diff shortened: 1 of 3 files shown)


server/message_hooks.go
```diff
@@ -586,7 +587,9 @@ func (p *Plugin) Update(teamID, channelID string, user *model.User, newPost, old
 
 	if err = client.UpdateMessage(teamID, channelID, parentID, postInfo.MSTeamsID, content, mentions); err != nil {
 		p.API.LogWarn("Error updating the post", "error", err)
-		return err
+		if !strings.Contains(err.Error(), "code: PaymentRequired") {
```
jespino, line 590: This looks like "masking" the error, instead of fixing the error. When this happens and is safe to ignore it? Should we at least log why this is happening?

ayusht2810 (approved):

server/message_hooks.go
manojmalik20: There's a separate issue created #152 for exploring this error. Here, we are ignoring it because the post is still getting updated in spite of the error.

PR #159, opened 2023-06-02
Adding certificate subscription support

jespino:


server/api.go
```diff
@@ -87,6 +173,13 @@ func (a *API) processActivity(w http.ResponseWriter, req *http.Request) {
 	a.p.API.LogDebug("Change activity request", "activities", activities)
 	errors := ""
 	for _, activity := range activities.Value {
+		if activity.EncryptedContent.Data != "" {
+			content, err := a.processEncryptedContent(activity.EncryptedContent)
+			if err != nil {
+				a.p.API.LogError("Invalid encrypted content", "error", err)
```
ayusht2810, line 179: don't we need to add `continue` here if we were not able to process the content

server/handlers/handlers.go
```diff
@@ -179,10 +179,26 @@ func (ah *ActivityHandler) handleActivity(activity msteams.Activity) {
 	switch activity.ChangeType {
 	case "created":
 		ah.plugin.GetAPI().LogDebug("Handling create activity", "activity", activity)
-		ah.handleCreatedActivity(activityIds)
+		var msg *msteams.Message
+		if len(activity.Content) > 0 {
+			var err error
+			msg, err = msteams.GetMessageFromJSON(activity.Content, activityIds.TeamID, activityIds.ChannelID, activityIds.ChatID)
+			if err != nil {
+				ah.plugin.GetAPI().LogDebug("Unable to unmarshal activity message", "activity", activity)
```
ayusht2810, line 187:
```suggestion
				ah.plugin.GetAPI().LogDebug("Unable to unmarshal activity message", "activity", activity, "error", err)
```

server/api.go
jespino: My way of thinking here is, I try to get the encrypted content, if it doesn't work, just keep going, it doesn't matter, because in that case you are going to try to use the "standard" method of getting the information from the server.

server/msteams/client.go
```diff
@@ -126,6 +126,15 @@ type Activity struct {
 	LifecycleEvent                 string
 	SubscriptionExpirationDateTime time.Time
 	SubscriptionID                 string
+	EncryptedContent               EncryptedContent
```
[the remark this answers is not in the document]
jespino: My way of thinking was, using pointer here makes it harder to handle later. The zero value of that structure is an empty `Data` field anyway, so I can check that. If there is no `Data` there is nothing to decrypt, so that is good enough for me, and I don't need to deal with the nil checks. In this case is not a big deal to manage the nil checks, because it is very localized, and probably is more semantically correct. So... I think I going to change it to your proposal.

server/msteams/client.go
```diff
@@ -893,6 +909,91 @@ func (tc *ClientImpl) GetReply(teamID, channelID, messageID, replyID string) (*M
 	return convertToMessage(res, teamID, channelID, ""), nil
 }
 
+func GetMessageFromJSON(data []byte, teamID, channelID, chatID string) (*Message, error) {
```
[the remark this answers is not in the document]
manojmalik20: What are your thoughts on this comment? @jespino

server/api.go
```diff
@@ -67,6 +76,83 @@ func (a *API) getAvatar(w http.ResponseWriter, r *http.Request) {
...
+		a.p.API.LogDebug("Unable to decode key", "error", err)
+		return nil, err
+	}
+
+	key, err := a.p.getPrivateKey()
+	if err != nil {
+		a.p.API.LogDebug("Unable to get private key", "error", err)
+		return nil, err
+	}
+	hash := sha1.New() //nolint:gosec
```
esarafianou, line 132: Let's use sha256 instead of sha1. I think that's supported in MS with the naming `RSA-OAEP-256`: https://learn.microsoft.com/en-us/azure/key-vault/keys/about-keys-details. MS doc on how to decrypt the data key only mentioned OAEP without specifying the hash, so I hope that sha256 is supported there too: https://learn.microsoft.com/en-us/graph/webhooks-with-resource-data#decrypting-resource-data

server/api.go
esarafianou: 1. By standard method you mean the webhook secret?
2. What are some cases where the encrypted content may be invalid but yet be a valid request overall?

server/plugin.go
```diff
@@ -169,6 +172,41 @@ func (p *Plugin) syncSince(syncSince time.Time) {
...
+
+func (p *Plugin) getPrivateKey() (*rsa.PrivateKey, error) {
+	keyPemString := p.getConfiguration().CertificateKey
+	if keyPemString == "" {
+		return nil, errors.New("certificate private key not configured")
+	}
+	privPem, _ := pem.Decode([]byte(keyPemString))
+	var err error
+	var parsedKey interface{}
+	if parsedKey, err = x509.ParsePKCS1PrivateKey(privPem.Bytes); err != nil {
```
esarafianou, line 192: Why are we checking all these 3 paddings?

ayusht2810 (approved):

server/api.go
jespino: By standard method I mean, using the current approach. Trying to get the information from the server doing a API request by the user that belongs to the channel.

The most common case of encrypted content to be invalid is the absence of encrypted content entirely. If you are doing "request forgery" there, you can do it by cleaning up the encrypted content, instead of passing an invalid content enayway.

server/msteams/client.go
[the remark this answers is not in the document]
jespino: I don't love `util.go` files or `utils` packages, because that means nothing, and normally they end up being a bucket full of entropy, so I prefer to have another filename that give meaning to the content, even if that means the file or package only contains that function.

server/api.go
jespino: And yes, in any case we are verifying the webhook secret.

server/plugin.go
jespino: To be honest, I was using an example of handling padding from certain go blog post, so... that was the approach to ensure everything was working there. But the padding should be an specific one, so I should change it to the one that is defined in the spec. I'll do the change.

esarafianou: Thanks!

esarafianou (approved):

esarafianou:
Pending the following change:
if the public cerificate is set in the config and we receive something unencrypted or we're unable to decrypt it, we should fail entirely

jespino:
@esarafianou the last commit address the concerns about the processing messages without encrypted data when encryption is activated.

esarafianou:
@jespino quite a late response from my side but the part with enforcing processing messages with encrypted data when encryption is activated looks good!

server/msteams/client.go
```diff
@@ -1154,6 +1170,83 @@ func (tc *ClientImpl) GetReply(teamID, channelID, messageID, replyID string) (*c
...
+	}{}
+
+	if err := json.Unmarshal(data, &msg); err != nil {
+		return nil, err
+	}
+
+	mentions := []clientmodels.Mention{}
+	for _, m := range msg.Mentions {
+		mention := clientmodels.Mention{}
+		if m.ID != 0 && m.MentionText != "" {
```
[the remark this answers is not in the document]
jespino: I think we should merge this and fix that later.

manojmalik20 (approved):
LGTM, apart from the mentions bug. We can definitely fix that in a separate PR but I think we should do a QA review on this before merging this. @AayushChaudhary0001 Can you please do a QA review of this?

AayushChaudhary0001:
I'll take a look at this on priority.

server/store/sqlstore/store.go
```diff
@@ -706,10 +718,10 @@ func (s *SQLStore) DeleteSubscription(subscriptionID string) error {
 }
 
 func (s *SQLStore) GetChannelSubscription(subscriptionID string) (*storemodels.ChannelSubscription, error) {
-	row := s.getQueryBuilder().Select("subscriptionID, msTeamsChannelID, msTeamsTeamID, secret, expiresOn").From(subscriptionsTableName).Where(sq.Eq{"subscriptionID": subscriptionID, "type": subscriptionTypeChannel}).Suffix("FOR UPDATE").QueryRow()
+	row := s.getQueryBuilder().Select("subscriptionID, msTeamsChannelID, msTeamsTeamID, secret, expiresOn, certificate").From(subscriptionsTableName).Where(sq.Eq{"subscriptionID": subscriptionID, "type": subscriptionTypeChannel}).Suffix("FOR UPDATE").QueryRow()
 	var subscription storemodels.ChannelSubscription
 	var expiresOn int64
-	if err := row.Scan(&subscription.SubscriptionID, &subscription.ChannelID, &subscription.TeamID, &subscription.Secret, &expiresOn); err != nil {
+	if err := row.Scan(&subscription.SubscriptionID, &subscription.ChannelID, &subscription.TeamID, &subscription.Secret, &expiresOn, &subscription.Certificate); err != nil {
```
manojmalik20, line 724: I tested this branch on a server using MySQL and this line errors out with the error:
```
sql: Scan error on column index 5, name "certificate": converting NULL to string is unsupported
```

I've not tested this on Postgres yet. I'll update on this PR what I find.

manojmalik20:
I'm getting the same error on Postgres as well.

This is when I'm uploading this build on a server where the plugin is already running (without the certificate based subscriptions).

server/api.go
```diff
@@ -107,8 +193,21 @@ func (a *API) processActivity(w http.ResponseWriter, req *http.Request) {
...
 	errors := ""
 	for _, activity := range activities.Value {
+		if activity.EncryptedContent != nil {
+			content, err := a.processEncryptedContent(*activity.EncryptedContent)
+			if err != nil {
+				errors += err.Error() + "\n"
+				continue
+			}
+			activity.Content = content
+		} else if requireEncryptedContent {
```
manojmalik20, line 206: I don't think we should error out if there's no encrypted content. We can use the old method to fetch the message content in that case. This can be useful for backwards compatibility. For eg - If there's an issue in updating the subscriptions by the monitoring system, the change notifications that come in before the next run of the monitoring system will be discarded and that should not happen.

server/store/sqlstore/store.go
jespino: I'll check it, thanks

server/api.go
jespino: In what scenario I going to receive a message without the encrypted content in a certificate based subscription?

server/store/sqlstore/store.go
jespino: Not should be fixed, if you can try again, it would be great.

server/api.go
manojmalik20: Suppose you have the plugin running without the certificate based subscriptions. When you update the plugin and add certificates in the config, the monitoring system attempts to delete the existing subscriptions and create new ones with certificates. But suppose there was an error in deleting the old subscription. Then we're gonna have a subscription without the certificate until the next run of the monitoring system. The monitoring system will run after one minute. In that one minute interval, we can receive change notifications which won't have the encrypted content as they are from the subscription without the certificate. So, should we discard those change notifications?

server/store/sqlstore/store.go
```diff
@@ -563,9 +563,15 @@ func (s *SQLStore) ListChatSubscriptionsToCheck() ([]storemodels.ChatSubscriptio
...
-		if scanErr := rows.Scan(&subscription.SubscriptionID, &subscription.UserID, &subscription.Secret, &expiresOn, &subscription.Certificate); scanErr != nil {
+		var certificate *string
+		if scanErr := rows.Scan(&subscription.SubscriptionID, &subscription.UserID, &subscription.Secret, &expiresOn, &certificate); scanErr != nil {
 			return nil, scanErr
 		}
+		if certificate == nil {
+			subscription.Certificate = ""
+		} else {
+			subscription.Certificate = *certificate
+		}
```
manojmalik20, line 574: Can't we just do this?
```
if certificate != nil {
    subscription.Certificate = *certificate
}
```

manojmalik20: We should initialize the `subscription` variable such the `Certificate` field already has the value `""`.

server/api.go
manojmalik20: While testing the latest build, I was also able to reproduce this scenario. After the new build is installed, the subscriptions were not recreated instantly. Instead, they were deleted and recreated after the 1st minute. During that interval, all of the change notifications were discarded.

server/store/sqlstore/store.go
jespino: Good point

jespino: Done

manojmalik20 (approved):

PR #161, opened 2023-06-05
[MI-3141] Added the logic to un-delete a MM post when Teams post is un-deleted

manojmalik20:
#### Summary
- Created a new store function to recover post from MM and un-delete it
- Added the handling of the event when user un-deletes a post on MS Teams

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/131


server/handlers/handlers.go
```diff
@@ -333,10 +333,17 @@ func (ah *ActivityHandler) handleUpdatedActivity(activityIds msteams.ActivityIds
...
 	} else {
-		post, err := ah.plugin.GetAPI().GetPost(postInfo.MattermostID)
-		if err != nil {
-			ah.plugin.GetAPI().LogError("Unable to find the original post", "error", err.Error())
-			return
+		post, postErr := ah.plugin.GetAPI().GetPost(postInfo.MattermostID)
+		if postErr != nil {
+			if strings.Contains(postErr.Error(), "Unable to get the post.") {
+				if err = ah.plugin.GetStore().RecoverPost(postInfo.MattermostID); err != nil {
+					ah.plugin.GetAPI().LogError("Unable to recover the post", "post", post, "error", err)
```
ayusht2810, line 340:
```suggestion
					ah.plugin.GetAPI().LogError("Unable to recover the post", "post", post, "error", err.Error())
```

ayusht2810, line 340: don't we need to return from here?

server/handlers/handlers.go
```diff
@@ -361,16 +368,22 @@ func (ah *ActivityHandler) handleUpdatedActivity(activityIds msteams.ActivityIds
 
 	_, appErr := ah.plugin.GetAPI().UpdatePost(post)
 	if appErr != nil {
-		ah.plugin.GetAPI().LogError("Unable to update post", "post", post, "error", appErr)
-		return
+		if strings.Contains(appErr.Error(), "Unable to get the post.") {
+			if err = ah.plugin.GetStore().RecoverPost(post.Id); err != nil {
+				ah.plugin.GetAPI().LogError("Unable to recover the post", "post", post, "error", err)
```
ayusht2810, line 373: same

server/store/store.go
```diff
@@ -745,6 +746,16 @@ func (s *SQLStore) DeleteDMAndGMChannelPromptTime(userID string) error {
 	return nil
 }
 
+func (s *SQLStore) RecoverPost(postID string) error {
+	query := s.getQueryBuilder().Update("Posts").Set("DeleteAt", 0).Where(sq.Eq{"Id": postID}, sq.NotEq{"DeleteAt": 0})
+	_, err := query.Exec()
+	if err != nil {
+		return err
+	}
```
ayusht2810, line 754:
```suggestion
	if _, err := query.Exec(); err != nil {
		return err
	}
```

server/handlers/handlers.go
manojmalik20: There's no fixed convention for this in this plugin.

server/handlers/handlers.go
```diff
@@ -333,10 +333,18 @@ func (ah *ActivityHandler) handleUpdatedActivity(activityIds msteams.ActivityIds
 			return
 		}
 	} else {
-		post, err := ah.plugin.GetAPI().GetPost(postInfo.MattermostID)
-		if err != nil {
-			ah.plugin.GetAPI().LogError("Unable to find the original post", "error", err.Error())
-			return
+		post, postErr := ah.plugin.GetAPI().GetPost(postInfo.MattermostID)
+		if postErr != nil {
+			if strings.Contains(postErr.Error(), "Unable to get the post.") {
```
jespino, line 338: The error is better compared by the Error.Id, instead of the `Error` string, that can be translated.

server/handlers/handlers.go
```diff
@@ -361,16 +369,23 @@ func (ah *ActivityHandler) handleUpdatedActivity(activityIds msteams.ActivityIds
 
 	_, appErr := ah.plugin.GetAPI().UpdatePost(post)
 	if appErr != nil {
-		ah.plugin.GetAPI().LogError("Unable to update post", "post", post, "error", appErr)
-		return
+		if strings.Contains(appErr.Error(), "Unable to get the post.") {
```
jespino, line 372: Same here, I prefer to check the `appErr.Id` to avoid problems with translations.

jespino (changes requested):
A couple of changes are needed, but it looks good to me. A very smart approach to trigger the WebSocket event.

Makefile
```diff
@@ -5,7 +5,7 @@ CURL ?= $(shell command -v curl 2> /dev/null)
 MM_DEBUG ?=
 MANIFEST_FILE ?= plugin.json
 GOPATH ?= $(shell go env GOPATH)
-GO_TEST_FLAGS ?= -race
```
jespino, line 8: Why are you removing the `-race` here? we want to check the race conditions in CI

ayusht2810 (approved):

Makefile
manojmalik20: Hey @jespino, this was an accidental change.

jespino (approved):
LGTM

PR #166, opened 2023-06-12
[MI-3157] Fix issue of replies from MM to Teams coming as new message…

ayusht2810:
…s on the Teams for DMs and GMs

#### Summary
Fixed the issue of replies from MM to Teams coming as new messages on the Teams for DMs and GMs.

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/164

### DIFF
--- a/server/msteams/client.go
+++ b/server/msteams/client.go
@@ -18,6 +18,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	azidentity "github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/mattermost/mattermost-plugin-msteams-sync/server/utils"
 	msgraphsdk "github.com/microsoftgraph/msgraph-beta-sdk-go"
 	"github.com/microsoftgraph/msgraph-beta-sdk-go/chats"
 	"github.com/microsoftgraph/msgraph-beta-sdk-go/drives"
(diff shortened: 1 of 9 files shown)


server/msteams/client.go
```diff
@@ -145,6 +146,22 @@ type GraphAPIError struct {
...
+	UserIdentityType string `json:"userIdentityType"`
+	ID               string `json:"id"`
+	DisplayName      string `json:"displayName"`
+}
+
+type ChatAttachmentSender struct {
+	User ChatAttachmentUser `json:"user"`
+}
+
+type ChatAttachment struct {
```
manojmalik20, line 159:
```suggestion
type ChatMessageAttachment struct {
```
I think it would be more suitable. Let me know your thoughts.

server/utils/utils.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/mattn/godown"
+)
+
+func ConvertToMD(text string) string {
+	if !strings.Contains(text, "<div>") && !strings.Contains(text, "<p>") {
+		return text
+	}
+	var sb strings.Builder
+	err := godown.Convert(&sb, strings.NewReader(text), nil)
+	if err != nil {
```
manojmalik20, line 15:
```suggestion
	if err := godown.Convert(&sb, strings.NewReader(text), nil); err != nil {
```

server/message_hooks_test.go
```diff
@@ -943,7 +943,7 @@ func TestSendChat(t *testing.T) {
 			},
 			SetupClient: func(client *clientmocks.Client, uclient *clientmocks.Client) {
 				uclient.On("CreateOrGetChatForUsers", mock.Anything).Return("mockChatID", nil).Times(1)
-				uclient.On("SendChat", "mockChatID", "", "<p>mockMessage</p>\n", []models.ChatMessageMentionable{}).Return(&msteams.Message{
+				uclient.On("SendChat", "mockChatID", "<p>mockMessage</p>\n", (*msteams.Message)(nil), []models.ChatMessageMentionable{}).Return(&msteams.Message{
```
manojmalik20, line 946: Please add a new unit test also for covering the parent message block.

manojmalik20 (approved):

server/utils/utils.go
```diff
@@ -0,0 +1,18 @@
+package utils
```
jespino, line 1: Utils means nothing, is the same thing of calling it "stuff", or "magic". In this case, we can create a "markdown" package, if you want, and put here any utilities functions related to markdown. But creating a `utils` package is a bad idea in general.

jespino (changes requested):
I like this PR, except for the fact of having the new `utils` package that I don't like, other than that, looks good to me.

ayusht2810:
@jespino fixed the review comments. Please re-review

jespino (approved):

jespino:
It needs conflict resolution, but other than that looks good to me.

PR #168, opened 2023-06-13
[MI-3161] Fixed the issue of reactions causing infinite loop in both channels and chats

manojmalik20:
#### Summary
- Created two new store functions for updating last post updateAt time by both MM post id and MS Teams post id
- Modified the return type of error handling function and modified some logs
- Added the logic of updating post last updateAt time when reactions are updated in MM

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/165

### DIFF
--- a/server/message_hooks.go
+++ b/server/message_hooks.go
@@ -241,13 +241,12 @@ func (p *Plugin) SetChatReaction(teamsMessageID, srcUser, channelID string, emoj
 
 	teamsMessage, err := client.GetChatMessage(chatID, teamsMessageID)
 	if err != nil {
-		p.API.LogWarn("Error getting the msteams post metadata", "error", err)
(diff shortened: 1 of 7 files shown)


server/message_hooks.go
```diff
@@ -239,6 +239,16 @@ func (p *Plugin) SetChatReaction(teamsMessageID, srcUser, channelID string, emoj
 		return err
 	}
 
+	teamsMessage, err := client.GetChatMessage(chatID, teamsMessageID)
```
jespino, line 242: I don't love this extra call needed, but I guess is the most robust approach. I would love to have a better way to do it. Anyway, this is not going to be a dramatic problem because the number of reactions should be relatively low, I guess.

jespino (approved):
LGTM

PR #175, opened 2023-06-19
[MI-3172] Fixed the issue of attachments being removed when post is edited in MM

manojmalik20:
#### Summary
Added a TODO comment in the update function for the channels to handle attachments
Added the logic for getting the existing attachments from a post and sending them in the update API

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/172


jespino (approved):
LGTM

PR #189, opened 2023-06-26
[MI-3208] Fix issue in naming of attachments

ayusht2810:
#### Summary
Fix the issue of attachments not working correctly due to issue in naming of attachments

### DIFF
--- a/server/message_hooks.go
+++ b/server/message_hooks.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"math"
-	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
@@ -418,11 +418,7 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post) (
 			continue
 		}
 
-		fileName, fileExtension := getExtension(fileInfo.Name)
-		if fileName == "" || fileExtension == "" {
-			continue
-		}
-
+		fileName, fileExtension := getFileNameAndExtension(fileInfo.Name)
 		var attachment *msteams.Attachment
 		attachment, err = client.UploadFile("", "", fileName+"_"+fileInfo.Id+fileExtension, int(fileInfo.Size), fileInfo.MimeType, bytes.NewReader(fileData))
 		if err != nil {
@@ -509,11 +505,7 @@ func (p *Plugin) Send(teamID, channelID string, user *model.User, post *model.Po
 			continue
 		}
 
-		fileName, fileExtension := getExtension(fileInfo.Name)
-		if fileName == "" || fileExtension == "" {
-			continue
-		}
-
+		fileName, fileExtension := getFileNameAndExtension(fileInfo.Name)
 		var attachment *msteams.Attachment
 		attachment, err = client.UploadFile(teamID, channelID, fileName+"_"+fileInfo.Id+fileExtension, int(fileInfo.Size), fileInfo.MimeType, bytes.NewReader(fileData))
 		if err != nil {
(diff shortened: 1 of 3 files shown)


server/message_hooks.go
```diff
@@ -416,8 +417,15 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post) (
 			continue
 		}
 
+		fileNameData := strings.Split(fileInfo.Name, ".")
+		if len(fileNameData) < 1 {
+			continue
+		}
+
+		fileName, fileExtension := strings.Join(fileNameData[:len(fileNameData)-1], "."), fileNameData[len(fileNameData)-1]
```
manojmalik20, line 425: Use this - https://cs.opensource.google/go/go/+/go1.20.5:src/path/filepath/path.go;l=273

server/message_hooks_test.go
```diff
@@ -1042,7 +1042,7 @@ func TestSendChat(t *testing.T) {
 			},
 			SetupClient: func(client *clientmocks.Client, uclient *clientmocks.Client) {
 				uclient.On("CreateOrGetChatForUsers", mock.Anything).Return("mockChatID", nil).Times(1)
-				uclient.On("UploadFile", "", "", "mockFileName"+"_"+testutils.GetID(), 1, "mockMimeType", bytes.NewReader([]byte("mockData"))).Return(&msteams.Attachment{
+				uclient.On("UploadFile", "", "", "mockFile.Name"+"_"+testutils.GetID()+".txt", 1, "mockMimeType", bytes.NewReader([]byte("mockData"))).Return(&msteams.Attachment{
```
manojmalik20, line 1045: Why is there `dot` in the filename? `mockFile.Name`

server/testutils/data.go
```diff
@@ -136,7 +136,7 @@ func GetChannelLinks(count int) []*storemodels.ChannelLink {
 func GetFileInfo() *model.FileInfo {
 	return &model.FileInfo{
 		Id:       GetID(),
-		Name:     "mockFileName",
+		Name:     "mockFile.Name.txt",
```
manojmalik20, line 139:
```suggestion
		Name:     "mockFileName.txt",
```

ayusht2810: this was done to test whether we are properly fetching the file extension

manojmalik20 (approved):

server/message_hooks.go
```diff
@@ -416,8 +418,13 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post) (
 			continue
 		}
 
+		fileName, fileExtension := getExtension(fileInfo.Name)
+		if fileName == "" || fileExtension == "" {
```
jespino, line 422: This can lead to not uploading files without extension what is a legit usage. I don't think not having file extension is a reason to ignore the file. I don't even think that not having a filename is a reason to ignore it.

server/message_hooks.go
```diff
@@ -502,8 +509,13 @@ func (p *Plugin) Send(teamID, channelID string, user *model.User, post *model.Po
...
```
jespino, line 513: same here

server/message_hooks.go
```diff
@@ -838,3 +850,13 @@ func (p *Plugin) getMentionsData(message, teamID, channelID, chatID string, clie
 
 	return message, mentions
 }
+
+func getExtension(path string) (string, string) {
```
jespino, line 854: This looks too manual, why not use `path/filepath.Base` to get the filename and `path/filepath.Ext` to get the extension, and then remove the extension from the filename? I think that would be ess error prone and way easier to understand.

jespino (changes requested):
some improvements suggested.

ayusht2810:
@jespino fixed the review fixes. Please re-review

ayusht2810:
@jespino reminder to review the PR

jespino (approved):
LGTM

PR #201, opened 2023-07-04
[MI-3237] Added the check for image resolution while handling attachments from MS Teams

manojmalik20:
#### Ticket Link
https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/197

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	github.com/containerd/containerd v1.6.19 // indirect
 	github.com/cpuguy83/dockercfg v0.3.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/disintegration/imaging v1.6.2 // indirect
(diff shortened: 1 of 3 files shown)


esarafianou (approved):

esarafianou:
Shouldn't we also check the maximum raw image file size being max 254 MB?

ayusht2810 (approved):

manojmalik20:
@esarafianou The resolution check automatically covers the raw image file size check. I checked in the Mattermost server code and there isn't any specific check for this there as well.

jespino (approved):
I'm missing a test to verify this code, I'll approve it just in case you want to merge it before adding the test. But I would love to have the test for it, at least after merge.

manojmalik20:
The tests for this are already in progress. Merging this now, will create a separate PR for unit tests.

PR #210, opened 2023-07-12
Propose naming the plugin "MS Teams Connector"

jasonblais:
Given this plugin now supports two use cases -- a) synchronizing channels between the two platforms b) embedding Mattermost in Microsoft Teams -- and our marketing collateral refers to our solution as [Mattermost for Microsoft Teams](https://mattermost.com/solutions/mattermost-for-microsoft-teams/), propose using consistent naming for the integration as well.

We should also consider naming this repo as 'mattermost-for-microsoft-teams' for consistency.

Definitely open for feedback & if there are additional conversations we've held around naming this integration.

cc @johndavidlugtu @jespino @wiggin77 @azigler @mkdbns

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -1,6 +1,6 @@
 {
     "id": "com.mattermost.msteams-sync",
-    "name": "MSTeams Sync",
+    "name": "Mattermost for Microsoft Teams",
     "description": "Plugin to sync Mattermost channels with Microsoft Teams channels",
     "homepage_url": "https://github.com/mattermost/mattermost-plugin-msteams-sync",
     "support_url": "https://github.com/mattermost/mattermost-plugin-msteams-sync/issues",
--- a/plugin.json
+++ b/plugin.json
@@ -1,6 +1,6 @@
 {
     "id": "com.mattermost.msteams-sync",
-    "name": "Mattermost for Microsoft Teams",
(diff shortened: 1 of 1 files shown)


wiggin77:
If we find that customers generally are not interested in both use cases at the same time, we should reconsider whether these should be in the same plugin.

wiersgallak:
Previously, we had landed on the MS Teams Connector when we built the marketplace page: https://mattermost.com/marketplace/mattermost-ms-teams-connector/ We included this in the newest 8.0 blog post, so if the name is changing, we need to cascade to these other materials.

jasonblais:
Thanks @johndavidlugtu! I'd consider them as two use cases that the connector supports.

@wiggin77 There are pros and cons, let's see how customers adopt & we can make adjustments as needed? We have at least one customer interested in both use cases.

@wiersgallak Thanks! Wasn't aware, have updated the name to `MS Teams Connector`

jespino:
I like this, but I don't know if we should change everything there. The plugin ID, the repo name, etcetera, in that case we need more changes. It is a good time now because we don't have any official big installation out there, at least that we know.

lieut-data:
Tracking this issue over at https://mattermost.atlassian.net/browse/MM-55478.

PR #227, opened 2023-07-26
[MI-3333] Add console config to handle guest users and deactivated users on MS Teams

### LINKED ISSUE #218: Guest accounts in MS Teams should not appear in MM
All users in the MS Teams tenant are being synced to MM. ServiceNow needs a way to filter out guests.

An idea they have is to have the ability to filter based on domain. In this case, they only want users with the @servicenow.com domain to be synced.

ayusht2810:
#### Summary

- Added logic to disable and enable syncing of guest users using console config.
- Added logic to deactivate synced users on Mattermost corresponding to MS Teams.

#### Ticket Link
Fixes #218

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -444,26 +444,24 @@ func (p *Plugin) syncUsers() {
 		mmUser, isUserPresent := mmUsersMap[msUser.Mail]
 
 		if isUserPresent && isRemoteUser(mmUser) {
-			if !msUser.IsAccountEnabled {
-				// Deactivate the active Mattermost user corresponding to MS Teams user.
-				if mmUser.DeleteAt == 0 {
-					p.API.LogDebug("Deactivating the Mattermost user account", "Email", msUser.Mail)
-					if err := p.API.UpdateUserActive(mmUser.Id, false); err != nil {
-						p.API.LogError("Unable to deactivate the Mattermost user account", "Email", mmUser.Email, "Error", err.Error())
-					}
-
-					continue
-				}
-			} else {
+			if msUser.IsAccountEnabled {
 				// Activate the deactived Mattermost user corresponding to MS Teams user.
 				if mmUser.DeleteAt != 0 {
 					p.API.LogDebug("Activating the inactive user", "Email", msUser.Mail)
 					if err := p.API.UpdateUserActive(mmUser.Id, true); err != nil {
 						p.API.LogError("Unable to activate the user", "Email", msUser.Mail, "Error", err.Error())
 					}
-
-					continue
 				}
+			} else {
+				// Deactivate the active Mattermost user corresponding to MS Teams user.
+				if mmUser.DeleteAt == 0 {
+					p.API.LogDebug("Deactivating the Mattermost user account", "Email", msUser.Mail)
+					if err := p.API.UpdateUserActive(mmUser.Id, false); err != nil {
+						p.API.LogError("Unable to deactivate the Mattermost user account", "Email", mmUser.Email, "Error", err.Error())
+					}
+				}
+
+				continue
 			}
 		}
 
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -420,7 +420,7 @@ func (p *Plugin) syncUsers() {
 	}
 
 	p.API.LogDebug("Count of MS Teams users", "count", len(msUsers))
-	mmUsers, appErr := p.API.GetUsers(&model.UserGetOptions{Active: true, Page: 0, PerPage: math.MaxInt32})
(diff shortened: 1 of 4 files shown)


server/plugin.go
```diff
@@ -460,6 +490,24 @@ func (p *Plugin) syncUsers() {
 			for {
 				newUser, appErr = p.API.CreateUser(newMMUser)
 				if appErr != nil {
+					if appErr.Id == "app.user.save.email_exists.app_error" {
```
manojmalik20, line 493: There will be no need for this check if we get all the users in the above API call.

server/plugin.go
```diff
@@ -460,6 +490,24 @@ func (p *Plugin) syncUsers() {
...
+						user, gErr := p.API.GetUserByEmail(msUser.Mail)
+						if user.RemoteId != nil && *user.RemoteId != "" && strings.HasPrefix(user.Username, "msteams_") {
```
manojmalik20, line 495: Create a function for this.

server/plugin.go
```diff
@@ -442,6 +443,35 @@ func (p *Plugin) syncUsers() {
...
+					p.API.LogError("Unable to deactivate the Mattermost user account", "Email", mmUser.Email, "Error", err.Error())
+				}
+			}
+
+			continue
+		}
+
+		if msUser.Type == "Guest" {
+			if !syncGuestUsers {
+				if !ok {
```
manojmalik20, line 459: Please add comments to make the code easier to understand.

server/plugin.go
```diff
@@ -460,6 +490,24 @@ func (p *Plugin) syncUsers() {
...
+							if gErr != nil {
+								p.API.LogError("Unable to get the invactive user", "Email", msUser.Mail, "Error", gErr.Error())
+								break
+							}
+
+							if err := p.API.UpdateUserActive(user.Id, true); err != nil {
+								p.API.LogError("Unable to activate the user", "Email", msUser.Mail, "Error", err.Error())
+							}
+						} else {
+							p.API.LogError("Unable to create new user", "Email", msUser.Mail, "Error", gErr.Error())
```
manojmalik20, line 506:
```suggestion
							p.API.LogError("Unable to create new user", "Email", msUser.Mail, "Error", appErr.Error())
```
Shouldn't it be like this?

server/plugin.go
```diff
@@ -441,26 +441,42 @@ func (p *Plugin) syncUsers() {
...
+					p.API.LogDebug("Deactivating the Mattermost user account", "Email", msUser.Mail)
+					if err := p.API.UpdateUserActive(mmUser.Id, false); err != nil {
+						p.API.LogError("Unable to deactivate the Mattermost user account", "Email", mmUser.Email, "Error", err.Error())
+					}
+
+					continue
 				}
-			}
+			} else {
+				// Activate the deactive Mattermost user corresponding to MS Teams user.
```
manojmalik20, line 458:
```suggestion
				// Activate the deactivated Mattermost user corresponding to the MS Teams user.
```

server/plugin.go
```diff
@@ -441,26 +441,42 @@ func (p *Plugin) syncUsers() {
...
+					}
 
-			continue
+					continue
+				}
+			}
 		}
 
 		if msUser.Type == "Guest" {
+			// Check if syncing of MS Teams guest user is disabled.
```
manojmalik20, line 471:
```suggestion
			// Check if syncing of MS Teams guest users is disabled.
```

server/plugin.go
```diff
@@ -441,26 +441,42 @@ func (p *Plugin) syncUsers() {
...
 			if !syncGuestUsers {
-				if !ok {
+				// Skip syncing of MS Teams guest user.
```
manojmalik20, line 473: I think it will be easier to read and understand if you reverse the if condition and move the comment inside `if` statement.

jespino (approved):
LGTM

server/plugin.go
```diff
@@ -444,36 +444,34 @@ func (p *Plugin) syncUsers() {
...
 			}
 		}
 
 		if msUser.Type == "Guest" {
 			// Check if syncing of MS Teams guest users is disabled.
 			if !syncGuestUsers {
 				if isUserPresent {
-					// Deactivate the Mattermost user corresponding to the MS Teams guest user.
-					p.API.LogDebug("Deactivating the guest user account", "Email", msUser.Mail)
 					if isRemoteUser(mmUser) {
```
manojmalik20: Maybe we should do this?
```
if isUserPresent && isRemoteUser(mmUser) {
```

PR #232, opened 2023-08-02
[MI-3362] Update documentation for Sync guest user config settings in the system console

ayusht2810:
#### Summary
Update documentation for Sync guest user config settings in the system console.


docs/plugin_setup.md
```diff
@@ -11,7 +11,8 @@
     - **Enforce connected accounts**: Enabling this will enforce all the users to connect their Mattermost accounts to their MS Teams accounts.
     - **Allow to temporarily skip connect user**: Enabling this will allow the users to temporarily skip connecting their accounts.
     - **Sync users**: This config is for the interval (in minutes) in which the users will be synced between Mattermost and MS Teams. If you leave it empty, it will not syncrhonize the users.
+    - **Sync guest users**: Enabling this configuration will sync the MS Teams guest users on Mattermost.
```
manojmalik20, line 14: We should also mention that it will affect the existing users also. That if any user was created before with the sync user functionality but now it has been converted to guest user, then it will be deactivated in Mattermost also. Something like this. We should mention all the implications of this setting.

manojmalik20 (approved):

jespino (approved):

PR #233, opened 2023-08-03
[MI-3368] Fix issue of messages being posted from inactive users on Mattermost corresponding to MS Teams user

ayusht2810:
#### Summary
Fix issue of messages being posted from inactive users on Mattermost corresponding to MS Teams user

### DIFF
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -241,6 +241,10 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 		senderID = ah.plugin.GetBotUserID()
 	}
 
+	if isActiveUser := ah.isActiveUser(senderID); !isActiveUser {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


server/handlers/handlers.go
```diff
@@ -486,3 +499,18 @@ func (ah *ActivityHandler) updateLastReceivedChangeDate(t time.Time) {
 		ah.plugin.GetAPI().LogError("Unable to store properly the last received change")
 	}
 }
+
+func (ah *ActivityHandler) isActiveUser(userID string) bool {
+	mmUser, err := ah.plugin.GetAPI().GetUser(userID)
```
manojmalik20, line 504: Every time this function is called, we are calling the Mattermost API `GetUser` to get the user object. Check if in some of the places, we're already calling this API or not. If yes, then there's no need to call it again here.

manojmalik20 (approved):

PR #239, opened 2023-08-14
[MI-3372] Add feature to disconnect user from MS Teams account using RHS

ayusht2810:
#### Summary

1. Add a feature to disconnect users from MS Teams accounts.
2. Added and updated icon positions in the RHS.
3. Added the display name of the currently logged-in user in the RHS.

image
image
image


webapp/src/constants/index.ts
```diff
@@ -1,6 +1,9 @@
 import {id as pluginId} from '../manifest';
 
 const iconUrl = `/plugins/${pluginId}/public/msteams-sync-icon.svg`;
+const mattermostIconUrl = `/plugins/${pluginId}/public/mattermost-icon.svg`;
+const msteamsPrivateChannelIconUrl = `/plugins/${pluginId}/public/msteams-private-channel-icon.svg`;
+const msteamsIconUrl = `/plugins/${pluginId}/public/msteams-icon.svg`;
```
raghavaggarwal2308, line 6: constant for common prefix

public/mattermost-icon.svg
raghavaggarwal2308, line 1: Move these files into the constants file

raghavaggarwal2308: Use the latest circular icon for mattermost

raghavaggarwal2308 (changes requested):

PR #281, opened 2023-08-31
[MI-3468] Improved the subscription monitoring system to handle chats subscriptions in a better way

manojmalik20:
#### Summary
- Modified the function ListGlobalSubscriptionsToCheck to return all global subscriptions
- Created a new client function to get a subscription by ID
- Removed the logic of creating new global subscriptions when the plugin starts
- Added the logic of checking the subscription on client side and then handle the subscription in the monitoring system


manojmalik20:
These changes are not required now.

PR #292, opened 2023-09-07
[MI-3485] Fix content-type of APIs and remove additional information from logs

ayusht2810:
#### Summary
Fix content-type of APIs and remove additional information from logs

### DIFF
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -232,7 +232,7 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 	}
 
 	post, errorFound := ah.msgToPost(channelID, senderID, msg, chat)
-	ah.plugin.GetAPI().LogDebug("Post generated", "PostID", post.Id)
+	ah.plugin.GetAPI().LogDebug("Post generated")
 
 	// Avoid possible duplication
 	postInfo, _ := ah.plugin.GetStore().GetPostInfoByMSTeamsID(msg.ChatID+msg.ChannelID, msg.ID)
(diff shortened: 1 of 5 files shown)


server/handlers/handlers.go
```diff
@@ -255,11 +245,11 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 	newPost, appErr := ah.plugin.GetAPI().CreatePost(post)
 
 	if appErr != nil {
-		ah.plugin.GetAPI().LogError("Unable to create post", "post", post, "error", appErr)
+		ah.plugin.GetAPI().LogError("Unable to create post", "error", appErr)
```
raghavaggarwal2308, line 248: Just a question. Why are we removing these?

server/handlers/handlers.go
```diff
@@ -242,7 +232,7 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 	}
 
 	post, errorFound := ah.msgToPost(channelID, senderID, msg, chat)
-	ah.plugin.GetAPI().LogDebug("Post generated", "post", post)
+	ah.plugin.GetAPI().LogDebug("Post generated")
```
manojmalik20, line 235: Can we log post ID only?

manojmalik20 (commented):
Please fix the CI. And also check once more if we're logging the message content anywhere.

server/handlers/handlers.go
ayusht2810: these logs contain post details like messages, which we don't want to log

server/handlers/handlers_test.go
```diff
@@ -260,7 +260,7 @@ func TestHandleCreatedActivity(t *testing.T) {
 			setupAPI: func(mockAPI *plugintest.API) {
 				mockAPI.On("GetDirectChannel", "mockUserID-1", "mockUserID-2").Return(&model.Channel{Id: testutils.GetChannelID()}, nil).Times(1)
 				mockAPI.On("GetUser", testutils.GetUserID()).Return(testutils.GetUser(model.ChannelAdminRoleId, "test@test.com"), nil).Once()
-				mockAPI.On("LogDebug", "Post generated").Times(1)
+				mockAPI.On("LogDebug", "Post generated", "PostID", mock.AnythingOfType("string")).Times(1)
```
manojmalik20, line 263: Why are you using mock.AnythingOfType? You should use this only when you can't use the ID here.

manojmalik20 (approved):

PR #293, opened 2023-09-07
[MI-3487] Fixed the issue of checking file size before downloading it from Teams

manojmalik20:
#### Summary
Fixed the issue of checking file size before downloading it from Teams

### DIFF
--- a/server/handlers/attachments.go
+++ b/server/handlers/attachments.go
@@ -55,7 +55,8 @@ func (ah *ActivityHandler) handleDownloadFile(weburl string, client msteams.Clie
 		return client.GetHostedFileContent(activityIDs)
 	}
 
-	data, err := client.GetFileContent(weburl)
+	fileSizeAllowed := *ah.plugin.GetAPI().GetConfig().FileSettings.MaxFileSize
+	data, err := client.GetFileContent(weburl, fileSizeAllowed)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


server/handlers/attachments.go
```diff
@@ -55,7 +55,8 @@ func (ah *ActivityHandler) handleDownloadFile(weburl string, client msteams.Clie
 		return client.GetHostedFileContent(activityIDs)
 	}
 
-	data, err := client.GetFileContent(weburl)
+	fileSizeAllowed := *ah.plugin.GetAPI().GetConfig().FileSettings.MaxFileSize
```
ayusht2810, line 58: are we sure this value will never be nil?

server/msteams/client.go
```diff
@@ -1218,6 +1218,11 @@ func (tc *ClientImpl) GetFileContent(weburl string) ([]byte, error) {
 		return nil, errors.New("downloadUrl not found")
 	}
 
+	fileSize := item.GetSize()
+	if fileSize != nil && *fileSize > fileSizeAllowed {
+		return nil, fmt.Errorf("skipping file download from MS Graph because file size is greater than allowed size")
```
ayusht2810, line 1223:
```suggestion
		return nil, fmt.Errorf("skipping file download from MS Teams because the file size is greater than allowed size")
```

ayusht2810 (approved):

PR #306, opened 2023-09-13
Adding initial infrestructure for metrics

jespino:


server/plugin.go
```diff
@@ -538,3 +558,31 @@ func getRandomString(characterSet string, length int) string {
...
+			logrus.WithError(err).Error("Metrics server could not be started")
+		}
+	}()
+}
+
+func (p *Plugin) runMetricsUpdaterTask(store store.Store, updateMetricsTaskFrequency time.Duration) {
+	metricsUpdater := func() {
+		stats, err := store.GetStats()
+		if err != nil {
+			logrus.WithError(err).Error("error updating metrics, playbooks_active_total")
```
lieut-data, line 580:
```suggestion
logrus.WithError(err).Error("failed to update computed metrics")
```

server/store/store.go
```diff
@@ -921,6 +928,35 @@ func (s *SQLStore) CompareAndSetJobStatus(jobName string, oldStatus, newStatus b
 	return isUpdated, nil
 }
 
+func (s *SQLStore) GetStats() (*Stats, error) {
+	query := s.getQueryBuilder().Select("count(mmChannelID, mmTeamID) as linkedChannels").From("msteamssync_links").Limit(maxLimitForLinks)
```
lieut-data, line 932: The `.Limit(maxLimitForLinks)` seems redundant here, given there will always be one row?

lieut-data: Also, is`(mmChannelID, mmTeamID)` a compound primary key?

lieut-data (changes requested):
Wow, that was fast @jespino! Just a few things I spotted.

lieut-data:
re: [Removing unneeded metrics](https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/306/commits/253d0bea6aeaa5eabdbe882d6e756d23db0fc387), are we sure we don't want these @jespino? Given this is a discrete Go process, I wonder if it's still useful?

jespino:
> re: [Removing unneeded metrics](https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/306/commits/253d0bea6aeaa5eabdbe882d6e756d23db0fc387), are we sure we don't want these @jespino? Given this is a discrete Go process, I wonder if it's still useful?

Good point, I'll revert that.

jespino:
This is still a bit "draft", I have to do some changes and test it locally.

server/store/store.go
jespino: Fixed all this

server/api.go
```diff
@@ -145,6 +153,28 @@ func (a *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+	now := time.Now()
+
+	a.router.ServeHTTP(recorder, r)
+
+	elapsed := float64(time.Since(now)) / float64(time.Second)
+
+	var routeMatch mux.RouteMatch
+	a.router.Match(r, &routeMatch)
+	if routeMatch.Route != nil {
```
lieut-data, line 169: If for some strange reason we didn't match a route, should we still observe it but categorize as "unknown"?

server/metrics/metrics.go
```diff
@@ -0,0 +1,147 @@
...
+	options := collectors.ProcessCollectorOpts{
+		Namespace: MetricsNamespace,
+	}
+	m.registry.MustRegister(collectors.NewProcessCollector(options))
+	m.registry.MustRegister(collectors.NewGoCollector())
+
+	additionalLabels := map[string]string{}
+	if info.InstallationID != "" {
+		additionalLabels[MetricsCloudInstallationLabel] = os.Getenv("MM_CLOUD_INSTALLATION_ID")
+	}
```
lieut-data, line 51: Should we actually be using `info.InstallationID` vs. fetching it again via `os.Getenv`?

server/api.go
```diff
@@ -145,6 +153,28 @@ func (a *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	a.router.ServeHTTP(w, r)
 }
 
+func (a *API) ServeHTTPWithMetrics(w http.ResponseWriter, r *http.Request) {
```
lieut-data, line 156: I'm curious why this isn't still a middleware?

server/msteams/client.go
```diff
@@ -1119,6 +1119,9 @@ func (tc *ClientImpl) GetReply(teamID, channelID, messageID, replyID string) (*M
 }
 
 func (tc *ClientImpl) GetUserAvatar(userID string) ([]byte, error) {
+	if tc.client == nil {
+		return nil, errors.New("client not initialized")
+	}
```
lieut-data, line 1124: This seems unrelated to these changes -- should we include it in another PR?

lieut-data (changes requested):
Nice, @jespino! Just a few clarifying questions / comments.

server/api.go
jespino: Yes, that is the idea

server/metrics/metrics.go
jespino: Yes, In theory is exactly the same, but makes a lot of sense. I copied that from playbooks, so if you want I can make a PR for that in playbooks too.

server/msteams/client.go
jespino: Let's create a ticket for ensure the plugin doesn't crashes if the plugin is not properly configured.

server/metrics/metrics.go
lieut-data: Haha, oops! Feel free to push a PR, but also fine to just leave it there -- just good to fix at least this while we're here :)

lieut-data (approved):
Thanks, @jespino!

server/metrics/service.go
```diff
@@ -0,0 +1,45 @@
...
+	"net/http"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/sirupsen/logrus"
+)
+
+// Service prometheus to run the server.
+type Server struct {
```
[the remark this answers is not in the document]
jespino: Actually, that is a server so I going to change the name of the file.

jespino:
@manojmalik20 PTAL

server/metrics/metrics.go
```diff
@@ -82,7 +82,7 @@ func NewMetrics(info InstanceInfo) *Metrics {
 		Namespace:   MetricsNamespace,
 		Subsystem:   MetricsSubsystemApp,
 		Name:        "connected_users_total",
-		Help:        "The total number of connected to MS Teams users.",
+		Help:        "The total number of MAttermost users connected to MS Teams users.",
```
lieut-data, line 85:
```suggestion
		Help:        "The total number of Mattermost users connected to MS Teams users.",
```

jespino:
@manojmalik20 feel free to merge it after your review is done. CI is passing, and it is working in my environment.

server/msteams/client.go
jespino: removed from here

server/metrics/metrics.go
jespino: Done

server/api.go
jespino: Done

server/api.go
jespino: This is fixed now.

manojmalik20 (approved):

PR #308, opened 2023-09-14
[MI-3514] Removed the logic of processing post attachments on receiving update activity from MS Teams

### LINKED ISSUE #307: Remove the logic of processing post attachments whenever an update activity is received from Teams

manojmalik20:
#### Summary
Removed the logic of processing post attachments on receiving update activity from MS Teams

#### Ticket Link
Fixes #307


server/handlers/converters.go
```diff
@@ -59,7 +59,10 @@ func (ah *ActivityHandler) msgToPost(channelID, senderID string, msg *msteams.Me
 		text = "## " + msg.Subject + "\n" + text
 	}
 
-	post := &model.Post{UserId: senderID, ChannelId: channelID, Message: text, Props: props, RootId: rootID, FileIds: attachments}
+	post := &model.Post{UserId: senderID, ChannelId: channelID, Message: text, Props: props, RootId: rootID}
+	if isUpdatedActivity {
+		post.FileIds = attachments
+	}
```
ayusht2810, line 65: shouldn't we add file IDs in all cases?

manojmalik20: It is a PATCH API. We should only send those fields which we want to update.

ayusht2810 (approved):

avas27JTG (approved):

Kshitij-Katiyar (approved):

PR #312, opened 2023-09-15
Adding new metrics to the metrics system

jespino:


server/handlers/handlers.go
```diff
@@ -194,21 +211,21 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 			}
 		}
 
-		return
+		return discardedReasonOther
```
jespino, line 214: @manojmalik20 do you mind to take a look to this. It looks like a bug. From my understanding of the code here is, if the user is a guest and the guest user are not synced, we stop processing the event here. My expectation is we still process the event.

server/handlers/handlers.go
```diff
@@ -25,11 +26,23 @@ const (
 	numberOfWorkers       = 20
 	activityQueueSize     = 1000
 	msteamsUserTypeGuest  = "Guest"
+
+	discardedReasonNone                   = ""
+	discardedReasonInvalidChangeType      = "invalid_change_type"
+	discardedReasonIsBotUser              = "is_bot_user"
+	discardedReasonUnableToGetTeamsData   = "unable_to_get_teams_data"
+	discardedReasonNotUserEvent           = "no_user_event"
+	discardedReasonOther                  = "other"
```
jespino, line 35: I'm using this other to not put every single possible type of failure here. But this are counters, so I don't think is that problematic, we only process 3 change types, so the metrics are going to be 3 * number of errors, and because it is a counter, it only add item to the list per combination.

server/handlers/handlers.go
manojmalik20: @jespino This was the requirement from ServiceNow. If a guest user is posting a message on Teams, and the setting `SyncGuestUsers` is off, then we have to discard all his activities. Although, I think that this may be a bug in the "UpdatedActivity" function as we're not doing this there. I'll check this out.

server/metrics/metrics.go
```diff
@@ -120,6 +152,24 @@ func (m *Metrics) ObserveConnectedUsersTotal(count int64) {
...
+
+func (m *Metrics) ObserveProcessedChangeEventTotal(changeType string, discardedReason string) {
+	if m != nil {
+		m.processedChangeEventTotal.With(prometheus.Labels{"change_type": changeType, "discarded_reason": discardedReason}).Inc()
+	}
+}
+
+func (m *Metrics) ObserveLifecycleEventTotal(lifecycleEventType string) {
+	if m != nil {
+		m.lifecycleEventTotal.With(prometheus.Labels{"lifecycle_event_type": lifecycleEventType}).Inc()
```
lieut-data, line 169: nit: is the `lifecycle_` prefix for the `event_type` necessary? (We don't do `change_event_change_type` above.)

lieut-data (approved):
Excellent!

server/metrics/metrics.go
```diff
@@ -78,6 +83,33 @@ func NewMetrics(info InstanceInfo) *Metrics {
...
+		Name:        "change_event_total",
+		Help:        "The total number of MS Teams change events received.",
+		ConstLabels: additionalLabels,
+	}, []string{"change_type"})
+	m.registry.MustRegister(m.changeEventTotal)
+
+	m.processedChangeEventTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemEvents,
+		Name:        "discared_change_event_total",
```
manojmalik20, line 98:
```suggestion
		Name:        "processed_change_event_total",
```

server/handlers/handlers.go
```diff
@@ -194,21 +211,21 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
...
 			// Skipping because direct/group messages are disabled
-			return
+			return discardedReasonDirectMessagesDisabled
 		}
 
 		channelID, err = ah.getChatChannelID(chat)
 		if err != nil {
 			ah.plugin.GetAPI().LogError("Unable to get original channel id", "error", err.Error())
-			return
+			return discardedReasonOther
```
manojmalik20, line 228:
```suggestion
			return discardedReasonUnableToGetTeamsData
```

lieut-data (approved):

manojmalik20 (approved):
Please fix CI.

jespino:
@manojmalik20 this is reviewed and CI is passing, and main branch is merged, feel free to merge it whenever you consider.

manojmalik20:
@jespino Will merge this after release 1.2.0 tomorrow.

PR #313, opened 2023-09-15
[MI-3521] Fixed the issue of emojis showing as html from Teams to MM

### LINKED ISSUE #309: Emoji being sent from Teams->MM showing up with html
It appears that the emoji content generated from the suggested response buttons are different

Image

was generated using similar buttons as shown here:

Image

manojmalik20:
#### Summary
- Modified the msgToPost function to add <p> tag if missing from message
- Fixed the failing unit test of converters file
- Modified the order of checking duplicate post in the handleCreatedActivity function

#### Ticket Link
Fixes #309

### DIFF
--- a/server/handlers/converters.go
+++ b/server/handlers/converters.go
@@ -38,6 +38,9 @@ func (ah *ActivityHandler) msgToPost(channelID, senderID string, msg *msteams.Me
 	var embeddedImages []msteams.Attachment
 	text, embeddedImages = ah.handleImages(text)
 	msg.Attachments = append(msg.Attachments, embeddedImages...)
+	if !strings.HasPrefix(text, "<p>") {
+		text = fmt.Sprintf("<p>%s</p>", text)
+	}
 	text = markdown.ConvertToMD(text)
 	props := make(map[string]interface{})
 	rootID := ""
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -174,6 +174,14 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 		return
 	}
 
+	// Avoid possible duplication
+	postInfo, _ := ah.plugin.GetStore().GetPostInfoByMSTeamsID(msg.ChatID+msg.ChannelID, msg.ID)
+	if postInfo != nil {
+		ah.plugin.GetAPI().LogDebug("duplicate post")
+		ah.updateLastReceivedChangeDate(msg.LastUpdateAt)
+		return
+	}
+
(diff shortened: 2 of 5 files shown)


ayusht2810 (commented):
please fix ci

ayusht2810 (approved):

server/handlers/handlers.go
```diff
@@ -174,6 +174,14 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 		return
 	}
 
+	// Avoid possible duplication
+	postInfo, _ := ah.plugin.GetStore().GetPostInfoByMSTeamsID(msg.ChatID+msg.ChannelID, msg.ID)
```
avas27JTG, line 178:
```suggestion
	postInfo, _ := ah.plugin.GetStore().GetPostInfoByMSTeamsID(msg.ChatID + msg.ChannelID, msg.ID)
```

avas27JTG (approved):

AayushChaudhary0001 (approved):

server/handlers/handlers.go
```diff
@@ -174,6 +174,14 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
...
+	if postInfo != nil {
+		ah.plugin.GetAPI().LogDebug("duplicate post")
```
Kshitij-Katiyar, line 180: Should we add additional informations like msg.ID here in the Log ?

Kshitij-Katiyar (approved):

server/handlers/handlers.go
manojmalik20: The linter doesn't allow this.

server/handlers/handlers.go
manojmalik20: I don't think there's any use for that.

server/handlers/converters.go
```diff
@@ -38,6 +38,9 @@ func (ah *ActivityHandler) msgToPost(channelID, senderID string, msg *msteams.Me
 	var embeddedImages []msteams.Attachment
 	text, embeddedImages = ah.handleImages(text)
 	msg.Attachments = append(msg.Attachments, embeddedImages...)
+	if !strings.HasPrefix(text, "<p>") {
```
jespino, line 41: I think a better way to do it to put it as an exception in the `ConvertToMD`  function, that has a check at the beginning for text that contains `<div>` or  `<p>`, you can add `<img ` or any other chunk of text that clearly means that there is some HTML to be process.

jespino (commented):
An improvement about how to handle the case here

PR #315, opened 2023-09-18
[MI-3519] Fix the issue of receiving multiple update activities from MS Teams

### LINKED ISSUE #316: Fix the issue of receiving multiple update activities from MS Teams
https://community.mattermost.com/files/3fod55pt1f8dmjs9gez8miaaiw/public?h=Ot9oQIwQzow2UKYHk9Y6lPpmuM1DpBkW_m22cikUa48

ayusht2810:
#### Summary
Fix the issue of receiving multiple update activities from MS Teams

#### Fixes
#316


server/handlers/handlers.go
```diff
@@ -43,9 +45,11 @@ type PluginIface interface {
...
 
 type ActivityHandler struct {
-	plugin PluginIface
-	queue  chan msteams.Activity
-	quit   chan bool
+	plugin               PluginIface
+	queue                chan msteams.Activity
+	quit                 chan bool
+	IgnorePluginHooksMap sync.Map
+	Mutex                sync.Mutex
```
raghavaggarwal2308, line 52: Do we need to make the casing consistent?

server/message_hooks.go
```diff
@@ -62,7 +62,14 @@ func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
...
 }
 
-func (p *Plugin) ReactionHasBeenAdded(_ *plugin.Context, reaction *model.Reaction) {
+func (p *Plugin) ReactionHasBeenAdded(c *plugin.Context, reaction *model.Reaction) {
+	updateRequired := true
+	if c.RequestId == "" {
+		if _, alreadyUpdated := p.activityHandler.IgnorePluginHooksMap.LoadAndDelete(fmt.Sprintf("reaction_%s_%s", reaction.UserId, reaction.EmojiName)); alreadyUpdated {
+			updateRequired = false
+		}
+	}
```
raghavaggarwal2308, line 71:
```suggestion
	updateRequired := true
	if c.RequestId == "" {
		_, alreadyUpdated := p.activityHandler.IgnorePluginHooksMap.LoadAndDelete(fmt.Sprintf("reaction_%s_%s", reaction.UserId, reaction.EmojiName))
	       updateRequired = !alreadyUpdated
	}
```
Maybe?

raghavaggarwal2308: Fix everywhere if changing

raghavaggarwal2308 (changes requested):

server/handlers/handlers.go
ayusht2810: discussed

raghavaggarwal2308 (approved):

manojmalik20 (approved):

server/command.go
```diff
@@ -255,6 +257,8 @@ func (p *Plugin) executeUnlinkCommand(args *model.CommandArgs) (*model.CommandRe
 		return &model.CommandResponse{}, nil
 	}
 
+	p.storeMutex.Lock()
```
manojmalik20, line 260: I think we should lock it before line no. 254.

server/message_hooks.go
```diff
@@ -144,8 +144,8 @@ func (p *Plugin) ReactionHasBeenRemoved(_ *plugin.Context, reaction *model.React
 func (p *Plugin) MessageHasBeenUpdated(c *plugin.Context, newPost, oldPost *model.Post) {
 	updateRequired := true
 	if c.RequestId == "" {
-		_, alreadyUpdated := p.activityHandler.IgnorePluginHooksMap.LoadAndDelete(fmt.Sprintf("post_%s", newPost.Id))
-		updateRequired = !alreadyUpdated
+		_, ignoreHookForReaction := p.activityHandler.IgnorePluginHooksMap.LoadAndDelete(fmt.Sprintf("post_%s", newPost.Id))
```
manojmalik20, line 147:
```suggestion
		_, ignoreHook := p.activityHandler.IgnorePluginHooksMap.LoadAndDelete(fmt.Sprintf("post_%s", newPost.Id))
```
This is `MessageHasBeenUpdated` hook. No need for reaction here.

server/monitor/subscriptions.go
```diff
@@ -178,6 +178,8 @@ func (m *Monitor) recreateChannelSubscription(subscriptionID, teamID, channelID,
 		}
 	}
 
+	m.storeMutex.Lock()
```
manojmalik20, line 181: I think we should lock it before line no. 175.

server/plugin.go
```diff
@@ -452,6 +458,8 @@ func (p *Plugin) syncUsers() {
 
 		username := "msteams_" + slug.Make(msUser.DisplayName)
 		if !isUserPresent {
+			p.storeMutex.Lock()
```
manojmalik20, line 461: Think about - 
Should we lock and unlock for each iteration of the loop OR should we lock once and unlock after the function returns?

server/handlers/handlers_test.go
```diff
@@ -736,6 +748,7 @@ func TestHandleUpdatedActivity(t *testing.T) {
 				getPostError.Id = "app.post.get.app_error"
 				mockAPI.On("GetPost", "mockMattermostID").Return(nil, getPostError).Times(1)
 				mockAPI.On("LogError", "Unable to recover the post", "postID", "mockMattermostID", "error", errors.New("unable to recover")).Times(1)
+				mockAPI.On("KVSet", lastReceivedChangeKey, mock.Anything).Return(nil).Times(1)
```
manojmalik20, line 751: Replace `mock.Anything` with the exact value. If not possible, use `mock.AnythingOfType`.

server/plugin.go
ayusht2810: I think it is fine as it is, as we don't want to lock it if the users are already present

manojmalik20 (approved):

server/command.go
```diff
@@ -249,6 +251,8 @@ func (p *Plugin) executeUnlinkCommand(args *model.CommandArgs) (*model.CommandRe
 
 	p.sendBotEphemeralPost(args.UserId, args.ChannelId, "The MS Teams channel is no longer linked to this Mattermost channel.")
 
+	p.storeMutex.Lock()
```
jespino, line 254: Why this `storeMutex` is needed? it looks like is going to generate a ton of contention.

ayusht2810: We added this to avoid concurrent reading and writing in the database. Let's say we add a reaction on Mattermost, and in the plugin hook we make an API call to add the corresponding reaction on MS Teams. Before storing the `lastUpdateAt` time in the database, we receive an update event from MS Teams, and we check for the `lastUpdateAt` time in the database for the corresponding post which might not be updated at that moment due to which we again make changes on Mattermost. To avoid these conditions we apply a mutex lock.

jespino: Databases are for handling concurrent reading and writting of the data, having this kind of database access is going to lead to a huge bottleneck in the database access.

My suggestion would be handle that case with transactions instead, but in reality, what I would expect here is to have certain degree of eventual consistency, I prefer to have some extra notifications there to handle there than this huge bottleneck.

Clearly we have to find a better way to solve the same problem, this is not something that we can deploy into production.

server/handlers/handlers.go
```diff
@@ -250,17 +247,6 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
...
-	ah.plugin.GetStoreMutex().RLock()
-	defer ah.plugin.GetStoreMutex().RUnlock()
-
-	// Avoid possible duplication
-	postInfo, _ := ah.plugin.GetStore().GetPostInfoByMSTeamsID(msg.ChatID+msg.ChannelID, msg.ID)
-	if postInfo != nil {
-		ah.plugin.GetAPI().LogDebug("duplicate post")
-		ah.updateLastReceivedChangeDate(msg.LastUpdateAt)
-		return
-	}
```
manojmalik20: Why have we removed this?

manojmalik20 (approved):

PR #325, opened 2023-09-22
[MI-3557] Fixed the issue of empty post being created on posting a large attachment on Teams

manojmalik20:
#### Summary
Fixed the issue of empty post being created on posting a large attachment on Teams

### DIFF
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -247,10 +247,14 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 	post, errorFound := ah.msgToPost(channelID, senderID, msg, chat, false)
 	ah.plugin.GetAPI().LogDebug("Post generated")
 
-	newPost, appErr := ah.plugin.GetAPI().CreatePost(post)
-	if appErr != nil {
-		ah.plugin.GetAPI().LogError("Unable to create post", "Error", appErr)
-		return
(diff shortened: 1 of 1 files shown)


server/handlers/handlers.go
```diff
@@ -247,10 +247,14 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 	post, errorFound := ah.msgToPost(channelID, senderID, msg, chat, false)
 	ah.plugin.GetAPI().LogDebug("Post generated")
 
-	newPost, appErr := ah.plugin.GetAPI().CreatePost(post)
-	if appErr != nil {
-		ah.plugin.GetAPI().LogError("Unable to create post", "Error", appErr)
-		return
+	var newPost *model.Post
+	if post.Message != "" || len(post.FileIds) > 0 {
```
ayusht2810, line 251: we are not creating a post here if the message is empty, but what if we later update that post from MS Teams to edit the post with some text?

manojmalik20: Yes, you're right. We should not include this change in the plugin.

manojmalik20:
Closing this PR as we cannot do this change. If we remove the empty post from Mattermost, there will be no MM post ID in the DB corresponding to the Teams post ID. If a user updates the post on Teams and adds some text, we won't be able to relay that change over to MM as the post will not be present on MM.

PR #336, opened 2023-10-06
[MI-3598] Handled all the nil pointer dereference possibilities in the client.go file

manojmalik20:
#### Summary
- Handled all the nil pointer dereference possibilities in the client.go file
- Modified the client struct to accept a log service instead of log function


ayusht2810 (approved):

ayusht2810 (approved):

server/msteams/client.go
```diff
@@ -547,7 +575,11 @@ func (tc *ClientImpl) UploadFile(teamID, channelID, filename string, filesize in
 		var chatFolder models.DriveItemable
 		var cErr error
 		folderName := "Microsoft Teams Chat Files"
-		itemID = *rootDirectory.GetId() + ":/" + folderName
+		rootDirectoryID := ""
+		if rootDirectory.GetId() != nil {
+			rootDirectoryID = *rootDirectory.GetId()
```
lieut-data, line 580: Here's an example of a case where I don't know what happens if we keep going below with a `rootDirectoryID == ""`. Should we just return an error in this case? Or do we have to be smarter below in realizing there is no root directory and so don't try to create the chat folder?

lieut-data (approved):
I've reviewed all the changes, and from the point of view of avoiding nil dereferences, I think this is "correct". I remain worried about the semantics of this change, but I don't have the full context to reason about the scope of that, and we should probably just iterate from here vs. pausing.

manojmalik20:
@lieut-data Our QA has tested this build and all of the features are working correctly. I acknowledge your concern about returning errors in some cases and I want to point out some things. IMO, the case where the SDK returns a `nil` id instead of an error is very rare and we weren't able to reproduce it even once. I think it was the result of some changes made in the Graph API and us not being on the latest version of the SDK. Now that we've upgraded the SDK, I don't think we'll see the same error again but still I've returned errors from some of the places. In the rest of the places, the errors will be handled and logged by the caller. We can change the handling in those places as well if we see an issue again OR we can keep this task on the board and pick this up later. I'm merging this PR for now and creating the release.

PR #338, opened 2023-10-09
prefer logging user ids over emails

lieut-data:
#### Summary
Prefer logging user ids over emails. For now, we follow the existing convention of naming the fields `TeamsUserID` and `MMUserID`.

#### Ticket Link
None.

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -406,7 +406,7 @@ func (p *Plugin) syncUsers() {
 			continue
 		}
 
-		p.API.LogDebug("Running sync user job for user with email", "email", msUser.Mail)
+		p.API.LogDebug("Running sync user job for user", "TeamsUserID", msUser.ID)
 
 		mmUser, isUserPresent := mmUsersMap[msUser.Mail]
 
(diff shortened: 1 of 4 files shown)


server/plugin_test.go
```diff
@@ -339,7 +339,7 @@ func TestSyncUsers(t *testing.T) {
 		{
 			Name: "SyncUsers: Unable to create the user",
 			SetupAPI: func(api *plugintest.API) {
-				api.On("LogError", "Unable to create new MM user during sync job", "email", "test@test.com", "error", mock.Anything).Times(1)
+				api.On("LogError", "Unable to create new MM user during sync job", "MMUserID", mock.Anything, "TeamsUserID", mock.Anything, "error", mock.Anything).Times(1)
```
lieut-data, line 342: Note that these API assertions don't actually test anything since we never call `api.AssertExpectations(t)`, but updated for completeness. (I can't add `api.AssertExpectations(t)` without failing all the tests. Ideally, we transition these away from mocks altogether.)

server/plugin.go
```diff
@@ -502,11 +502,11 @@ func (p *Plugin) syncUsers() {
 				Value:    "0",
 			}}
 			if prefErr := p.API.UpdatePreferencesForUser(newUser.Id, preferences); prefErr != nil {
-				p.API.LogError("Unable to disable email notifications for new user", "MMUserID", newUser.Id, "error", prefErr.Error())
+				p.API.LogError("Unable to disable email notifications for new user", "MMUserID", mmUser.Id, "TeamsUserID", msUser.ID, "error", prefErr.Error())
```
jespino, line 505: Is `mmUser.Id` correct here?, should it be `newUser.Id`?

server/plugin.go
```diff
@@ -502,11 +502,11 @@ func (p *Plugin) syncUsers() {
...
 			}
 
 			if err = p.store.SetUserInfo(newUser.Id, msUser.ID, nil); err != nil {
-				p.API.LogError("Unable to set user info during sync user job", "email", msUser.Mail, "error", err.Error())
+				p.API.LogError("Unable to set user info during sync user job", "MMUserID", mmUser.Id, "TeamsUserID", msUser.ID, "error", err.Error())
```
jespino, line 509: Same here, it shouldn't be `newUser.Id` instead of `mmUser.Id`?

server/plugin.go
lieut-data: Yes, you're right! I commented that the tests were broken to catch this and failed to do my own due diligence instead :)

server/plugin.go
lieut-data: Fixed!

manojmalik20 (approved):

Issue #330, opened 2023-09-26
When multiple users add reaction on Teams on any post, the reactions causes a weird behaviour on MM


AayushChaudhary0001:
If multiple users add reaction on any single post on Teams, then on MM the latest reaction added by the recent user is shown, and all of the other reaction by other users disappear.
**Steps to Reproduce:-**
1. Add a reaction on a post on Teams
2. Try to add multiple reactions through the same user on Teams on the same post.
3. Add a different reaction from any other user's account on the same post on Teams.

The latest reaction added by the second user is displayed on MM, but the reaction added by the previous user disappears.
(This issue is random and may not get reproduced every time)

In the pictures below, the reaction added on Teams were from different accounts, but the latest added reaction is synced and the rest got disappear.

![image](https://github.com/mattermost/mattermost-plugin-msteams-sync/assets/116162036/0b825cdd-748f-48a6-b602-cc077e29f40b)

![image](https://github.com/mattermost/mattermost-plugin-msteams-sync/assets/116162036/99277022-69eb-4c5c-91cf-0b3e5a1c4ec9)

Issue #327, opened 2023-09-25
Handle race conditions at various places using mutex


ayusht2810:
Handle race conditions at various places:

- Create/update activities from Mattermost to Teams: Let's say we add a reaction on Mattermost, and in the plugin hook we make an API call to add the corresponding reaction on MS Teams. Before storing the `lastUpdateAt` time in the database, we receive an update event from MS Teams, and we check for the `lastUpdateAt` time in the database for the corresponding post which might not be updated at that moment due to which we again make changes on Mattermost. This also causes an issue of reactions not syncing from MS Teams to Mattermost.
- Creating/renewing of subscriptions
- Creating synthetic users during the user sync job or via the function `getOrCreateSyntheticUser` https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/dfab068dd25eb243fc0a7cadf748d03a4c607c0f/server/handlers/getters.go#L91

[Issue 327_ Handle race conditions at various places using mutex.pdf](https://github.com/mattermost/mattermost-plugin-msteams-sync/files/12882717/Issue.327_.Handle.race.conditions.at.various.places.using.mutex.pdf)

PR #344, opened 2023-10-12
[MI-3591] Add transactions in the functions to avoid race conditions

ayusht2810:
#### Summary
Add transactions in the functions to avoid race conditions

#### Ticket Link
#327 
#330 

Duplicate of https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/340


manojmalik20 (approved):
Please fix CI before merging.

server/store/store.go
```diff
@@ -403,72 +408,125 @@ func (s *SQLStore) MattermostToTeamsUserID(userID string) (string, error) {
...
+	if txErr = row.Scan(&postInfo.MattermostID, &lastUpdateAt); txErr != nil {
+		return nil, txErr
 	}
 	postInfo.MSTeamsLastUpdateAt = time.UnixMicro(lastUpdateAt)
 	return &postInfo, nil
 }
 
 func (s *SQLStore) GetPostInfoByMattermostID(postID string) (*storemodels.PostInfo, error) {
-	query := s.getQueryBuilder().Select("msTeamsPostID, msTeamsChannelID, msTeamsLastUpdateAt").From("msteamssync_posts").Where(sq.Eq{"mmPostID": postID})
+	tx, err := s.BeginTx()
```
jespino, line 446: Same here, and probably most of the cases in this file. The transaction is only needed when you operate with multiple queries, and any of those queries is a modification query.

server/store/store.go
```diff
@@ -403,72 +408,125 @@ func (s *SQLStore) MattermostToTeamsUserID(userID string) (string, error) {
...
+		return nil, txErr
 	}
+
 	postInfo.MSTeamsLastUpdateAt = time.UnixMicro(lastUpdateAt)
 	return &postInfo, nil
 }
 
-func (s *SQLStore) SetPostLastUpdateAtByMattermostID(postID string, lastUpdateAt time.Time) error {
-	query := s.getQueryBuilder().Update("msteamssync_posts").Set("msTeamsLastUpdateAt", lastUpdateAt.UnixMicro()).Where(sq.Eq{"mmPostID": postID})
+func (s *SQLStore) SetPostLastUpdateAtByMattermostID(postID string, lastUpdateAt time.Time, tx *sql.Tx) error {
```
jespino, line 480: This is more a matter of style, I would use the `tx` as the first parameter if we want to pass the `tx` to the store methods. Because my expectation is that every query has the `tx` parameter at the beginning. but that is not the case here. Anyway, I would love to hide all the transaction complexity from the rest of the application. For example, not using transactions where there are not needed, and where they are needed use something special, like a "two steps function", something like a function that starts the operation, and returns a function that finish (or rollback) the operation depending on the parameter passed (maybe something like `err == nil`).

PR #348, opened 2023-10-13
metric: change event queue length & capacity

lieut-data:
#### Summary
Track the capacity of the change event queue (a constant, for now), as well as the length of the queue with a count for each `change_type`.

As I wrote this, I realized we have another, implicit queue of goroutines that are spawned to handle each HTTP request even if the queue is at capacity. This feels like it somewhat defeats the point of the queue, since we'll still accumulate those goroutines until we run out of memory anyway. 

I'm wondering if we should reject the HTTP request if the queue is at capacity. Thoughts? (Alternatively, we should probably at least add a metric for unqueued requests...)

#### Ticket Link
None.

### DIFF
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -88,11 +88,15 @@ func New(plugin PluginIface) *ActivityHandler {
 }
 
 func (ah *ActivityHandler) Start() {
+	// This is constant for now, but report it as a metric to future proof dashboards.
+	ah.plugin.GetMetrics().ObserveChangeEventQueueCapacity(activityQueueSize)
+
 	for i := 0; i < numberOfWorkers; i++ {
 		go func() {
 			for {
 				select {
 				case activity := <-ah.queue:
+					ah.plugin.GetMetrics().DecrementChangeEventQueueLength(activity.ChangeType)
 					ah.handleActivity(activity)
 				case <-ah.quit:
 					// we have received a signal to stop
@@ -113,6 +117,8 @@ func (ah *ActivityHandler) Stop() {
 
 func (ah *ActivityHandler) Handle(activity msteams.Activity) error {
 	ah.queue <- activity
+	ah.plugin.GetMetrics().IncrementChangeEventQueueLength(activity.ChangeType)
+
 	return nil
 }
 
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -151,7 +151,7 @@ func NewMetrics(info InstanceInfo) *Metrics {
 
 	m.changeEventQueueLength = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace:   MetricsNamespace,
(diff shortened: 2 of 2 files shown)


server/handlers/handlers.go
```diff
@@ -113,6 +117,8 @@ func (ah *ActivityHandler) Stop() {
 
 func (ah *ActivityHandler) Handle(activity msteams.Activity) error {
 	ah.queue <- activity
+	ah.plugin.GetMetrics().IncrementChangeEventQueueLength(activity.ChangeType)
```
lieut-data, line 120: I queue after, and not before, since we might block if the queue is already full. But see my larger comment in the PR description for thoughts on this.

server/metrics/metrics.go
```diff
@@ -137,6 +140,24 @@ func NewMetrics(info InstanceInfo) *Metrics {
 	})
 	m.registry.MustRegister(m.linkedChannelsTotal)
 
+	m.changeEventQueueCapacity = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemApp,
```
jespino, line 145: Why one is `MetricsSubsystemApp` and the other is `MetricsSubsystemAPI`, shouldn't be both `App` or both `API`?

server/metrics/metrics.go
```diff
@@ -137,6 +140,24 @@ func NewMetrics(info InstanceInfo) *Metrics {
...
+	})
+	m.registry.MustRegister(m.changeEventQueueCapacity)
+
+	m.changeEventQueueLength = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemAPI,
+		Name:        "change_event_queue_length",
+		Help:        "The length of the change event queue.",
+		ConstLabels: additionalLabels,
+	}, []string{"change_type"})
```
jespino, line 158: I love that we are tracking the queue by change type, that's cool

manojmalik20 (approved):

server/metrics/metrics.go
lieut-data: Good catch! Fixed :)

manojmalik20:
> I'm wondering if we should reject the HTTP request if the queue is at capacity. Thoughts? (Alternatively, we should probably at least add a metric for unqueued requests...)

@lieut-data I agree with you on this. We discussed the need for a third party queue earlier also, and I think this adds another point in favour of that. But till then, we should reject the HTTP requests so that MS Graph can retry the same change notification after a while. And we can add a metric for rejected requests as well.

lieut-data:
Sounds good, @manojmalik20! I've filed https://github.com/mattermost/mattermost-plugin-msteams-sync/issues/353 to track rejecting requests when the queue is full.

PR #349, opened 2023-10-13
reduce log volume

lieut-data:
#### Summary
Remove the major source of log spam now that we have metrics tracking these calls. We lose the visibility of the raw payload, but -- to date -- this hasn't been useful. There are also potential PII concerns.

#### Ticket Link
None.

### DIFF
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -41,6 +41,7 @@ const (
 	discardedReasonInactiveUser           = "inactive_user"
 	discardedReasonDuplicatedPost         = "duplicated_post"
 	discardedReasonAlreadyAppliedChange   = "already_applied_change"
+	discardedReasonExpiredSubscription    = "expired_subscription"
 )
 
 type PluginIface interface {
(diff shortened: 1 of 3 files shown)


server/handlers/handlers.go
```diff
@@ -151,31 +152,30 @@ func (ah *ActivityHandler) checkSubscription(subscriptionID string) bool {
...
 	case "updated":
-		ah.plugin.GetAPI().LogDebug("Handling update activity", "activity", activity)
 		discardedReason = ah.handleUpdatedActivity(activityIds)
 	case "deleted":
-		ah.plugin.GetAPI().LogDebug("Handling delete activity", "activity", activity)
 		discardedReason = ah.handleDeletedActivity(activityIds)
 	default:
 		discardedReason = discardedReasonInvalidChangeType
-		ah.plugin.GetAPI().LogWarn("Unhandled activity", "activity", activity, "error", "Not handled activity")
+		ah.plugin.GetAPI().LogError("Unsupported change type", "change_type", activity.ChangeType)
```
lieut-data, line 176: Keeping this log, and raising it to an error, because it if happens something is /probably/ broken on our side.

server/handlers/handlers.go
```diff
@@ -151,31 +152,30 @@ func (ah *ActivityHandler) checkSubscription(subscriptionID string) bool {
...
 
 func (ah *ActivityHandler) handleActivity(activity msteams.Activity) {
+	metrics := ah.plugin.GetMetrics()
+
 	activityIds := msteams.GetResourceIds(activity.Resource)
 
 	if activityIds.ChatID == "" {
 		if !ah.checkSubscription(activity.SubscriptionID) {
-			ah.plugin.GetAPI().LogDebug("The subscription is no longer active", "subscriptionID", activity.SubscriptionID)
+			metrics.ObserveProcessedChangeEventTotal(activity.ChangeType, discardedReasonExpiredSubscription)
```
lieut-data, line 161: Added visibility to this type of discard.

manojmalik20 (approved):

PR #342, opened 2023-10-10
[MI-3606] Added the feature of limiting the no. of connected users allowed through a whitelist

manojmalik20:
#### Summary
- Created a new config setting called connectedUsersAllowed
- Added validation on this config setting that its value should be greater than or equal to the no. of connected users already present in the DB
- Added the logic to validate this in the OnActivate function
- Created several store functions for different usecases
- Modified the connect command handler to check for whitelist
- Added the logic to store the user in whitelist in the oAuth redirect API handler

#### Ticket Link
Fixes #339

PR #351, opened 2023-10-16
[MI-3629] Added cluster mutexes and solved the problem of race conditions for whitelist

manojmalik20:
#### Summary
- Created store functions for locking and unlocking the whitelist
- Modified the store fuctions to use transactions for accessing the whitelist
- Created new functions for processing and validating the configuration and used them everywhere
- Fixed an existing bug in the needsConnect API
- Added the logic of locking and accessing the whitelist in the complete OAuth API
- Added the logic of reverting the OAuth if there's an error in accessing the whitelist

#### Related PRs
#342


server/api.go
```diff
@@ -399,7 +399,47 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	if whitelistSize >= a.p.getConfiguration().ConnectedUsersAllowed {
+		if err = a.p.store.SetUserInfo(mmUserID, msteamsUser.ID, nil); err != nil {
+			a.p.API.LogError("Unable to delete the OAuth token for user", "UserID", mmUserID, "Error", err.Error())
+		}
+		http.Error(w, "You cannot connect your account because the maximum limit of users allowed to connect has been reached. Please contact your system administrator.", http.StatusBadRequest)
+		return
+	}
+
+	if err := a.p.store.StoreUserInWhitelist(mmUserID, tx); err != nil {
```
ayusht2810, line 442: pass the tx argument before other arguments

server/configuration.go
```diff
@@ -37,6 +38,53 @@ type configuration struct {
...
+	c.ClientID = strings.TrimSpace(c.ClientID)
+	c.ClientSecret = strings.TrimSpace(c.ClientSecret)
+	c.EncryptionKey = strings.TrimSpace(c.EncryptionKey)
+	c.WebhookSecret = strings.TrimSpace(c.WebhookSecret)
+	c.EnabledTeams = strings.TrimSpace(c.EnabledTeams)
+}
+
+func (p *Plugin) validateConfiguration(configuration *configuration) error {
+	configuration.ProcessConfiguration()
+	if len(configuration.TenantID) == 0 {
```
ayusht2810, line 52:
```suggestion
	if configuration.TenantID == "" {
```

ayusht2810: will this not work?

ayusht2810: update everywhere if changing

server/configuration.go
```diff
@@ -37,6 +38,53 @@ type configuration struct {
...
+	}
+
+	if p.store != nil {
+		whitelistSize, err := p.store.GetSizeOfWhitelist(nil)
+		if err != nil {
+			return errors.New("failed to get the size of whitelist from the DB")
+		}
+
+		if configuration.ConnectedUsersAllowed < whitelistSize {
+			return errors.New("failed to save configuration, no. of connected users allowed should be greater than the current size of the whitelist")
```
ayusht2810, line 81:
```suggestion
			return errors.New("failed to save configuration, no. of connected users allowed should be greater than or equal to the current size of the whitelist")
```

server/store/store.go
```diff
@@ -1109,11 +1113,20 @@ func (s *SQLStore) PrefillWhitelist() error {
...
-	rows, err := query.Query()
-	if err != nil {
-		return 0, err
+	var rows *sql.Rows
+	var err error
+	if tx == nil {
+		rows, err = query.Query()
+		if err != nil {
+			return 0, err
+		}
```
ayusht2810, line 1124: we can have these outside conditions as this is repeated for the below condition as well

server/api.go
```diff
@@ -399,7 +399,47 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	defer func() {
+		if err = a.p.store.UnlockWhitelist(tx); err != nil {
+			a.p.API.LogError("Unable to unlock whitelist", "error", err.Error())
+			return
+		}
+
+		if err = a.p.store.CommitTx(tx); err != nil {
```
ayusht2810, line 421: why are we only committing? Shouldn't we rollback the transaction if we get any error during the transaction?

server/store/store.go
ayusht2810, line 1124: same

server/plugin.go
```diff
@@ -344,13 +344,8 @@ func (p *Plugin) OnActivate() error {
...
-	whitelistSize, err := p.store.GetSizeOfWhitelist()
-	if err != nil {
-		return errors.New("failed to get the size of whitelist from the DB")
-	}
-
-	if p.getConfiguration().ConnectedUsersAllowed < whitelistSize {
-		return errors.New("failed to save configuration, no. of connected users allowed should be greater than the current size of the whitelist")
+	if err := p.validateConfiguration(p.getConfiguration()); err != nil {
+		return err
 	}
```
ayusht2810, line 349: i think we can also apply a lock while prefilling the users in the whitelist

ayusht2810 (commented):
Fix ci

server/api.go
manojmalik20: Because there's no operation in the transaction that needs to be rolled back.

server/plugin.go
manojmalik20: Added a TODO comment for this.

server/store/store_test.go
```diff
@@ -1311,24 +1312,28 @@ func testLockAndUnlockWhitelist(t *testing.T, store *SQLStore, _ *plugintest.API
...
-		assert.Nil(err)
+		txErr = store.CommitTx(tx)
+		assert.Nil(txErr)
 	}()
 
 	<-ch
 	_, err = store.GetSizeOfWhitelist(nil)
 	assert.Nil(err)
 	assert.Greater(time.Since(currentTime), 1*time.Second)
+	wg.Wait()
```
ayusht2810, line 1338: why are we not applying this before fetching the size of the whitelist?

ayusht2810 (approved):

server/store/store_test.go
manojmalik20: We don't want the whole goroutine to finish when we are fetching the size of the whitelist. We are putting this at the end so that the other goroutine can finish before out test ends.

jespino (approved):
It looks good to me.

server/api.go
```diff
@@ -399,7 +399,45 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	if err = a.p.store.StoreUserInWhitelist(mmUserID); err != nil {
+	tx, err := a.p.store.BeginTx()
+	if err != nil {
+		a.p.API.LogError("Unable to begin transaction", "error", err.Error())
+		http.Error(w, "Something went wrong", http.StatusInternalServerError)
+		return
+	}
+
+	if err = a.p.store.LockWhitelist(tx); err != nil {
```
jespino, line 409: What happens if this fails? the transaction doesn't get commited  or rollbacked here right? because the defer that commits the transaction is after this error handling, right?

server/api.go
jespino: I'm not 100% sure but I think you should still close the transaction, right?

server/api.go
```diff
@@ -399,7 +399,45 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
...
+	if whitelistSize >= a.p.getConfiguration().ConnectedUsersAllowed {
+		if err = a.p.store.SetUserInfo(mmUserID, msteamsUser.ID, nil); err != nil {
+			a.p.API.LogError("Unable to delete the OAuth token for user", "UserID", mmUserID, "Error", err.Error())
+		}
+		http.Error(w, "You cannot connect your account because the maximum limit of users allowed to connect has been reached. Please contact your system administrator.", http.StatusBadRequest)
+		return
+	}
+
+	if err := a.p.store.StoreUserInWhitelist(tx, mmUserID); err != nil {
 		if !strings.Contains(err.Error(), "Duplicate entry") {
```
ayusht2810, line 441: please verify whether this check works for other databases as well

lieut-data (changes requested):
I'm hesitant to see something like `LOCK TABLE` used outside of exceptional scenarios. Is there a problem description of what race conditions we're trying to solve?

manojmalik20:
> Is there a problem description of what race conditions we're trying to solve?

@lieut-data Consider this scenario: There is only one slot left in the whitelist. Two users run the `/msteams connect` command simultaneously. The code checks the size of the whitelist and sees that one slot is empty, so it responds with the link to connect their accounts. Now, both users click on the link to connect their accounts and are going through the OAuth process, the user who completes the OAuth first should be connected and acquire the spot on the whitelist. The other should see the respective error message that the max limit of connected users has been reached and he can't connect even though his OAuth was successful. But before this PR, both users were able to connect and the whitelist size was increasing more than what is allowed.

lieut-data:
Thanks for clarifying @manojmalik20! Might we consider using our `cluster.Mutex` for this scenario instead? 

See https://github.com/mattermost/mattermost/blob/4a0cf786bece1209a918ef5c19c99b69349e88e2/server/public/pluginapi/cluster/mutex.go#L32-L41

server/plugin_test.go
```diff
@@ -499,8 +499,10 @@ func TestStart(t *testing.T) {
 		t.Run(test.Name, func(t *testing.T) {
 			p := newTestPlugin(t)
 			p.metricsService.(*metricsmocks.Metrics).On("ObserveChangeEventQueueCapacity", int64(5000)).Times(1)
-			mutex, _ := cluster.NewMutex(p.API, clusterMutexKey)
-			p.clusterMutex = mutex
+			subscriptionsMutex, _ := cluster.NewMutex(p.API, subscriptionsClusterMutexKey)
+			whitelistMutex, _ := cluster.NewMutex(p.API, whitelistClusterMutexKey)
```
ayusht2810, line 503: check error as well

server/store/sqlstore/store.go
```diff
@@ -356,32 +356,69 @@ func (s *SQLStore) MattermostToTeamsUserID(userID string) (string, error) {
 }
 
 func (s *SQLStore) GetPostInfoByMSTeamsID(chatID string, postID string) (*storemodels.PostInfo, error) {
-	query := s.getQueryBuilder().Select("mmPostID, msTeamsLastUpdateAt").From(postsTableName).Where(sq.Eq{"msTeamsPostID": postID, "msTeamsChannelID": chatID}).Suffix("FOR UPDATE")
+	tx, err := s.BeginTx()
```
ayusht2810, line 359: revert this change

server/store/sqlstore/store.go
```diff
@@ -356,32 +356,69 @@ func (s *SQLStore) MattermostToTeamsUserID(userID string) (string, error) {
...
+		return nil, txErr
 	}
-
 	postInfo.MSTeamsLastUpdateAt = time.UnixMicro(lastUpdateAt)
 	return &postInfo, nil
 }
 
 func (s *SQLStore) GetPostInfoByMattermostID(postID string) (*storemodels.PostInfo, error) {
-	query := s.getQueryBuilder().Select("msTeamsPostID, msTeamsChannelID, msTeamsLastUpdateAt").From(postsTableName).Where(sq.Eq{"mmPostID": postID}).Suffix("FOR UPDATE")
+	tx, err := s.BeginTx()
```
ayusht2810, line 394: same

server/store/sqlstore/store.go
```diff
@@ -703,13 +740,31 @@ func (s *SQLStore) DeleteSubscription(subscriptionID string) error {
 }
 
 func (s *SQLStore) GetChannelSubscription(subscriptionID string) (*storemodels.ChannelSubscription, error) {
+	tx, err := s.BeginTx()
```
ayusht2810, line 743: same

server/api.go
```diff
@@ -398,8 +398,25 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
...
+	if whitelistSize >= a.p.getConfiguration().ConnectedUsersAllowed {
+		if err = a.p.store.SetUserInfo(mmUserID, msteamsUser.ID, nil); err != nil {
+			a.p.API.LogError("Unable to delete the OAuth token for user", "UserID", mmUserID, "Error", err.Error())
+		}
+		http.Error(w, "You cannot connect your account because the maximum limit of users allowed to connect has been reached. Please contact your system administrator.", http.StatusBadRequest)
+		return
+	}
+
+	if err := a.p.store.StoreUserInWhitelist(nil, mmUserID); err != nil {
+		if !(strings.Contains(err.Error(), "Duplicate entry") || strings.Contains(err.Error(), "duplicate key value")) {
```
ayusht2810, line 419: we can add a comment here for using two different comparisons

ayusht2810 (approved):

server/store/sqlstore/store.go
```diff
@@ -989,9 +987,16 @@ func (s *SQLStore) PrefillWhitelist() error {
 	return nil
 }
 
-func (s *SQLStore) GetSizeOfWhitelist() (int, error) {
+func (s *SQLStore) GetSizeOfWhitelist(tx *sql.Tx) (int, error) {
```
lieut-data, line 990: Can we drop the `tx` now that it's no longer required?

server/store/sqlstore/store.go
```diff
@@ -1007,10 +1012,21 @@ func (s *SQLStore) GetSizeOfWhitelist() (int, error) {
 	return result, nil
 }
 
-func (s *SQLStore) StoreUserInWhitelist(userID string) error {
+func (s *SQLStore) StoreUserInWhitelist(tx *sql.Tx, userID string) error {
```
lieut-data, line 1015: Similarly, can we drop the `tx` now?

server/store/sqlstore/store.go
```diff
@@ -1007,10 +1012,21 @@ func (s *SQLStore) GetSizeOfWhitelist() (int, error) {
...
+		_, err = query.Exec()
+	} else {
+		_, err = query.RunWith(tx).Exec()
+	}
 	if err != nil {
+		// We are checking two different strings here as the error is different for MySQL and Postgres
+		if strings.Contains(err.Error(), "Duplicate entry") || strings.Contains(err.Error(), "duplicate key value") {
+			s.api.LogDebug("UserID already present in whitelist", "UserID", userID)
+			return nil
+		}
```
lieut-data, line 1028: Let's copy/paste the more robust implementation from the server: https://github.com/mattermost/mattermost/blob/e5adf25fa095e6495d730233d39a45641d01b250/server/channels/store/sqlstore/store.go#L1157-L1174

lieut-data (changes requested):
Thanks, @manojmalik20! Awesome to see this simplify things. A few follow up comments.

lieut-data (approved):
Thanks, @manojmalik20!

PR #352, opened 2023-10-17
[MI-3634] Refactored some of the code according to review comments by Jesus

manojmalik20:
#### Summary
- Removed the use of transactions from where it was not needed
- Changed the definition of some store functions by changing the order of parameters

#### Related PRs
#344

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -222,13 +222,13 @@ func (p *Plugin) executeLinkCommand(args *model.CommandArgs, parameters []string
 		}
 	}()
 
-	if txErr = p.store.SaveChannelSubscription(storemodels.ChannelSubscription{
+	if txErr = p.store.SaveChannelSubscription(tx, storemodels.ChannelSubscription{
 		SubscriptionID: channelsSubscription.ID,
 		TeamID:         channelLink.MSTeamsTeam,
 		ChannelID:      channelLink.MSTeamsChannel,
 		ExpiresOn:      channelsSubscription.ExpiresOn,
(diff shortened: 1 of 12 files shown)


ayusht2810 (approved):

jespino (approved):
This is one step froward in the right direction, I still don't see as a good aproach using the database transaction outside the store. I prefer to have other kind of abstraction on top of that.

ayusht2810:
@jespino regarding the usage of database transactions outside the store. We do this to avoid reading from the database while we are waiting for the response from MS Graph API calls.
For example:
If we are updating the Teams message from this [line](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/17092cd65bdfc29f9d3240aaa6251d0bbc026b47/server/message_hooks.go#L760), we want to lock a particular record to avoid the reading of outdated data from the handlers file at this [line](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/17092cd65bdfc29f9d3240aaa6251d0bbc026b47/server/handlers/handlers.go#L330). To do this, we start a transaction, lock the record before calling the MS Graph API, and remove the lock after the updates are done to the store, due to which the [line](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/17092cd65bdfc29f9d3240aaa6251d0bbc026b47/server/handlers/handlers.go#L330) waits for the transaction to commit and fetch the updated value. This helps us avoid extra Mattermost hook calls and MS Graph API calls.

PR #354, opened 2023-10-18
[MI-3624] Add some additional metrics

ayusht2810:
#### Summary
Added metrics for messages, reactions and files count for different sources and actions.


server/message_hooks.go
```diff
@@ -606,9 +619,11 @@ func (p *Plugin) Send(teamID, channelID string, user *model.User, post *model.Po
 		attachment, err = client.UploadFile(teamID, channelID, fileName+"_"+fileInfo.Id+fileExtension, int(fileInfo.Size), fileInfo.MimeType, bytes.NewReader(fileData), nil)
 		if err != nil {
 			p.API.LogWarn("Error in uploading file attachment to MS Teams", "error", err)
+			p.metricsService.ObserveFilesCount(actionCreated, actionSourceMattermost, isNotDirectMessage, discardedReasonUnableToUploadFileOnTeams, 1)
 			continue
 		}
 		attachments = append(attachments, attachment)
+		p.metricsService.ObserveFilesCount(actionCreated, actionSourceMattermost, isNotDirectMessage, "", 1)
```
raghavaggarwal2308, line 626: Constant for this `1`?

raghavaggarwal2308 (changes requested):

server/handlers/handlers.go
```diff
@@ -41,12 +41,24 @@ const (
...
+	actionDeleted       = "deleted"
+	reactionSetAction   = "set"
+	reactionUnsetAction = "unset"
 )
 
 type PluginIface interface {
 	GetAPI() plugin.API
 	GetStore() store.Store
-	GetMetrics() *metrics.Metrics
+	GetMetrics() metrics.Metrics
```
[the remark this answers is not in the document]
ayusht2810: it is returning an interface now instead of struct

raghavaggarwal2308 (approved):

server/handlers/attachments.go
```diff
@@ -114,19 +115,26 @@ func (ah *ActivityHandler) handleAttachments(channelID, userID, text string, msg
 		return "", nil, "", errorFound
 	}
 
+	isDirect := directMessageFalse
```
Kshitij-Katiyar, line 118:
```suggestion
	isDirectMessage := directMessageFalse
```
or 
```suggestion
	isDirectMsg := directMessageFalse
```

server/handlers/attachments.go
```diff
@@ -184,12 +196,15 @@ func (ah *ActivityHandler) handleAttachments(channelID, userID, text string, msg
...
 		if fileInfoID == "" {
+			ah.plugin.GetMetrics().ObserveFilesCount(actionCreated, actionSourceMSTeams, isDirect, discardedReasonEmptyFileID, increaseFileCountByOne)
 			continue
 		}
 		attachments = append(attachments, fileInfoID)
-		countAttachments++
-		if countAttachments == 10 {
+		ah.plugin.GetMetrics().ObserveFilesCount(actionCreated, actionSourceMSTeams, isDirect, "", increaseFileCountByOne)
+		countFileAttachments++
+		if countFileAttachments == 10 {
```
Kshitij-Katiyar, line 205: constant for this variable?

server/handlers/handlers.go
```diff
@@ -279,6 +292,11 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 	}
 
 	ah.plugin.GetAPI().LogDebug("Post created")
+	isDirect := directMessageFalse
```
Kshitij-Katiyar, line 295: same

server/handlers/handlers.go
```diff
@@ -279,6 +292,11 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
...
+	if activityIds.ChatID != "" {
+		isDirect = directMessageTrue
+	}
```
Kshitij-Katiyar, line 298: Should we create a util for these 3 lines?

server/message_hooks.go
```diff
@@ -515,9 +523,11 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post) (
 		attachment, err = client.UploadFile("", "", fileName+"_"+fileInfo.Id+fileExtension, int(fileInfo.Size), fileInfo.MimeType, bytes.NewReader(fileData), chat)
 		if err != nil {
 			p.API.LogWarn("Error in uploading file attachment to MS Teams", "error", err)
+			p.metricsService.ObserveFilesCount(actionCreated, actionSourceMattermost, directMessageTrue, discardedReasonUnableToUploadFileOnTeams, increaseFileCountByOne)
```
Kshitij-Katiyar, line 526: Any specific reason why we are using string for True and false ?

server/metrics/metrics.go
```diff
@@ -16,12 +16,31 @@ const (
...
 
+type Metrics interface {
+	ObserveAPIEndpointDuration(handler, method, statusCode string, elapsed float64)
+	ObserveConnectedUsersTotal(count int64)
+	ObserveChangeEventTotal(changeType string)
+	ObserveProcessedChangeEventTotal(changeType string, discardedReason string)
+	ObserveLifecycleEventTotal(lifecycleEventType string)
+	ObserveMessagesCount(action, source, isDirect string)
+	ObserveReactionsCount(action, source, isDirect string)
+	ObserveFilesCount(action, source, isDirect, discardedReason string, count int)
```
Kshitij-Katiyar, line 27: Why we have not used int64 here?

server/plugin.go
```diff
@@ -41,6 +41,19 @@ const (
 
 	metricsExposePort          = ":9094"
 	updateMetricsTaskFrequency = 15 * time.Minute
+
+	discardedReasonUnableToGetMMData         = "unable_to_get_mm_data"
```
Kshitij-Katiyar, line 45: I think there are some constants with the same prefix defined in some other file.

server/plugin.go
```diff
@@ -41,6 +41,19 @@ const (
...
+	discardedReasonUnableToUploadFileOnTeams = "unable_to_upload_file_on_teams"
+
+	actionSourceMattermost = "mattermost"
+	directMessageTrue      = "true"
```
Kshitij-Katiyar, line 49: Maybe this as well ?

server/plugin.go
ayusht2810: that was related to Teams data https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/f34054a67779e0885fa9febe4e94c86914759e5b/server/handlers/handlers.go#L37

server/plugin.go
ayusht2810: Yes, there are some constants being repeated because we currently don't have a common constant package. Added a TODO comment regarding the same

server/message_hooks.go
ayusht2810: Yes, the `prometheus.Labels` is a type of `map[string]string` so we need to have it as a string.

server/plugin.go
Kshitij-Katiyar: Cant we just export this variable if this does not cause import cycle. @ayusht2810

Kshitij-Katiyar (approved):

server/plugin.go
ayusht2810: removed repeated variables

server/handlers/handlers.go
```diff
@@ -41,12 +42,25 @@ const (
...
+
+	actionSourceMSTeams    = "msteams"
+	DirectMessageTrue      = "true"
+	DirectMessageFalse     = "false"
+	ActionCreated          = "created"
+	ActionUpdated          = "updated"
+	ActionDeleted          = "deleted"
+	ReactionSetAction      = "set"
+	ReactionUnsetAction    = "unset"
+	IncreaseFileCountByOne = 1
```
manojmalik20, line 57: This variable name is not making sense.

server/handlers/handlers.go
```diff
@@ -41,12 +42,25 @@ const (
 	discardedReasonInactiveUser           = "inactive_user"
 	discardedReasonDuplicatedPost         = "duplicated_post"
 	discardedReasonAlreadyAppliedChange   = "already_applied_change"
+	discardedReasonFileLimitReached       = "file_limit_reached"
+	discardedReasonEmptyFileID            = "empty_file_id"
+	discardedReasonMaxFileSizeExceeded    = "max_file_size_exceeded"
+
+	actionSourceMSTeams    = "msteams"
+	DirectMessageTrue      = "true"
+	DirectMessageFalse     = "false"
```
manojmalik20, line 51:
```suggestion
	True      = "true"
	False     = "false"
```

server/message_hooks.go
```diff
@@ -515,9 +524,11 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post) (
 		attachment, err = client.UploadFile("", "", fileName+"_"+fileInfo.Id+fileExtension, int(fileInfo.Size), fileInfo.MimeType, bytes.NewReader(fileData), chat)
 		if err != nil {
 			p.API.LogWarn("Error in uploading file attachment to MS Teams", "error", err)
+			p.metricsService.ObserveFilesCount(handlers.ActionCreated, actionSourceMattermost, handlers.DirectMessageTrue, discardedReasonUnableToUploadFileOnTeams, int64(handlers.IncreaseFileCountByOne))
```
manojmalik20, line 527: If we are not able to successfully upload the file on Teams, should we still increment the files count by one?

server/handlers/handlers.go
lieut-data: Why isn't this just a boolean, converted into `"true"` and `"false"` within the metrics layer?

server/metrics/metrics.go
```diff
@@ -16,12 +16,31 @@ const (
 	MetricsCloudInstallationLabel = "installationId"
 )
 
+type Metrics interface {
```
lieut-data, line 19: Looks like multiple PRs are all touching this interface -- lots of merge conflicts ahead!

lieut-data (changes requested):
Thanks, @ayusht2810! Just giving a quick preliminary review as it looks like there's already some feedback to address.

server/handlers/attachments.go
```diff
@@ -184,12 +196,15 @@ func (ah *ActivityHandler) handleAttachments(channelID, userID, text string, msg
...
 			continue
 		}
 		attachments = append(attachments, fileInfoID)
-		countAttachments++
-		if countAttachments == 10 {
+		ah.plugin.GetMetrics().ObserveFilesCount(ActionCreated, actionSourceMSTeams, isDirectMessage, "", int64(IncreaseFileCountByOne))
+		countFileAttachments++
+		if countFileAttachments == maxFileAttachmentsSupported {
 			ah.plugin.GetAPI().LogDebug("discarding the rest of the attachments as Mattermost supports only 10 attachments per post")
+			ah.plugin.GetMetrics().ObserveFilesCount(ActionCreated, actionSourceMSTeams, isDirectMessage, discardedReasonFileLimitReached, int64(len(msg.Attachments)-countAttachments-countFileAttachments))
```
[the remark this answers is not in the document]
ayusht2810: `countFileAttachments` is used just for counting files, whereas `countAttachments` is for other type of attachments

server/handlers/handlers.go
ayusht2810: yes, because in one case we are incrementing the file count by a custom number rather than 1.

server/message_hooks.go
ayusht2810: yes, becuase we are also adding a discarded reason which will denote the count of files discared due to a particular reason

manojmalik20 (approved):

server/metrics/metrics.go
```diff
@@ -174,73 +225,95 @@ func NewMetrics(info InstanceInfo) *Metrics {
...
+	if m != nil {
+		m.reactionsCount.With(prometheus.Labels{"action": action, "source": source, "is_direct": strconv.FormatBool(isDirectMessage)}).Inc()
+	}
+}
+
+func (m *metrics) ObserveFilesCount(action, source, discardedReason string, isDirectMessage bool, count int64) {
+	if m != nil {
+		m.filesCount.With(prometheus.Labels{"action": action, "source": source, "is_direct": strconv.FormatBool(isDirectMessage), "discarded_reason": discardedReason}).Add(float64(count))
+	}
+}
```
lieut-data, line 278: I stumbled over the `IncreaseFileCountByOne` pattern -- could we just add a `ObserveFile` and `ObserveFiles(... count)`, with the former doing an `Inc()` and the latter an `Add(count)`?

lieut-data (approved):
Thanks, @ayusht2810! A few comments, but not strictly blocking.

PR #355, opened 2023-10-18
[MI-3642] Add metrics to count upstream users

ayusht2810:
#### Summary
- Added metrics to count upstream users
- Fixed unit test cases
- Fixed issue of panic on running sync user job


raghavaggarwal2308 (approved):

server/metrics/metrics.go
```diff
@@ -27,6 +27,7 @@ type Metrics interface {
 	ObserveFilesCount(action, source, isDirect, discardedReason string, count int)
 	ObserveSyntheticUsersTotal(count int64)
 	ObserveLinkedChannelsTotal(count int64)
+	ObserveUpstreamUsersTotal(count int64)
```
Kshitij-Katiyar, line 30: Why we have chosen Total and not count here ?

server/metrics/metrics.go
```diff
@@ -202,6 +204,15 @@ func NewMetrics(info InstanceInfo) Metrics {
 	})
 	m.registry.MustRegister(m.linkedChannelsTotal)
 
+	m.upstreamUsersTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemApp,
+		Name:        "upstream_users_total",
```
Kshitij-Katiyar, line 210: constant maybe ?

server/metrics/metrics.go
```diff
@@ -286,6 +297,11 @@ func (m *metrics) ObserveLinkedChannelsTotal(count int64) {
 		m.linkedChannelsTotal.Set(float64(count))
 	}
 }
+func (m *metrics) ObserveUpstreamUsersTotal(count int64) {
+	if m != nil {
```
Kshitij-Katiyar, line 301: why do we need this check here and not when we are actually calling this ObserveUpstreamUsersTotal

server/plugin.go
```diff
@@ -468,7 +469,7 @@ func (p *Plugin) syncUsers() {
 					}
 				} else {
 					// Skip syncing of MS Teams guest user.
-					p.API.LogDebug("Skipping syncing of the guest user", "MMUserID", mmUser.Id, "TeamsUserID", msUser.ID)
+					p.API.LogDebug("Skipping syncing of the guest user", "TeamsUserID", msUser.ID)
```
Kshitij-Katiyar, line 472: why this change ?

server/metrics/metrics.go
ayusht2810: `ObserveUpstreamUsersTotal` is a metrics whereas `ObserveFilesCount` is a counter metrics

server/metrics/metrics.go
ayusht2810: this is not used anywhere else. I don't think we need constant for this currently

server/metrics/metrics.go
ayusht2810: this is fixed in PR https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/369 where we have added checks where the function is called

server/plugin.go
ayusht2810: this was causing panic in the code as the Mattermost user don't exist for this case

server/metrics/metrics.go
Kshitij-Katiyar: If I want to see all these names in 1 place all these metrics like in this case upstream_users_total, Can I do that ?

Kshitij-Katiyar: @ayusht2810

server/metrics/metrics.go
Kshitij-Katiyar: Add a same comment below these interface functions as well and segregate them as well. @ayusht2810

Kshitij-Katiyar (approved):

server/metrics/metrics.go
ayusht2810: we already have interface present, from which we can check all the metrics present in the file

server/metrics/metrics.go
```diff
@@ -17,22 +17,29 @@ const (
 )
 
 type Metrics interface {
+	// Histogram metrics
```
lieut-data, line 20: Can you elaborate on the purpose of these comments? At first glance, this isn't needful information for the client. Furthermore, grouping them by type isn't meaningful -- if we were going to reorganize, I'd propose grouping them by subsystem.

server/metrics/metrics.go
```diff
@@ -17,22 +17,29 @@ const (
...
 	ObserveLinkedChannelsTotal(count int64)
-	IncrementHTTPRequests()
-	IncrementHTTPErrors()
-	GetRegistry() *prometheus.Registry
+	ObserveUpstreamUsersTotal(count int64)
 	ObserveChangeEventQueueCapacity(count int64)
 	IncrementChangeEventQueueLength(changeType string)
 	DecrementChangeEventQueueLength(changeType string)
+
+	GetRegistry() *prometheus.Registry
```
lieut-data, line 42: I know this was already in the interface, but wondering why we need `GetRegistry` exposed at all? This largely "breaks" the point of the interface if we intend to expose a concrete implementation detail.

server/metrics/metrics.go
```diff
@@ -202,6 +210,15 @@ func NewMetrics(info InstanceInfo) Metrics {
 	})
 	m.registry.MustRegister(m.linkedChannelsTotal)
 
+	m.upstreamUsersTotal = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemApp,
+		Name:        "upstream_users_total",
```
lieut-data, line 216: `_total` is a "special" suffix in Prometheus that's specific to a counter, but as this is a gauge, we should drop it in favour of just `upstream_users`.

server/plugin.go
```diff
@@ -410,6 +410,7 @@ func (p *Plugin) syncUsers() {
 		return
 	}
 
+	p.metricsService.ObserveUpstreamUsersTotal(int64(len(msUsers)))
 	p.API.LogDebug("Count of MS Teams users", "count", len(msUsers))
```
lieut-data, line 414: Can we drop this log while we're at it, given it's now redundant? (We can also drop the `Count of MM users` below.)

server/plugin.go
lieut-data: I've seen this change in another PR. Should we pull it out from this one for clarity?

lieut-data (changes requested):
Thanks, @ayusht2810!

server/metrics/metrics.go
ayusht2810: this was needed because we want the registry at this [part](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/423b7cc1f32dafb25850550b3bfe561a0d4e5008/server/metrics/server.go#L30) while defining the metrics server, so we defined an interface method to expose it.

server/plugin.go
ayusht2810: this isn't present in the main branch currently that's why it is showing as a change in this PR. Also, the debug log we updated in this PR https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/361, was incorrect, so I have reverted that change as well.

ayusht2810: synced the PR and updated the changes

server/metrics/metrics.go
lieut-data: Proposing we rename the `Name` here as well:
```suggestion
		Name:        "upstream_users",
```

server/metrics/metrics.go
```diff
@@ -279,13 +294,18 @@ func (m *metrics) ObserveFilesCount(action, source, discardedReason string, isDi
...
 	}
 }
 
 func (m *metrics) ObserveLinkedChannelsTotal(count int64) {
 	if m != nil {
-		m.linkedChannelsTotal.Set(float64(count))
+		m.linkedChannels.Set(float64(count))
+	}
+}
+func (m *metrics) ObserveUpstreamUsersTotal(count int64) {
```
lieut-data, line 306: Not blocking, but can we drop `Total` here too?

lieut-data (approved):

manojmalik20 (approved):

lieut-data (approved):

PR #356, opened 2023-10-18
[MI-3643] Add metrics to show messages confirmed count

ayusht2810:
#### Summary
- Add metrics to show messages confirmed count
- Fixed unit test cases


Kshitij-Katiyar (approved):

server/metrics/metrics.go
```diff
@@ -177,6 +179,15 @@ func NewMetrics(info InstanceInfo) Metrics {
 	}, []string{"action", "source", "is_direct", "discarded_reason"})
 	m.registry.MustRegister(m.filesCount)
 
+	m.messagesConfirmedCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemEvents,
+		Name:        "messages_confirmed",
```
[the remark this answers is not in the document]
ayusht2810: same https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/355#discussion_r1368384826

server/handlers/handlers.go
```diff
@@ -47,6 +47,7 @@ const (
 	discardedReasonMaxFileSizeExceeded    = "max_file_size_exceeded"
 
 	actionSourceMSTeams    = "msteams"
+	ActionSourceMattermost = "mattermost"
```
lieut-data, line 50: This doesn't feel like a constant that `handlers` should define. Can we export this from `metrics` instead?

server/metrics/metrics.go
```diff
@@ -183,6 +185,15 @@ func NewMetrics(info InstanceInfo) Metrics {
...
```
lieut-data, line 191: Can we append `_total` as a suffix? I missed this in the original design document.

lieut-data (changes requested):
Thanks again, @ayusht2810!

manojmalik20 (approved):

lieut-data (approved):

Issue #350, opened 2023-10-13
Bug: Emoji reaction from Mattermost to MS Teams did not synchronize at 12:28pm PT on Oct 12, 2023


jasonblais:
Due to the confidentiality of the details, see bug report below
https://docs.google.com/document/d/11f4qaAvH8-0hx8jQmRPq-PjYSWu3BdAVbLdl-07fMeo/edit

lieut-data:
Looks likely to be a panic from `post` being `nil` on line 349 and having not checked the error there. We fetch i ton line 355

https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/370fddc72cf8f4ec6644a9a217ed5234aff645d9/server/handlers/handlers.go#L349-L355

PR #357, opened 2023-10-18
[MI-3645] Fix issue of panics on running of sync users and adding a reaction

ayusht2810:
#### Summary

- Fix panic on running of sync user job
- Fix issue 350 of panic causing on adding of the reaction on a post

#### Ticket Link
#350

### DIFF
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -352,9 +352,9 @@ func (ah *ActivityHandler) handleUpdatedActivity(activityIds msteams.ActivityIds
 					ah.plugin.GetAPI().LogError("Unable to recover the post", "postID", postInfo.MattermostID, "error", err)
 					return discardedReasonOther
 				}
-				post, err = ah.plugin.GetAPI().GetPost(postInfo.MattermostID)
-				if err != nil {
-					ah.plugin.GetAPI().LogError("Unable to find the original post after recovery", "postID", postInfo.MattermostID, "error", err.Error())
+				post, postErr = ah.plugin.GetAPI().GetPost(postInfo.MattermostID)
+				if postErr != nil {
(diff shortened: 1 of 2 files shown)


server/plugin.go
```diff
@@ -456,7 +456,7 @@ func (p *Plugin) syncUsers() {
 					}
 				} else {
 					// Skip syncing of MS Teams guest user.
-					p.API.LogDebug("Skipping syncing of the guest user", "MMUserID", mmUser.Id, "TeamsUserID", msUser.ID)
```
lieut-data, line 459: Thanks for catching this! Looks like I missed that `isUserPresent` might be `false` altogether here leaving `mmUser` as `nil`.

server/handlers/handlers.go
```diff
@@ -352,7 +352,11 @@ func (ah *ActivityHandler) handleUpdatedActivity(activityIds msteams.ActivityIds
 					ah.plugin.GetAPI().LogError("Unable to recover the post", "postID", postInfo.MattermostID, "error", err)
 					return discardedReasonOther
 				}
-				post, _ = ah.plugin.GetAPI().GetPost(postInfo.MattermostID)
+				post, err = ah.plugin.GetAPI().GetPost(postInfo.MattermostID)
+				if err != nil {
```
lieut-data, line 356: I think we need to rename as `appErr` here since the plugin API doesn't return the `error` interface but an instance of `model.AppError` making this always true. (Love the inline linter comments!)

lieut-data (changes requested):
Thanks, @ayusht2810!

server/handlers/handlers.go
ayusht2810: updated and fixed ci

lieut-data (approved):

PR #366, opened 2023-10-19
[MI-3648] Add metrics for message sync time

ayusht2810:
#### Summary

- Add metrics for message sync time
- Fixed unit test cases


jespino:
@ayusht2810 You are adding plenty of nil checks, but that nil checks were done internally in the Metrics struct, we handle the "nil" case of the metrics inside the struct pointer with nil value. But even if want to do that, I would prefer to keep that two things separated in two different PRs.

server/metrics/metrics.go
```diff
@@ -109,6 +111,18 @@ func NewMetrics(info InstanceInfo) Metrics {
...
 
+	m.messageSyncTime = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace:   MetricsNamespace,
+			Subsystem:   MetricsSubsystemApp,
+			Name:        "message_sync_time",
+			Help:        "Time to sync the messages",
+			ConstLabels: additionalLabels,
+		},
+		[]string{"source"},
```
jespino, line 122: I don't love the "source" label name, it is difficult to know exactly what it means, I would prefer something like `msteams_event_type` that is self explanatory.

jespino: Nevermind, I saw that is not what I was thinking. Ok, source I guess is ok here.

server/handlers/handlers.go
```diff
@@ -186,7 +188,7 @@ func (ah *ActivityHandler) handleActivity(activity msteams.Activity) {
 	switch activity.ChangeType {
 	case "created":
 		ah.plugin.GetAPI().LogDebug("Handling create activity", "activity", activity)
-		discardedReason = ah.handleCreatedActivity(activityIds)
+		discardedReason = ah.handleCreatedActivity(activityIds, activity.ActivityReceiveTime)
```
jespino, line 199: We are doing this only on creation, should we do it also on update and delete?

jespino: Maybe with a `action_type` label?

server/message_hooks.go
```diff
@@ -460,7 +461,7 @@ func (p *Plugin) UnsetReaction(teamID, channelID, userID string, post *model.Pos
 	return nil
 }
 
-func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post) (string, error) {
+func (p *Plugin) SendChat(srcUser string, usersIDs []string, messageCreateTime time.Time, post *model.Post) (string, error) {
```
jespino, line 472: Should we do this for other events, like update or maybe reactions?

jespino:
@lieut-data I want your opinion here about this metric. From the mattermost side I think it is ok, we are mesuring the time between we received the notification of a new post and the time where the post is finally published in MSTeams (or at least deliver accordingly to the API)

But from the Mattermost side we are measuring the time between we receive the notification and we deliver the message. That is useful information, but I would love to have also the time between it is created in MSTeams and it is created in Mattermost. Maybe we can add another metric using the same information, but adding after the creation of the post in mattermost the time between the original post craetion date (accordingly to the microsoft post data) and the time where we finally created that in mattermost.

I think both metrics are really useful.

Also, do we want to track only creations or we want also to track updates, deletes, reactions?

ayusht2810:
@jespino I have removed the commit for checking nil for metrics. Also, we added it at various places because we were getting panics on the startup of the plugion due to this [line](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/17092cd65bdfc29f9d3240aaa6251d0bbc026b47/server/handlers/handlers.go#L92) as the function was returning nil. I will be creating a separate PR for it.

Also, I have updated the logic to calculate the message sync time. Please re-review and let me know if I have to add sync time for other events as well.

cc: @lieut-data

lieut-data:
> @jespino I have removed the commit for checking nil for metrics. Also, we added it at various places because we were getting panics on the startup of the plugion due to this [line](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/17092cd65bdfc29f9d3240aaa6251d0bbc026b47/server/handlers/handlers.go#L92) as the function was returning nil. I will be creating a separate PR for it.
> 
> Also, I have updated the logic to calculate the message sync time. Please re-review and let me know if I have to add sync time for other events as well.
> 
> cc: @lieut-data

Thanks for the heads up, @ayusht2810! Do you have the panic from what was `nil` there? At first glance, it's not obvious where the failure might be, and I'd love to better understand what went wrong.

server/metrics/metrics.go
```diff
@@ -109,6 +111,18 @@ func NewMetrics(info InstanceInfo) Metrics {
 	)
 	m.registry.MustRegister(m.apiTime)
 
+	m.messageSyncTime = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace:   MetricsNamespace,
+			Subsystem:   MetricsSubsystemApp,
+			Name:        "message_sync_time",
```
lieut-data, line 118: Should this be named `messages_delay_seconds` as per https://docs.google.com/document/d/1CFGMXUg10oZeoNm_rduLoVIwi0SRiHxj529NJaYtEjc/edit? Note the trailing unit as per Prometheus naming conventions: https://prometheus.io/docs/practices/naming/.

Open to proposing a rename, but want to make sure we're aligned on the meaning first.

ayusht2810:
@lieut-data if we disable the `Enable Performance Monitoring` setting from the system console settings. Then we will receive `nil` from the function `GetMetrics` [here](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/17092cd65bdfc29f9d3240aaa6251d0bbc026b47/server/handlers/handlers.go#L92), and the plugin itself will crash.

server/handlers/handlers.go
```diff
@@ -299,6 +299,12 @@ func (ah *ActivityHandler) handleCreatedActivity(activityIds msteams.ActivityIds
 		return discardedReasonOther
 	}
 
+	var timeElapsed float64
+	if newPost != nil {
+		timeElapsed = float64(time.Since(msg.CreateAt)) / float64(time.Second)
```
lieut-data, line 304: Should we use `post.CreateAt` for consistency here? 

Also, this division doesn't make sense to me, since `time.Second = 1000 * 1000 * 1000``. If we're measuring seconds for this metric, we shouldn't need to divide at all. (But will need to do some calculations if we start using `post.CreateAt` as above.)

server/handlers/handlers.go
lieut-data: I'd be ok starting with just create so we can evaluate if this is a useful metric. The primary goal was to measure the "roundtrip time" in both directions, and while we could technically do that for all change events, it perhaps suffices to just track created posts for now as the logic is probably simpler.

lieut-data (changes requested):
Thanks @ayusht2810! A few comments inline.

> From the mattermost side I think it is ok, we are mesuring the time between we received the notification of a new post and the time where the post is finally published in MSTeams (or at least deliver accordingly to the API) But from the Mattermost side we are measuring the time between we receive the notification and we deliver the message. That is useful information, but I would love to have also the time between it is created in MSTeams and it is created in Mattermost.

@jespino, I commented to this effect inline, but wondering if we just use `post.CreateAt` and do the same math as we do for MSTeams? Or maybe I'm misunderstanding.

> Also, do we want to track only creations or we want also to track updates, deletes, reactions?

Also commented inline, but inclined to think we can start with create just to get the coarse "round trip" time. I suppose it's possible one type of activity takes longer than create, but I suspect we have enough granularity in the other metrics to debug that.

server/handlers/handlers.go
ayusht2810: We can't use `post.CreateAt` here as we are assigning the Mattermost post `createAt` time while creating a post [here](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/f34054a67779e0885fa9febe4e94c86914759e5b/server/handlers/converters.go#L63) the same as that of MS Teams posts `createAt` time to keep both the posts `createAt` time in sync.

The `time.Since()` returns the value in seconds, and by converting it to float64, we get a value in multiples of `10^9`. Similarly, `time.Second` returns value 1s, and by converting it to float64, we get it as `1e+9`. So we divide them to get the result in seconds.

jespino: The only concern that I have here is if our server time is too much out of sync with the microsoft servers. Not a big concern but this measure is going to be a bit unreliable. I like the previous one because was measuring other thing, but in a more reliable way. We were measuring before how much time it takes to deliver the message since it hits our server until it his our database. So... that is helpful too. I see value having both metrics.

server/message_hooks.go
```diff
@@ -549,6 +549,8 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post) (
 		return "", err
 	}
 
+	timeElapsed := float64((newMessage.CreateAt.UnixMilli() - post.CreateAt)) / float64(1000)
```
jespino, line 552: This is useful too, but it is again generating an unnecesary uncontrolled information. If I deliver the message to MSTeams, I don't need to use the date from MSTeams, I can use my date, using the ellapsed time.

server/handlers/handlers.go
lieut-data: > The time.Since() returns the value in seconds

Ah, right, I stumbled over `time.Since()` returning a duration which has to be converted back to seconds. Thanks for clarifying!

>  I like the previous one ...

We should pin down exactly what we want to measure with this metric. In my mind, the goal is to keep track of the "sync delay" or the timeliness of message delivery between Mattermost and MSTeams. In particular, I think we are interested in delays that originate from outside the plugin, e.g. networking delays or MSTeams falling behind. We have other metrics to track database performance which might influence when the enduser actually sees the post, but this graph would be most useful if we measure as close to the "network layer" as possible.

So for messages that originate from MSTeams, we really only have the timestamp that MSTeams gives us as well as `time.Now()`:

image

I think there's really nothing more to do that report `M1 - T0` which `time.Since` does nicely. Though one ask here is to record `time.Now` as early as possible in the request instead of after we've done any work. (Maybe even as early as the incoming HTTP request?)

For messages that originate from Mattermost, we can do something similar, but we have to measure 2x the value:

image

The ideal would be to measure `M2-M1` in this case, but of course we don't remember `M1`. So we stick with `M2-M0` as a compromise.

In practice, `M2-M0` for messages that originate from Mattermost should be roughly double `M1 - T0` for messages that originate from Teams. (There will be clock skew between MSTeams and Mattermost as well as the delta that is `M1-M0`.) So we might consider reporting `(M2-M0)/2` if we keep it in this metric. Alternatively, we could use separate metrics and report `sync_delay_seconds` and `round_trip_seconds` or some such.

Thoughts @jespino & @ayusht2810? Are there other ways of approaching this?

lieut-data (changes requested):
Requesting changes just to block until we settle on the nuances of how to measure the metric.

server/handlers/handlers.go
ayusht2810: @lieut-data this approach looks good to me, and I think we can move forward with it. I will start updating the code once @jespino agrees with the above approach as well.

ayusht2810:
@jespino what do you think of this approach https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/366#discussion_r1367797097? Should we start working on this?

lieut-data:
@ayusht2810, we're going to revisit this later and internally at https://mattermost.atlassian.net/browse/MM-55620.

PR #367, opened 2023-10-20
Metrics for the DB Store

lieut-data:
#### Summary
Replicate a version of what the server does to track DB Store metrics. This pull request includes discrete commits that refactor the existing `store` into `sqlstore` and `timerlayer` packages, as well as the containing `store` layer with the interface in question.

#### Ticket Link
None.


server/plugin.go
```diff
@@ -165,12 +167,6 @@ func (p *Plugin) start(syncSince *time.Time) {
...
 
 	if enableMetrics != nil && *enableMetrics {
-		p.metricsService = metrics.NewMetrics(metrics.InstanceInfo{
-			InstallationID: os.Getenv("MM_CLOUD_INSTALLATION_ID"),
-		})
-
-		// run metrics server to expose data
-		p.runMetricsServer()
 		// run metrics updater recurring task
 		p.runMetricsUpdaterTask(p.store, updateMetricsTaskFrequency)
```
lieut-data, line 171: But we keep only starting the updater task on `start`.

server/plugin.go
```diff
@@ -332,13 +339,14 @@ func (p *Plugin) OnActivate() error {
...
 
-		p.store = store.New(
+		store := sqlstore.New(
 			db,
 			p.apiClient.Store.DriverName(),
 			p.API,
 			func() []string { return strings.Split(p.configuration.EnabledTeams, ",") },
 			func() []byte { return []byte(p.configuration.EncryptionKey) },
 		)
+		p.store = timerlayer.New(store, p.metricsService)
```
lieut-data, line 349: Wrap the `sqlstore` with the `timerlayer`.

server/store/layer_generators/main.go
```diff
@@ -0,0 +1,232 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
lieut-data, line 1: This is mostly just a simplified version of what's in the server.

server/store/layer_generators/main.go
```diff
@@ -0,0 +1,232 @@
...
+	file.Close()
+	f, err := parser.ParseFile(fset, "", src, parser.AllErrors|parser.ParseComments)
+	if err != nil {
+		return nil, err
+	}
+
+	topLevelFunctionsToSkip := map[string]bool{
+		"BeginTx":    true,
+		"RollbackTx": true,
+		"CommitTx":   true,
```
lieut-data, line 112: We don't track these methods. (Ideally, these wouldn't be exposed outside `sqlstore` at all, but not touching that in this PR.)

server/store/storemodels/storemodels.go
```diff
@@ -2,6 +2,12 @@ package storemodels
 
 import "time"
 
+type Stats struct {
+	ConnectedUsers int64
+	SyntheticUsers int64
+	LinkedChannels int64
+}
+
```
lieut-data, line 10: Moved this to the `storemodels` even though it's not persisted to simplify the import story.

jespino (approved):
LGTM. @manojmalik20 I know you were working on things around it, so just to keep you in the loop.

server/store/layer_generators/main.go
```diff
@@ -0,0 +1,232 @@
...
+			return strings.Join(results, ", ")
+		},
+		"joinResultsForSignature": func(results []string) string {
+			if len(results) == 0 {
+				return ""
+			}
+			returns := []string{}
+			for _, result := range results {
+				switch result {
+				case "*PostReminderMetadata":
```
manojmalik20, line 153: When does this case occur?

server/store/layer_generators/timer_layer.go.tmpl
```diff
@@ -0,0 +1,52 @@
...
+	elapsed := float64(time.Since(start)) / float64(time.Second)
+        success := "false"
+        if {{$element.Results | errorToBoolean}} {
+                success = "true"
+        }
+        s.metrics.ObserveStoreMethodDuration("Store.{{$index}}", success, elapsed)
+	{{ with (genResultsVars $element.Results false ) -}}
+	return {{ . }}
+	{{- else -}}
+	}
```
manojmalik20, line 42:
```suggestion
```
Can you verify if there's a need for this `else` condition or not?

manojmalik20 (approved):

server/store/layer_generators/main.go
lieut-data: It doesn't! Good catch -- copy and paste error on my part :)

server/store/layer_generators/timer_layer.go.tmpl
lieut-data: I believe this handles the case when the store method returns nothing. I can't find a case where that happens today, but I think it makes sense for the layer generator to handle it seamlessly regardless.

manojmalik20: When I was working on this template file for the client package, I got error here. If a function is not returning anything, then this `else` condition adds an extra `}` that gives error.

lieut-data: Oh, I see! My apologies, I didn't follow the thread fully.

PR #373, opened 2023-10-23
Fix mmUser == nil panic

lieut-data:
#### Summary
Some of my recent logging changes failed to take into account the code paths in which `mmUser` was `nil`. One of those was found and fixed in v1.2.1, this fixes the other.

#### Ticket Link
None.


crspeller (approved):
I like the check that guarantees a nil panic :)

hmhealey (approved):

PR #378, opened 2023-10-25
Metrics & naming conventions

lieut-data:
#### Summary
Drop the `changeEventTotal` metric as its redundant, then rename the old `processedChangeEventTotal` to just `changeEventTotal`.

Update the various metrics to match Prometheus naming conventions. I missed doing this correctly in my original [design document](https://docs.google.com/document/d/1CFGMXUg10oZeoNm_rduLoVIwi0SRiHxj529NJaYtEjc/edit), but it would be nice to "get this right" early. Note that this is a "breaking change" relative to the current metric names, but I'll own updating the dashboard.

#### Ticket Link
None.


server/api.go
```diff
@@ -114,7 +114,6 @@ func (a *API) processActivity(w http.ResponseWriter, req *http.Request) {
 			continue
 		}
 
-		a.p.GetMetrics().ObserveChangeEventTotal(activity.ChangeType)
```
lieut-data, line 117: This is now handled by the activity handler (two metrics counting the same thing previously).

server/handlers/handlers.go
```diff
@@ -33,27 +33,12 @@ const (
...
-	discardedReasonNotUserEvent           = "no_user_event"
-	discardedReasonOther                  = "other"
-	discardedReasonDirectMessagesDisabled = "direct_messages_disabled"
-	discardedReasonInactiveUser           = "inactive_user"
-	discardedReasonDuplicatedPost         = "duplicated_post"
-	discardedReasonAlreadyAppliedChange   = "already_applied_change"
-	discardedReasonFileLimitReached       = "file_limit_reached"
-	discardedReasonEmptyFileID            = "empty_file_id"
-	discardedReasonMaxFileSizeExceeded    = "max_file_size_exceeded"
-	discardedReasonExpiredSubscription    = "expired_subscription"
```
lieut-data, line 50: Move this to be an export from the `metrics` package.

server/metrics/metrics.go
```diff
@@ -121,7 +138,7 @@ func NewMetrics(info InstanceInfo) Metrics {
 	m.pluginStartTime = prometheus.NewGauge(prometheus.GaugeOpts{
 		Namespace:   MetricsNamespace,
 		Subsystem:   MetricsSubsystemSystem,
-		Name:        "plugin_start_time",
+		Name:        "plugin_start_timestamp_seconds",
```
lieut-data, line 141: These, among the rest of the changes, are all intended to align with https://prometheus.io/docs/practices/naming/.

server/metrics/metrics.go
```diff
@@ -158,91 +175,74 @@ func NewMetrics(info InstanceInfo) Metrics {
...
-	})
-	m.registry.MustRegister(m.changeEventQueueRejectedTotal)
-
-	m.changeEventTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
-		Namespace:   MetricsNamespace,
-		Subsystem:   MetricsSubsystemEvents,
-		Name:        "change_event_total",
-		Help:        "The total number of MS Teams change events received.",
-		ConstLabels: additionalLabels,
-	}, []string{"change_type"})
```
lieut-data, line 176: No longer distinguishing "processed change events" from "change events", as the coverage was redundant.

server/metrics/metrics.go
```diff
@@ -293,26 +293,35 @@ func NewMetrics(info InstanceInfo) Metrics {
...
 	m.registry.MustRegister(m.changeEventQueueLength)
 
+	m.changeEventQueueRejectedTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemEvents,
+		Name:        "change_event_queue_rejected_total",
+		Help:        "The total number of change events rejected due to the activity queue size being full.",
+		ConstLabels: additionalLabels,
+	})
+	m.registry.MustRegister(m.changeEventQueueRejectedTotal)
```
lieut-data, line 303: Not new, just relocated.

server/metrics/metrics.go
```diff
@@ -158,91 +175,74 @@ func NewMetrics(info InstanceInfo) Metrics {
...
-		Name:        "files_count",
+		Name:        "files_total",
 		Help:        "The total number of files for different actions and sources",
 		ConstLabels: additionalLabels,
 	}, []string{"action", "source", "is_direct", "discarded_reason"})
-	m.registry.MustRegister(m.filesCount)
+	m.registry.MustRegister(m.filesTotal)
 
-	m.messagesConfirmedCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+	// TODO: Why?
```
lieut-data, line 223: Left intact for now, but it's not clear if we need this over and above `messagesTotal` above.

server/plugin.go
```diff
@@ -45,9 +45,6 @@ const (
 
 	metricsExposePort          = ":9094"
 	updateMetricsTaskFrequency = 15 * time.Minute
-
-	discardedReasonUnableToGetMMData         = "unable_to_get_mm_data"
-	discardedReasonUnableToUploadFileOnTeams = "unable_to_upload_file_on_teams"
```
lieut-data, line 50: Moved to metrics package.

wiggin77 (approved):
LGTM :+1:

lieut-data:
@agnivade, wondering if I could get your feedback on a "fix" to the Prometheus metric naming conventions we're adopting for MS Teams to make sure I'm not missing anything? This is a "breaking change" relative to the dashboard we've deployed, but I'm inclined to think it's worth getting these right now vs. forever maintaining them.

agnivade:
It's just the MSteams dashboard right? No issues with that.

server/metrics/metrics.go
```diff
@@ -333,57 +342,51 @@ func (m *metrics) ObserveConnectedUsers(count int64) {
...
-func (m *metrics) ObserveProcessedChangeEventTotal(changeType string, discardedReason string) {
+func (m *metrics) ObserveChangeEvent(changeType string, discardedReason string) {
 	if m != nil {
-		m.processedChangeEventTotal.With(prometheus.Labels{"change_type": changeType, "discarded_reason": discardedReason}).Inc()
+		m.changeEventsTotal.With(prometheus.Labels{"change_type": changeType, "discarded_reason": discardedReason}).Inc()
 	}
 }
 
-func (m *metrics) ObserveLifecycleEventTotal(lifecycleEventType string) {
+func (m *metrics) ObserveLifecycleEvent(lifecycleEventType string) {
```
agnivade, line 351: What's the reason to drop the suffix Count/Total? Is this a recommended practice as well? I am not seeing any mention of that here: https://prometheus.io/docs/practices/naming/

agnivade (approved):

server/metrics/metrics.go
lieut-data: For the actual metric, we're keeping the `_total` suffix for counters like this one:
https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/456916c3f744fb6b099696be7c83832580e1b2dc/server/metrics/metrics.go#L187-L190

In naming the API method, I opted for the more correct "ObserveThing" -- i.e. we saw one thing, and we're incrementing our count of that in a variable named `..._total`. In the examples on the linked Prometheus page, if there's an obvious unit, we use that (e.g. `_seconds`). But then for no units it gives:

> `http_requests_total` (for a unit-less accumulating count)

As most of our counters are unitless, I've converged on `_total` in most cases.

Note that for gauges, I inferred from other documentation /not/ to use `..._total` as that was meant to be reserved for counters. So in those cases, I just state the name of the thing, e.g. `synthetic_users` and `linked_channels`.

agnivade: Yeah I was talking about the API method naming. Your reasoning makes sense. :+1:

PR #380, opened 2023-10-26
support nil-metrics

lieut-data:
#### Summary
With the change to passing around an interface instead of a concrete type, it was necessary to check `metrics != nil`. Improve this by always setting `metrics` to a concrete value, even if the underlying value is `nil`. This allows all calling code to safely use `metrics` without a `nil` check.

#### Ticket Link
None.

### DIFF
--- a/server/handlers/attachments_test.go
+++ b/server/handlers/attachments_test.go
@@ -274,7 +274,7 @@ func TestHandleAttachments(t *testing.T) {
 				p.On("GetClientForApp").Return(client).Times(1)
 				p.On("GetAPI").Return(mockAPI).Times(2)
 				p.On("GetMaxSizeForCompleteDownload").Return(1).Times(1)
-				p.On("GetMetrics").Return(mockmetrics).Times(1)
+				p.On("GetMetrics").Return(mockmetrics).Maybe()
 			},
 			setupAPI: func(mockAPI *plugintest.API) {
 				mockAPI.On("GetConfig").Return(&model.Config{
@@ -324,7 +324,7 @@ func TestHandleAttachments(t *testing.T) {
 				p.On("GetClientForApp").Return(client).Once()
 				p.On("GetAPI").Return(mockAPI).Times(3)
 				p.On("GetMaxSizeForCompleteDownload").Return(1).Times(1)
-				p.On("GetMetrics").Return(mockmetrics).Times(1)
+				p.On("GetMetrics").Return(mockmetrics).Maybe()
 			},
 			setupAPI: func(mockAPI *plugintest.API) {
 				mockAPI.On("GetConfig").Return(&model.Config{
@@ -355,7 +355,7 @@ func TestHandleAttachments(t *testing.T) {
 				p.On("GetClientForApp").Return(client).Once()
 				p.On("GetAPI").Return(mockAPI)
 				p.On("GetMaxSizeForCompleteDownload").Return(1).Times(10)
-				p.On("GetMetrics").Return(mockmetrics).Times(1)
+				p.On("GetMetrics").Return(mockmetrics).Maybe()
 			},
 			setupAPI: func(mockAPI *plugintest.API) {
 				mockAPI.On("GetConfig").Return(&model.Config{
@@ -384,7 +384,7 @@ func TestHandleAttachments(t *testing.T) {
 			description: "Attachment type code snippet",
 			setupPlugin: func(p *mocksPlugin.PluginIface, mockAPI *plugintest.API, client *mocksClient.Client, store *mocksStore.Store, mockmetrics *mocksMetrics.Metrics) {
 				p.On("GetClientForApp").Return(client)
-				p.On("GetMetrics").Return(mockmetrics).Times(1)
+				p.On("GetMetrics").Return(mockmetrics).Maybe()
 			},
 			setupAPI: func(mockAPI *plugintest.API) {
 				mockAPI.On("GetConfig").Return(&model.Config{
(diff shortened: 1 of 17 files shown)


server/handlers/attachments_test.go
```diff
@@ -274,7 +274,7 @@ func TestHandleAttachments(t *testing.T) {
 				p.On("GetClientForApp").Return(client).Times(1)
 				p.On("GetAPI").Return(mockAPI).Times(2)
 				p.On("GetMaxSizeForCompleteDownload").Return(1).Times(1)
-				p.On("GetMetrics").Return(mockmetrics).Times(1)
```
lieut-data, line 277: This is a wider conversation, but in general these types of tests aren't terribly valuable. I'd love to start a conversation around "e2e tests" (all written in Go) that minimizes the use of mocks unless absolutely needed. We adopted this approach in Playbooks with great success.

server/metrics/metrics.go
```diff
@@ -99,6 +99,14 @@ type metrics struct {
...
 }
 
+// NilMetrics returns the Metrics interface with a concrete, but nil value. Since all methods on
+// the metrics type are safe to call with a nil receiver, this gives the caller a "no-op" version
+// of the metrics interface to use when metrics are disabled.
+func NilMetrics() Metrics {
+	var m *metrics
+	return m
+}
+
```
lieut-data, line 109: This is an important part of the change.

server/plugin.go
```diff
@@ -302,6 +302,10 @@ func (p *Plugin) OnActivate() error {
 
 		// run metrics server to expose data
 		p.runMetricsServer()
+	} else {
+		// Give metricsService a concrete, but nil value, making the interface itself
+		// not-nil and safe to use without `metrics != nil` checks everywhere.
+		p.metricsService = metrics.NilMetrics()
```
lieut-data, line 308: This is the other important part of this change. Another option would actually be to always initialize `metrics` but just not expose the port. Thoughts?

server/monitor/subscriptions.go
```diff
@@ -135,9 +135,8 @@ func (m *Monitor) CreateAndSaveChatSubscription(mmSubscription *storemodels.Glob
 		return
 	}
 
-	if m.metrics != nil {
-		m.metrics.ObserveSubscriptionsCount(metrics.SubscriptionConnected)
-	}
+	m.metrics.ObserveSubscriptionsCount(metrics.SubscriptionConnected)
```
jespino, line 138: Why here we are not using the `GetMetrics` trick?

server/plugin.go
```diff
@@ -363,7 +367,7 @@ func (p *Plugin) OnActivate() error {
 			func() []string { return strings.Split(p.configuration.EnabledTeams, ",") },
 			func() []byte { return []byte(p.configuration.EncryptionKey) },
 		)
-		p.store = timerlayer.New(store, p.metricsService)
+		p.store = timerlayer.New(store, p.GetMetrics())
```
jespino, line 370: Maybe we can get rid of the whole timerlayer if there is no metrics.

server/middleware.go
```diff
@@ -20,34 +20,30 @@ func (r *StatusRecorder) WriteHeader(status int) {
 
 func (a *API) metricsMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if a.p.metricsService != nil {
-			a.p.metricsService.IncrementHTTPRequests()
-			recorder := &StatusRecorder{
-				ResponseWriter: w,
-				Status:         http.StatusOK,
-			}
+		a.p.GetMetrics().IncrementHTTPRequests()
```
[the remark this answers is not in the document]
lieut-data: Yeah, when I removed this I was asking myself, "Under what world would we want to reduce this overhead and simultaneously shutoff visibility into that overhead?" And figured we probably just always want this on.

server/metrics/metrics.go
jespino: I don't fully understand why we need that trick, In theory the attribute should be already a nil interface with the concrete type defined by the attribute. I guess you already have checked that. But I don't understand the why behind.

lieut-data: Until we set `p.metricsService := NewMetrics(...)`, the interface is a `(value, type)` of `(nil, nil)`. So none of the `m != nil` safeguards apply since there's no concrete type. 

The `NilMetrics` constructor returns an interface tuple of `(nil, Metrics)` which -- combined with the `m != nil` implementation allows safe calling despite the value not being initialized.

Another approach -- less magic -- is just to instantiate with `NewMetrics()` and simply decide whether or not to expose the port via the config. So always collect metrics, unconditionally.

Thoughts?

server/monitor/subscriptions.go
lieut-data: In this case, it's just that the `Monitor` object doesn't expose a `GetMetrics`, unlike the other cases where code accesses the metrics through a pointer to the `Plugin` struct.

server/plugin.go
lieut-data: We could, but feels like additional complexity, and makes it impossible to support toggling metrics. (It's actually already not supported anyway, but I digress.)

server/metrics/metrics.go
lieut-data: @wiggin77, the `NilMetrics` is something I'm thinking is unnecessarily complexity. I'm inclined to just always return the metrics, but only conditionally expose the port. Thoughts?

wiggin77: Agreed.

server/plugin.go
wiggin77: Always providing a non-nil Metrics would be safest as things are changed/maintained by future developers, and easier to reason about.

server/metrics/metrics.go
lieut-data: Sounds good: done!

lieut-data:
@wiggin77 & @jespino, any concerns with merging this and continuing on to the other PRs building on top of this?

wiggin77 (approved):
LGTM :+1:

manojmalik20 (approved):

jespino (approved):
LGTM

PR #381, opened 2023-10-27
add discardedReason to lifecycle event metrics

lieut-data:
#### Summary
Add `discardedReason` to lifecycle event metrics and improving error handling when an invalid webhook secret occurs.

#### Ticket Link
None.

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-plugin-msteams-sync/server/metrics"
 	"github.com/mattermost/mattermost-plugin-msteams-sync/server/msteams"
 	"github.com/mattermost/mattermost-plugin-msteams-sync/server/store"
 	"github.com/mattermost/mattermost-plugin-msteams-sync/server/store/storemodels"
@@ -147,13 +148,18 @@ func (a *API) processLifecycle(w http.ResponseWriter, req *http.Request) {
 
 	a.p.API.LogDebug("Lifecycle activity request", "activities", lifecycleEvents)
 
+	errors := ""
 	for _, event := range lifecycleEvents.Value {
 		if event.ClientState != a.p.getConfiguration().WebhookSecret {
-			a.p.API.LogError("Invalid webhook secret received in lifecycle event")
+			a.p.metricsService.ObserveLifecycleEvent(event.LifecycleEvent, metrics.DiscardedReasonInvalidWebhookSecret)
+			errors += "Invalid webhook secret"
 			continue
 		}
-		a.p.GetMetrics().ObserveLifecycleEvent(event.LifecycleEvent)
 		a.p.activityHandler.HandleLifecycleEvent(event)
+	}
+	if errors != "" {
+		http.Error(w, errors, http.StatusBadRequest)
+		return
 	}
 
 	w.WriteHeader(http.StatusOK)
--- a/server/api_test.go
+++ b/server/api_test.go
@@ -388,11 +388,13 @@ func TestProcessLifecycle(t *testing.T) {
 		ExpectedResult     string
 	}{
 		{
-			Name:            "ProcessLifecycle: With validation token present",
-			SetupAPI:        func(api *plugintest.API) {},
-			SetupClient:     func(client *clientmocks.Client, uclient *clientmocks.Client) {},
-			SetupStore:      func(store *storemocks.Store) {},
-			SetupMetrics:    func(mockmetrics *metricsmocks.Metrics) {},
+			Name:        "ProcessLifecycle: With validation token present",
+			SetupAPI:    func(api *plugintest.API) {},
+			SetupClient: func(client *clientmocks.Client, uclient *clientmocks.Client) {},
+			SetupStore:  func(store *storemocks.Store) {},
+			SetupMetrics: func(mockmetrics *metricsmocks.Metrics) {
+				mockmetrics.On("ObserveLifecycleEvent", "mockLifecycleEvent", "").Times(1)
+			},
 			ValidationToken: "mockValidationToken",
 			RequestBody: `{
 				"Value": [{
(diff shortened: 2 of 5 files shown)


server/api.go
```diff
@@ -147,14 +148,19 @@ func (a *API) processLifecycle(w http.ResponseWriter, req *http.Request) {
 
 	a.p.API.LogDebug("Lifecycle activity request", "activities", lifecycleEvents)
 
+	errors := ""
 	for _, event := range lifecycleEvents.Value {
 		if event.ClientState != a.p.getConfiguration().WebhookSecret {
-			a.p.API.LogError("Invalid webhook secret received in lifecycle event")
+			a.p.metricsService.ObserveLifecycleEvent(event.LifecycleEvent, metrics.DiscardedReasonInvalidWebhookSecret)
+			errors += "Invalid webhook secret"
```
lieut-data, line 155: This errors handling strategy is currently copy pasted from the change event handler. I think we can probably improve on this, but opted not to touch that logic except to copy it here. Notably, we now return `http.StatusBadRequest` when this occurs vs. treating it as a valid event. Thoughts?

server/api.go
```diff
@@ -147,14 +148,19 @@ func (a *API) processLifecycle(w http.ResponseWriter, req *http.Request) {
 
 	a.p.API.LogDebug("Lifecycle activity request", "activities", lifecycleEvents)
 
+	errors := ""
 	for _, event := range lifecycleEvents.Value {
 		if event.ClientState != a.p.getConfiguration().WebhookSecret {
-			a.p.API.LogError("Invalid webhook secret received in lifecycle event")
+			a.p.metricsService.ObserveLifecycleEvent(event.LifecycleEvent, metrics.DiscardedReasonInvalidWebhookSecret)
```
lieut-data, line 154: Track this event as a metric, if it occurs.

server/api_test.go
```diff
@@ -388,11 +388,13 @@ func TestProcessLifecycle(t *testing.T) {
...
-			SetupAPI:        func(api *plugintest.API) {},
-			SetupClient:     func(client *clientmocks.Client, uclient *clientmocks.Client) {},
-			SetupStore:      func(store *storemocks.Store) {},
-			SetupMetrics:    func(mockmetrics *metricsmocks.Metrics) {},
+			Name:        "ProcessLifecycle: With validation token present",
+			SetupAPI:    func(api *plugintest.API) {},
+			SetupClient: func(client *clientmocks.Client, uclient *clientmocks.Client) {},
+			SetupStore:  func(store *storemocks.Store) {},
+			SetupMetrics: func(mockmetrics *metricsmocks.Metrics) {
+				mockmetrics.On("ObserveLifecycleEvent", "mockLifecycleEvent", "").Times(1)
```
lieut-data, line 396: Move the metrics setup here to make the testing easier to reason about.

server/api_test.go
```diff
@@ -430,7 +437,8 @@ func TestProcessLifecycle(t *testing.T) {
 				"ChangeType": "mockChangeType",
 				"LifecycleEvent": "mockLifecycleEvent"
 			}]}`,
-			ExpectedStatusCode: http.StatusOK,
+			ExpectedStatusCode: http.StatusBadRequest,
+			ExpectedResult:     "Invalid webhook secret\n",
```
lieut-data, line 441: Note this change from adding error handling.

server/handlers/handlers.go
```diff
@@ -115,20 +115,25 @@ func (ah *ActivityHandler) Handle(activity msteams.Activity) error {
 
 func (ah *ActivityHandler) HandleLifecycleEvent(event msteams.Activity) {
 	if !ah.checkSubscription(event.SubscriptionID) {
+		ah.plugin.GetMetrics().ObserveLifecycleEvent(event.LifecycleEvent, metrics.DiscardedReasonFailedSubscriptionCheck)
```
lieut-data, line 118: Track this discard type.

server/handlers/handlers.go
```diff
@@ -115,20 +115,25 @@ func (ah *ActivityHandler) Handle(activity msteams.Activity) error {
...
 	if event.LifecycleEvent == "reauthorizationRequired" {
 		expiresOn, err := ah.plugin.GetClientForApp().RefreshSubscription(event.SubscriptionID)
 		if err != nil {
 			ah.plugin.GetAPI().LogError("Unable to refresh the subscription", "error", err.Error())
-		} else {
-			ah.plugin.GetMetrics().ObserveSubscriptionsCount(metrics.SubscriptionRefreshed)
-			if err = ah.plugin.GetStore().UpdateSubscriptionExpiresOn(event.SubscriptionID, *expiresOn); err != nil {
-				ah.plugin.GetAPI().LogError("Unable to store the subscription new expiry date", "subscriptionID", event.SubscriptionID, "error", err.Error())
-			}
+			ah.plugin.GetMetrics().ObserveLifecycleEvent(event.LifecycleEvent, metrics.DiscardedReasonFailedToRefresh)
```
lieut-data, line 126: Here we track the failure to refresh.

server/handlers/handlers.go
```diff
@@ -121,20 +121,25 @@ func (ah *ActivityHandler) Handle(activity msteams.Activity) error {
...
-		} else {
-			ah.plugin.GetMetrics().ObserveSubscription(metrics.SubscriptionRefreshed)
-			if err = ah.plugin.GetStore().UpdateSubscriptionExpiresOn(event.SubscriptionID, *expiresOn); err != nil {
-				ah.plugin.GetAPI().LogError("Unable to store the subscription new expiry date", "subscriptionID", event.SubscriptionID, "error", err.Error())
-			}
+			ah.plugin.GetMetrics().ObserveLifecycleEvent(event.LifecycleEvent, metrics.DiscardedReasonFailedToRefresh)
+			return
+		}
+
+		ah.plugin.GetMetrics().ObserveSubscription(metrics.SubscriptionRefreshed)
```
lieut-data, line 136: Note that we observe the refreshed subscription at the same rate we observe the lifecycle event. There's a possibility one of these metrics is redundant, but keeping for now.

server/api.go
manojmalik20: According to Microsoft's documentation, we should invalidate a change notification in this case, as it means that the change notification may have originated somewhere else. See this - https://learn.microsoft.com/en-us/graph/change-notifications-delivery-webhooks?tabs=http#processing-the-change-notification

However, it's also possible that the client state received in the change notification is correct but a system administrator has changed the webhook secret in the plugin's configuration. Right now, we don't have any logic that deletes the existing subscriptions and recreates them with the new webhook secret whenever it changes. Also, we're not storing the previous values of the webhook secret anywhere. So, right now, there's no way to validate if the client state received in the change notification is one of the webhook secrets previously configured for the plugin.

So, to sum up, we have to respond with `BadRequest` as the change notification can be from somewhere else and we can't risk security. But, we have to come up with a plan to confirm the source of the change notification and how to properly validate the client state.

ayusht2810 (approved):

PR #382, opened 2023-10-31
delay metrics server start

lieut-data:
#### Summary
As part of e82ec25fc832b95b16d53d856911ac24ed088e13, I created the metrics service earlier to make it available to the DB Store initialization that occurs in `OnActivate()`. At the time, I took along the code to actually run the metrics server exposed on `:9094`, but for reasons yet unknown, exposing the port during `OnActivate` early in a k8s pod's life cycle makes it unreachable.

Delaying the server start until after `OnActivate` returns -- where it was in v1.2 -- resolves the issue. More investigation is required to understand the root cause, but this should resolve the metric scraping issues.

#### Ticket Link
None.


cpoile (approved):

lieut-data:
cc @streamer45, moving to https://github.com/mattermost/mattermost/pull/24249 will be fantastic to side-step this issue.

PR #385, opened 2023-11-06
Grafana Dashboard (Mattermost Cloud)

lieut-data:
#### Summary
Check in the most recent version of our Grafana dashboards for this plugin. As noted in the README, these resources are specific to Mattermost Cloud and would require some tweaking to be directly useful with self-managed customers. (We will likely publish such a version at some point in the future.)

#### Ticket Link
None.


cwarnermm (approved):
Thanks, @lieut-data!

wiggin77 (approved):
LGTM :+1:

PR #386, opened 2023-11-09
[MI-3709] Add logic to auto promote synthetic users on the basis of different auth services

ayusht2810:
#### Summary

- Added logic to auto-promote synthetic users on the basis of different Auth services
- Updated configuration

[screen-capture (34).webm](https://github.com/mattermost/mattermost-plugin-msteams-sync/assets/100013900/ded034dd-dae7-4048-80f8-02cc7426a489)


plugin.json
```diff
@@ -119,6 +119,50 @@
             "type": "number", 
             "help_text": "The maximum number of users that may connect their MS Teams account. Once connected, the user is added to a white-list and may reconnect at any time.",
             "default": 1000
+        },{
+            "key": "syntheticUserAuthService",
+            "display_name": "Synthetic User Auth Service",
+            "type": "dropdown",
+              "help_text": "Select the authentication service to use when creating synthetic users. This should match the service used for member user access to Mattermost. Default is 'SAML'.",
```
manojmalik20, line 126:
```suggestion
              "help_text": "Select the authentication service to use when creating synthetic users. This should match the service used for normal users. Default is 'SAML'.",
```

plugin.json
```diff
@@ -119,6 +119,50 @@
...
+                      "display_name": "SAML",
+                      "value": "saml"
+                  },
+                  {
+                      "display_name": "AD/LDAP",
+                      "value": "ldap"
+                  }
+              ]   
+        },{
+        },{
```
manojmalik20, line 139:
```suggestion
        },{
```

plugin.json
```diff
@@ -119,6 +119,50 @@
...
+                      "display_name": "id",
+                      "value": "ID"
+                  },
+                  {
+                      "display_name": "mail",
+                      "value": "Mail"
+                  },
+                  {
+                      "display_name": "userPrincipalName",
+                      "value": "UserPrincipalName"
```
manojmalik20, line 156: Aren't these opposite? I guess the display name should be value and the value should be display name for all three options.

plugin.json
```diff
@@ -119,6 +119,50 @@
...
+                      "display_name": "AD/LDAP",
+                      "value": "ldap"
+                  }
+              ]   
+        },{
+        },{
+            "key": "syntheticUserAuthData",
+            "display_name": "Synthetic User Auth Data",
+            "type": "dropdown",
+              "help_text": "Select the MS Teams user property to use as the authentication identifier. For AD/LDAP and SAML, the identifier's value should match the value provided by the Id Attribute. ",
```
manojmalik20, line 143:
```suggestion
              "help_text": "Select the MS Teams user property to use as the authentication identifier. For AD/LDAP and SAML, the identifier's value should match the value provided by the ID Attribute. ",
```

plugin.json
```diff
@@ -119,6 +119,50 @@
...
+                      "display_name": "mail",
+                      "value": "Mail"
+                  },
+                  {
+                      "display_name": "userPrincipalName",
+                      "value": "UserPrincipalName"
+                  }
+              ]         
+        },{
+            "key": "automaticallyPromoteSyntheticUsers",
```
manojmalik20, line 160: This setting should be above the other two.

server/message_hooks.go
```diff
@@ -14,13 +14,28 @@ import (
...
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/plugin"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/pkg/errors"
 	"gitlab.com/golang-commonmark/markdown"
 )
 
+func (p *Plugin) UserWillLogIn(_ *plugin.Context, user *model.User) string {
+	p.API.LogDebug("Promoting synthetic user", "UserID", user.Id)
+	if user.RemoteId != nil && *user.RemoteId != "" {
```
manojmalik20, line 26:
```suggestion
	if user.RemoteId != nil && *user.RemoteId != "" {
	    p.API.LogDebug("Promoting synthetic user", "UserID", user.Id)
```

server/plugin.go
```diff
@@ -557,6 +570,15 @@ func (p *Plugin) syncUsers() {
 				break
 			}
 		}
+
+		if (configuration.AutomaticallyPromoteSyntheticUsers && (mmUser.AuthService != configuration.SyntheticUserAuthService || mmUser.AuthData != &configuration.SyntheticUserAuthData)) && mmUser.RemoteId != nil {
```
manojmalik20, line 574: How can user's auth data be compared to `configuration.SyntheticUserAuthData`? The configuration contains field name while the user's auth data contains value. Also, why haven't we put this logic in the `if` condition at line no. 455.

lieut-data (changes requested):
Thanks, @ayusht2810! A few questions and comments below.

server/message_hooks.go
```diff
@@ -14,13 +14,28 @@ import (
...
-	"github.com/mattermost/mattermost-server/v6/plugin"
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/plugin"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/pkg/errors"
 	"gitlab.com/golang-commonmark/markdown"
 )
 
+func (p *Plugin) UserWillLogIn(_ *plugin.Context, user *model.User) string {
+	p.API.LogDebug("Promoting synthetic user", "UserID", user.Id)
```
lieut-data, line 25: Also, we should only log this if we actually promote the user -- otherwise this log gets emitted for every user that logs in.

server/msteams/client.go
```diff
@@ -1712,6 +1712,9 @@ func (tc *ClientImpl) ListUsers() ([]clientmodels.User, error) {
 	users := []clientmodels.User{}
 	err = pageIterator.Iterate(context.Background(), func(u models.Userable) bool {
 		user := clientmodels.User{}
+		if u.GetUserPrincipalName() != nil {
+			user.UserPrincipalName = strings.ToLower(*u.GetUserPrincipalName())
```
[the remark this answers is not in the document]
ayusht2810: This change was considered in favor of this [PR](https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/141)

server/plugin.go
```diff
@@ -489,18 +491,29 @@ func (p *Plugin) syncUsers() {
 		}
 
 		username := "msteams_" + slug.Make(msUser.DisplayName)
+		authData := ""
+		if configuration.AutomaticallyPromoteSyntheticUsers {
+			authData = reflect.ValueOf(msUser).FieldByName(configuration.SyntheticUserAuthData).String()
+		}
```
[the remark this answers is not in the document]
ayusht2810: We are using dropdowns for setting `AuthData` and not an input field. So the user needs to select some fixed values from dropdown. So I don't think we need to have a switch case here. Let me know your thoughts on it.

manojmalik20 (approved):

server/plugin.go
```diff
@@ -208,11 +208,13 @@ func (p *Plugin) start(syncSince *time.Time) {
 
 	if p.getConfiguration().SyncUsers > 0 {
 		p.API.LogDebug("Starting the sync users job")
-		p.syncUsers()
 
 		// Close the previous background job if exists.
 		p.stopSyncUsersJob()
 
+		// Start a syncing of users on plugin start. The below job just schedules the job to run at a given interval of time but does not run it while scheduling. To avoid this, we call the function once separately to sync the users.
```
manojmalik20, line 215:
```suggestion
		// Start syncing the users on plugin start. The below job just schedules the job to run at a given interval of time but does not run it while scheduling. To avoid this, we call the function once separately to sync the users.
```

manojmalik20 (dismissed):

lieut-data:
@lindalumitchell, heads up that this will be a key deliverable in the upcoming release.

server/plugin.go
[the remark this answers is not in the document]
lieut-data: The dropdowns prevent the UI from exposing an interface to set other values, but is there anything that prevents me from writing an arbitrary field via `config.json`? Either way, using `reflect` for a case like this just adds surface area to think through. Hard-coding the supported cases is a simpler and safer route.

server/msteams/client.go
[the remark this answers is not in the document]
lieut-data: I admit I don't immediately see the connection. Emails might be case insensitive, but are we certain this field is?

plugin.json
```diff
@@ -119,6 +119,50 @@
...
+            "display_name": "Automatically Promote Synthetic Users",
+            "type": "bool",
+            "help_text": "When true, synthetic users will be converted to members when they login for the first time.",
```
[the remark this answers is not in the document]
lieut-data: My expectation for the enduser experience is that they arrive at Mattermost, click the appropriate LDAP or SAML button to authenticate, and the server gets the signed payload from the authorization provider. At that point, it makes a decision about allowing the user to login, calls `UserWillLogIn` and the plugin promotes the user from `Email` to one of `saml` or `ldap`. But we already know at that point how the user authenticated, so it seems unnecessary to have to "configure" this.

Or am I misunderstanding the way the feature works?

server/plugin.go
lieut-data: I ended up downloading the .webm video -- couldn't watch inline -- and realized there's a part of this I wasn't expecting: changing the configuration sets all those synthetic users from `Email` to `ldap` or `saml` immediately: even before the login. This surprised me, as I thought this only happened on login, but I guess makes sense in hindsight if that's what's required to let them login in the first place.

Can you confirm?

server/plugin.go
```diff
@@ -507,6 +519,16 @@ func (p *Plugin) syncUsers() {
 
 					continue
 				}
+
+				if configuration.AutomaticallyPromoteSyntheticUsers && (mmUser.AuthService != configuration.SyntheticUserAuthService || p.isSyntheticUserAuthDataUpdated) {
```
lieut-data, line 523: I don't understand the role `isSyntheticUserAuthDataUpdated` plays here. Why isn't the condition that `mmAuthService != configuration.SyntheticUserAuthService` sufficient in this case?

server/message_hooks.go
```diff
@@ -14,13 +14,28 @@ import (
...
-	"github.com/mattermost/mattermost-server/v6/model"
-	"github.com/mattermost/mattermost-server/v6/plugin"
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/plugin"
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
 	"github.com/pkg/errors"
 	"gitlab.com/golang-commonmark/markdown"
 )
 
+func (p *Plugin) UserWillLogIn(_ *plugin.Context, user *model.User) string {
```
lieut-data, line 24: I'd like to better understand why we do this /and/ have the sync job do it. If we can promote on login, do we need the sync job to do anything, and could we simplify things accordingly?

lieut-data (changes requested):

server/plugin.go
ayusht2810: Yes, you are correct. We update the `Auth Sevice` of a user on the basis of the configurations, as it is required for them to login and get auto-promoted.

plugin.json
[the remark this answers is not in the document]
ayusht2810: According to my knowledge, when a user clicks on the LDAP or SAML button, before redirecting the user to the authorization provider, the server checks the authentication method for the user. If we don't update the authentication method to SAML or LDAP using configurations, the user will not pass the above check and will not be able to login. After the authorization provider authorizes the user, the user gets logged in, and then the `UserWillLogIn` hooks get called, after which a session is created.

server/plugin.go
ayusht2810: The condition `mm.AuthService != configuration.SyntheticUserAuthService` is not sufficient as system admin might change `Auth Data` from, say, `ID` to 'Mail', and we are capturing this change by the variable `isSyntheticUserAuthDataUpdated` as other conditions are not sufficient to capture that change.

server/message_hooks.go
ayusht2810: We need `sync job` to update users authentication methods from `email` to other auth services. I have explained the flow of user login in the [comment](https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/386#discussion_r1400218678).

server/msteams/client.go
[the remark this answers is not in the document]
ayusht2810: I tried updating the user principal name of a user on Azure Portal with the existing user principal name, but with a different casing, but I wasn't able to update the user property. So I think it is a case insensitive property, like the email.

server/message_hooks.go
lieut-data: Right, so I better understand now that the job sets all these users to the right authentication type. Can you expand on what it means to clear the `RemoteId` bit? Does that have some effect elsewhere?

server/plugin.go
lieut-data: Ah, that makes sense. Unfortunately, it also means that if this setting is changed while the server is offline (e.g. plugin.json), the job won't be guaranteed to run and update these values.

Can we instead check the `AuthData` for the user here? While we don't get `AuthData` returned from the `GetUsers` call, we do get it from a discrete `GetUser` call and can use that to sidestep `isSyntheticUserAuthDataUpdatedl` altogether.

server/message_hooks.go
ayusht2810: By clearing the `RemoteId` here, we convert a user from a remote user to a normal user.
> Does that have some effect elsewhere?

There is an edge case here: if another plugin also creates some remote user with some type of authentication method, say SAML, and we have `automaticallyPromoteSyntheticUsers` enabled, the remote users created by other plugins will also be converted to a normal user because of the above hook.

lieut-data: cc @wiggin77, as I believe we realized that this directly conflicts with the shared channels functionality. While not currently in use where we use this plugin, do you think this is a blocker or something we can revisit later?

lieut-data (dismissed):
Thanks, @ayusht2810!

manojmalik20 (dismissed):

manojmalik20 (dismissed):

lieut-data (dismissed):

lieut-data:
Tested successfully on https://msteams-sync-test.test.mattermost.cloud/, we'll complete the QA and Security reviews asynchronously.

server/message_hooks.go
wiggin77: @lieut-data we can revisit later, but this makes the plugin incompatible with any installation also using Shared Channels.  With the remoteId cleared, those users will appear to be local and will not sync.

server/message_hooks.go
```diff
@@ -14,13 +14,28 @@ import (
...
+	if user.RemoteId != nil && *user.RemoteId != "" && p.getConfiguration().AutomaticallyPromoteSyntheticUsers {
+		p.API.LogDebug("Promoting synthetic user", "UserID", user.Id)
+		user.RemoteId = nil
```
wiggin77, line 27: May be safer to keep nulls out of the database and use:
```suggestion
		*user.RemoteId = ""
```

wiggin77 (dismissed):
LGTM :+1:   One tiny nit.

lieut-data (approved):

wiggin77 (approved):

PR #389, opened 2023-11-14
Adding e2e-poc

jespino:


server/e2e/sync_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStopPluginSendMessageAndReplyToMSTeamsLinkedChannelStartPlugin(t *testing.T) {
+	t.Skip("Not implemented yet")
```
wiggin77, line 18: Is this still intended to be skipped?

server/e2e/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+var msClient msteams.Client
+var mmClient *model.Client4
+var mmClientAdmin *model.Client4
+var testCfg *TestConfig
+
+func setup(t *testing.T) {
+	if testCfg == nil {
+		data, err := os.ReadFile("testconfig.toml")
```
wiggin77, line 43: Nit:  would be nice if this error lead people to a small README that explained how to set this up.

wiggin77 (approved):
LGTM :+1:   Minor comments.

server/e2e/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+package main
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/mattermost/mattermost-plugin-msteams-sync/server/msteams"
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/pelletier/go-toml/v2"
```
lieut-data, line 13: nit: almost all our configuration files are `json`, and I wonder if we should stick with that for consistency?

lieut-data (approved):

server/e2e/sync_test.go
jespino: Yes, this is for whenever we have the sync mechanism done

server/e2e/main_test.go
jespino: Done, changed to json

PR #390, opened 2023-11-17
[MI-3749] Fix issue in disabling of sync user job

ayusht2810:
#### Summary
Fix issue in disabling of sync user job

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -284,6 +284,8 @@ func (p *Plugin) stop() {
 	if p.activityHandler != nil {
 		p.activityHandler.Stop()
 	}
+
+	p.stopSyncUsersJob()
 }
 
 func (p *Plugin) restart() {


lieut-data:
@ayusht2810, can you share context into what this fixes?

ayusht2810:
@lieut-data when we disable `Sync user` functionality from the system console by updatting its value to 0, the `OnConfigurationChange` hook of the plugin fires and updates the configuration and calls a [restart](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/7193fdebdc76b0ed5d67caa37a428e786a5b9bf5/server/configuration.go#L169) function, which simply [stops](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/7193fdebdc76b0ed5d67caa37a428e786a5b9bf5/server/plugin.go#L290C6-L290C6) some things (like monitor system, metrics server) and calls a [start](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/7193fdebdc76b0ed5d67caa37a428e786a5b9bf5/server/plugin.go#L176) function, which initialises some of the job. At this [line](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/7193fdebdc76b0ed5d67caa37a428e786a5b9bf5/server/plugin.go#L205), we check if the `Sync user` is enabled or not, as we have disabled it by updating its value to 0, we don't enter the if condition, and the sync user job doesn't stop as we are [stopping the job](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/7193fdebdc76b0ed5d67caa37a428e786a5b9bf5/server/plugin.go#L209) in the condition itself. To stop the job, we are stopping it in the stop function like other jobs as updated in PR iteltf.

manojmalik20 (approved):

lieut-data (approved):

PR #411, opened 2023-11-22
recovery.Go to stop unhandled panics in goroutines

lieut-data:
#### Summary
A `panic` inside a goroutine will crash the plugin unless `recover()`ed. Introduce a `recovery.Go` pattern that abstracts this recovery and error logging, and then apply it in all cases where the plugin should not terminate.

Some existing goroutines remain unchanged, e.g. the `start` method for the plugin without which the plugin will no longer be meaningfully running anyway.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-55498


server/monitor/subscriptions.go
```diff
@@ -35,7 +36,9 @@ func (m *Monitor) checkChannelsSubscriptions(msteamsSubscriptionsMap map[string]
 	for _, link := range links {
 		ws <- struct{}{}
 		wg.Add(1)
-		go func(link storemodels.ChannelLink) {
+
+		link := link
```
lieut-data, line 40: Looking forward to Go 1.22 when this won't be required anymore :tada:

server/recovery/goroutine.go
```diff
@@ -0,0 +1,20 @@
+package recovery
+
+import (
+	"fmt"
+	"runtime/debug"
+)
+
+// Go runs a goroutine with a recovery handler that prevents the plugin from crashing while logging
+// the panic and traceback.
+func Go(name string, logError func(msg string, keyValuePairs ...any), callback func()) {
```
lieut-data, line 10: I'll extend this later with metrics to capture crashes as well.

server/recovery/worker.go
```diff
@@ -1,11 +1,14 @@
...
 import (
 	"fmt"
 	"runtime/debug"
 )
 
-func goWithRecovery(name string, doStart func(), isQuitting func() bool, logError func(msg string, keyValuePairs ...any)) {
+// GoWorker runs a goroutine with a recovery handler that both prevents the plugin from crashing
+// in the face of a panic, and also restarts the goroutine powering a worker unless the termination
+// was expected.
+func GoWorker(name string, doStart func(), isQuitting func() bool, logError func(msg string, keyValuePairs ...any)) {
```
lieut-data, line 11: Tweaking code previously reviewed to align everything inside the `recovery` package. A discrete package was overkill, but necessary to allow import from the current sea of packages.

lieut-data (commented):
@devinbinnie, I know this is outside your normal project of focus, but figured I could solicit a server-side review from you given your increasing effort there these days. If you can, let me know your thoughts!

wiggin77 (dismissed):
LGTM :+1:

server/recovery/goroutine.go
devinbinnie, line 10: Do we not have direct access to `logError` here? Or are there cases where we would want to use a different method than what's provided by the Plugin API?

devinbinnie (approved):
LGTM! Just one question for my own knowledge :P

server/recovery/goroutine.go
lieut-data: Unfortunately, we don't have global access to the `p.API` that contains`LogError` that routes the error through the RPC protocol with the server. The ergonomics of this pattern here isn't ideal, but I'm hoping to revisit the overall strategy with logging to at least have a generic logging interface passed around vs. this oddly specific callback.

devinbinnie: Gotcha, so the plan would be to always use `logError`.

wiggin77 (approved):

PR #414, opened 2023-11-23
[MI-3766] Fixed the issue of mentions not working properly after using certificate based subscriptions

### LINKED ISSUE #391: Bug: Mentions are not working properly from Teams to MM
## Steps to reproduce the bug

1. Go to linked channel in MS Teams.
2. Mention any desired user
3. Mention all the channel users with "@channelnameonTeams".
4. Go back to MM channel.

## Expected behaviour

The desired users and complete channel should get mentioned in MM channel also.

## Actual behaviour

The desired mentioned user is getting mentioned twice, and also the complete channel is not getting mentioned.

## Screenshots (optional)

image

image

manojmalik20:
#### Summary
Fixed the issue of mentions not working properly after using certificate based subscriptions

#### Ticket Link
Fixes #391

### DIFF
--- a/server/handlers/converters.go
+++ b/server/handlers/converters.go
@@ -76,12 +76,18 @@ func (ah *ActivityHandler) msgToPost(channelID, senderID string, msg *clientmode
 
 func (ah *ActivityHandler) handleMentions(msg *clientmodels.Message) string {
 	userIDVsNames := make(map[string]string)
-	if msg.ChatID != "" {
-		for _, mention := range msg.Mentions {
+	for _, mention := range msg.Mentions {
+		if mention.UserID != "" {
 			if userIDVsNames[mention.UserID] == "" {
 				userIDVsNames[mention.UserID] = mention.MentionedText
 			} else if userIDVsNames[mention.UserID] != mention.MentionedText {
 				userIDVsNames[mention.UserID] += " " + mention.MentionedText
+			}
+		} else if mention.ConversationID != "" {
+			if userIDVsNames[mention.ConversationID] == "" {
+				userIDVsNames[mention.ConversationID] = mention.MentionedText
+			} else if userIDVsNames[mention.ConversationID] != mention.MentionedText {
+				userIDVsNames[mention.ConversationID] += " " + mention.MentionedText
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


ayusht2810 (approved):

lieut-data (changes requested):
Thanks for driving this, @manojmalik20! I understand these changes are the result of payloads received while certificate based subscriptions are enabled. Was this also broken previously? Are these changes compatible with messages received when certificate based subscriptions are disabled?

Can we add unit tests to the converters to showcase the cases we're fixing?

manojmalik20:
Hey @lieut-data , 

Yes, the changes are compatible with messages received when certificate based subscriptions are disabled. Actually, we misinterpreted the whole thing. I tested the original code with certifcate based subscriptions disabled and I found that the mentions were broken there too. Recently, the PR for certificate based subscriptions was merged so we thought that the mentions were breaking due to that, but actually MS Graph has changed the way they return mentions in a chat message recently. Due to that, our logic for parsing the mentions from MS Teams and converting them into Mattermost mentions was breaking. If we see the documentation for the [Get ChatMessage](https://learn.microsoft.com/en-us/graph/api/chatmessage-get?view=graph-rest-1.0&tabs=http), the documentation was updated on November 18th, so I think that's when the mentions started breaking.
Can we verify this in the ServiceNow test or prod environment?

I've fixed this issue now and updated the PR title.

server/handlers/converters.go
```diff
@@ -97,8 +98,12 @@ func (ah *ActivityHandler) handleMentions(msg *clientmodels.Message) string {
 		mmMention := ""
 		mention := msg.Mentions[idx]
 		idx++
-		switch {
-		case mention.UserID != "":
+		if mention.UserID != "" {
+			if userOrConversationIDVsNames[mention.UserID] != mention.MentionedText {
+				msg.Text = strings.Replace(msg.Text, fmt.Sprintf("&nbsp;<at id=\"%s\">%s</at>", fmt.Sprint(mention.ID), mention.MentionedText), "", 1)
```
ayusht2810, line 103: why are we also replacing the initial space here?

server/handlers/converters.go
```diff
@@ -75,19 +75,20 @@ func (ah *ActivityHandler) msgToPost(channelID, senderID string, msg *clientmode
 }
 
 func (ah *ActivityHandler) handleMentions(msg *clientmodels.Message) string {
-	userIDVsNames := make(map[string]string)
+	// This map is used to store user or conversation IDs vs the first word of the multi-word mention
```
ayusht2810, line 78: is it the first or last word?

server/handlers/converters.go
manojmalik20: The mention `Dummy User` on Teams should be converted to `@dummyuser` on Mattermost without a trailing space. This is done to remove that trailing space.

ayusht2810 (approved):

manojmalik20:
Hey @lieut-data , 

Yes, the changes are compatible with messages received when certificate based subscriptions are disabled. Actually, we misinterpreted the whole thing. I tested the original code with certifcate based subscriptions disabled and I found that the mentions were broken there too. Recently, the PR for certificate based subscriptions was merged so we thought that the mentions were breaking due to that, but actually MS Graph has changed the way they return mentions in a chat message recently. Due to that, our logic for parsing the mentions from MS Teams and converting them into Mattermost mentions was breaking. If we see the documentation for the [Get ChatMessage](https://learn.microsoft.com/en-us/graph/api/chatmessage-get?view=graph-rest-1.0&tabs=http), the documentation was updated on November 18th, so I think that's when the mentions started breaking.
Can we verify this in the Hercules test or prod environment?

I've fixed this issue now and updated the PR title.

lieut-data (changes requested):
Proposing we defer this from v1.4.0 and aim for the release in two weeks with v1.5.0. There's a few things I'd like to suggest we tackle here, but I'm focused elsewhere today -- will aim to provide feedback later, but for now let's hold on merging this.

lieut-data:
Will need to revisit this in v1.6.0 or later.

PR #419, opened 2023-11-24
MM-55514: count panics & recover in background jobs too

lieut-data:
#### Summary
A few changes here to the recent `recover()` handlers for various goroutines:
* add a `msteams_connect_system_plugin_goroutine_failures_total` counter to expose failing goroutines in Grafana
* simplify and isolate the `recovery` package to be more idiomatic
* add `recover()` handling for the final two remaining background jobs invoked by `cluster.Schedule`.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-55724
Fixes: https://mattermost.atlassian.net/browse/MM-55514


lieut-data:
@cpoile & @crspeller, I know this is outside your area of focus at the moment, but would love your feedback on these last set of changes to a `recovery` package used in this plugin to help guard against goroutine panics killing the entire plugin. I imagine this is something we can lift to the pluginapi if it proves useful in the longrun.

server/recovery/handler.go
```diff
@@ -0,0 +1,57 @@
+package recovery
+
+import (
+	"fmt"
+	"runtime/debug"
+)
+
+// Wrap wraps a callback with a handler that recovers from a panic, logging and tracking metrics.
+func Wrap(name string, logError func(msg string, keyValuePairs ...any), metrics Metrics, callback func()) func() {
```
crspeller, line 9: I think it should be the functions responsibility to deal with panics and not the caller. Do I just wrap every function call in this?

server/plugin.go
```diff
@@ -408,7 +408,7 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 
-	recovery.Go("prefill_whitelist", p.API.LogError, func() {
+	recovery.Go("prefill_whitelist", p.API.LogError, p.GetMetrics(), func() {
```
crspeller, line 411: This is an example of why I don't understand the strategy here. Why is this store call special and needs a recovery, but the next p.start call does not? Why are they in goroutines at all? Can we really recover from whatever PrefillWhitelist is doing if it panics?

crspeller (changes requested):
Not sure I understand the overall strategy here. I think we may be handling panics in the wrong places.

lieut-data:
@crspeller, I realize this incremental review lacked sufficient context: for clarity, I've pushed a temporary branch that showcases the last few PRs plus this one against main: https://github.com/mattermost/mattermost-plugin-msteams-sync/compare/main...temp-review.

The core idea to tackle is uncaught panics inside spawned goroutines. This is easier to see in the linked branch diff, but there are a few cases to consider:
* previous calls to `go blah()` which now become `recovery.Go`
* previous calls to `cluster.Schedule` with a callback which now gets wrapped with `recovery.Wrap`
* the activity handler queue that processes change notifications

The first case using `recovery.Go` is just syntactic sugar for writing this by hand:
```
go func() {
    defer func() {
        if r := recover(); r != nil {
            logError(...)
            reportMetrics(...)
        }
    }

    blah()
}
```

The second case for `recovery.Wrap` is for when we aren't directly responsible for invoking the callback. Some other goroutine -- in our case, the pluginapi package managing the cluster scheduler -- runs our callback in a goroutine and (intentionally) doesn't guard against panics itself.

The last case is a pattern lifted directly out of our [web hub](https://github.com/mattermost/mattermost/blob/5e94af13022d16dd29c079bc2a68c84a5bac6e9e/server/channels/app/platform/web_hub.go#L370-L375) processing code, but slightly generalized and additionally unit tested in this project. We don't want the activity handler's worker to really die in that case, but just restart itself as often as needed.

Does this help add clarity?

server/recovery/handler.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Go wraps the given callback with a handler to recover from a panic, and then invokes the
+// callback in a goroutine.
+func Go(name string, logError func(msg string, keyValuePairs ...any), metrics Metrics, callback func()) {
+	go Wrap(name, logError, metrics, callback)()
+}
+
+// GoWorker wraps and invokes the given callback in a goroutine, automatically restarting in a new
+// goroutine on any unrecovered panic or unexpected termination.
+func GoWorker(name string, logError func(msg string, keyValuePairs ...any), metrics Metrics, isQuitting func() bool, callback func()) {
```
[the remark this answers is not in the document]
lieut-data: Provided more detail as a high-level comment, but let me also link to the original PR that gave context to where this came from: https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/394

server/recovery/handler.go
lieut-data: It's definitely the functions responsibility and not the caller, but as we have to replicate the same pattern in half a dozen places, this wrapper just makes that easier. I still think it's the function doing the work here: just a slightly fatter function being given to the caller.

server/plugin.go
lieut-data: This one is definitely up for debase as to whether it needed a goroutine in the first place. I think the only reason it's run asynchronously is because of the lock, but even then it's debatable. For now, I've left the underlying semantics of the code unchanged, but we're definitely reviewing all of this :)

server/recovery/handler.go
crspeller: It's not the function doing the work because I as the caller have to remember to wrap it. If we know the function has to handle panics we should handle them within that function.
I would rather see the recovery code in the function with maybe some explanation of why it exists rather than having every call wrapped. 
What happens when someone comes along and calls the function we are wrapping in another place and does not wrap it?

crspeller:
I am not sure I want the sugar because panic handling should be explicit and clear. 

> Some other goroutine -- in our case, the pluginapi package managing the cluster scheduler -- runs our callback in a goroutine and (intentionally) doesn't guard against panics itself.

Sounds like the function we are passing should handle the recovery?

Just to be clear, the alternate strategy I am suggesting is handling panics directly in functions that are susceptible to panics and be clear why it's there. Otherwise I think we will end up with a bunch of calls to `recovery.Go` and `recovery.Wrap` littered around seemingly randomly.

lieut-data:
Got it, thank you for that pushback. I took the opportunity to inline the recovery handling (and simplify the logging and metrics at the same time), with the changes updated in the https://github.com/mattermost/mattermost-plugin-msteams-sync/compare/main...temp-review. Is that more idiomatic?

Note that I opted to keep the `GoWorker` paradigm, though renamed and used internally within the `handlers` package, as I found it valuable to unit test the functional logic there.

crspeller:
@lieut-data That is exactly what I was suggesting. I think it is more idiomatic. 

I agree with the decision to keep the helper function for the workers. The `startWorker` function makes sense.

server/command.go
```diff
@@ -333,9 +333,16 @@ func (p *Plugin) executeShowLinksCommand(args *model.CommandArgs) (*model.Comman
 	}
 
 	p.sendBotEphemeralPost(args.UserId, args.ChannelId, commandWaitingMessage)
-	recovery.Go("send_links_with_details", p.API.LogError, func() {
+	go func() {
+		defer func() {
```
crspeller, line 337: This could be moved inside the `SendLinksWithDetails` function.

server/handlers/attachments.go
```diff
@@ -226,9 +226,16 @@ func (ah *ActivityHandler) GetFileFromTeamsAndUploadToMM(downloadURL string, cli
 		return ""
 	}
 
-	recovery.Go("get_file_content_stream", ah.plugin.GetAPI().LogError, func() {
+	go func() {
+		defer func() {
```
crspeller, line 230: We could move this inside the client and return an error.

crspeller (changes requested):
Couple of optional improvements. Really just wanting to get the recovery close to where the actual problems are. Since a function taking out the main goroutine with a direct call is just as bad as it happening in another goroutine.
Other places might be able to do this but it's not as obvious.

server/command.go
lieut-data: This one makes sense -- pushed! -- as the goroutine and implementation are all co-located.

server/handlers/attachments.go
lieut-data: This one I'm less convinced about, as the decision to run in a goroutine is the caller's. Would I have to add `recover()` to all client methods in case the caller decides to invoke in a goroutine?

crspeller: Your right, you could pass a pipe that has a goroutine waiting on read. Then the parent goroutine would handle recovery (or not :P)

crspeller (approved):

cpoile (approved):
Not much I can add here, looks good! (I was a little thrown by the isQuitting pattern, but that's just context I was lacking)

PR #421, opened 2023-11-27
Register user settings

larkox:
#### Summary
Register user settings, as continuation to https://github.com/mattermost/mattermost/pull/25517
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55624

### DIFF
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -23,6 +23,33 @@ export default class Plugin {
 
         registry.registerAdminConsoleCustomSetting('appManifestDownload', MSTeamsAppManifestSetting);
         registry.registerAdminConsoleCustomSetting('ConnectedUsersReportDownload', ListConnectedUsers);
+        registry.registerUserSettings?.({
+            id: 'com.mattermost.msteams-sync',
+            icon: 'https://upload.wikimedia.org/wikipedia/commons/thumb/c/c9/Microsoft_Office_Teams_%282018%E2%80%93present%29.svg/1200px-Microsoft_Office_Teams_%282018%E2%80%93present%29.svg.png',
+            uiName: 'MS Teams Sync',
+            sections: [{
+                settings: [{
+                    name: 'primary_platform',
+                    options: [
+                        {
+                            text: 'Mattermost will be my primary platform',
+                            value: 'mm',
+                            helpText: 'You will need to disable notifications in Microsoft Teams to avoid duplicates. **[Learn more](http://google.com)**'
+                        },
+                        {
+                            text: 'Microsoft Teams will be my primary platform',
+                            value: 'teams',
+                            helpText: 'Notifications in Mattermost will be muted for linked channels and DMs to prevent duplicates. Unread statuses in linked channels and DMs will also be disabled in Mattermost. **[Learn more](http://google.com)**',
+                        },
+                    ],
+                    type: 'radio',
+                    default: 'mm',
+                    helpText: 'Note: Unread statuses for linked channels and DMs will not be synced between Mattermost & Microsoft Teams.',
+                }],
+                title: 'Primary platform for communication',
+            }]
+            
+        });
     }
 }
 
--- a/webapp/src/types/mattermost-webapp/index.d.ts
+++ b/webapp/src/types/mattermost-webapp/index.d.ts
@@ -2,6 +2,41 @@ export interface PluginRegistry {
     registerPostTypeComponent(typeName: string, component: React.ElementType)
     registerRootComponent(component: React.ElementType)
     registerAdminConsoleCustomSetting(key: string, component: React.ElementType)
+    registerUserSettings?(setting: PluginConfiguration)
 
     // Add more if needed from https://developers.mattermost.com/extend/plugins/webapp/reference
 }
+
+export type PluginConfiguration = {
+    id: string;
+    uiName: string;
+    icon?: string;
+    sections: PluginConfigurationSection[];
+}
+
+export type PluginConfigurationSection = {
+    settings: PluginConfigurationSetting[];
+    title: string;
+    onSubmit?: (changes: {[name: string]: string}) => void;
+}
+
+export type BasePluginConfigurationSetting = {
+    name: string;
+    title?: string;
+    helpText?: string;
+    default?: string;
+}
+
+export type PluginConfigurationRadioSetting = BasePluginConfigurationSetting & {
+    type: 'radio';
+    default: string;
+    options: PluginConfigurationRadioSettingOption[];
+}
+
+export type PluginConfigurationRadioSettingOption = {
+    value: string;
+    text: string;
+    helpText?: string;
+}
+
+export type PluginConfigurationSetting = PluginConfigurationRadioSetting


webapp/src/index.tsx
```diff
@@ -16,13 +16,40 @@ export default class Plugin {
...
+                    options: [
+                        {
+                            text: 'Mattermost will be my primary platform',
+                            value: 'mm',
+                            helpText: 'You will need to disable notifications in Microsoft Teams to avoid duplicates. **[Learn more](http://google.com)**',
+                        },
+                        {
+                            text: 'Microsoft Teams will be my primary platform',
+                            value: 'teams',
+                            helpText: 'Notifications in Mattermost will be muted for linked channels and DMs to prevent duplicates. Unread statuses in linked channels and DMs will also be disabled in Mattermost. **[Learn more](http://google.com)**',
```
manojmalik20, line 43: I think we should also mention email notifications and include GMs too.

webapp/src/types/mattermost-webapp/index.d.ts
```diff
@@ -1,7 +1,42 @@
...
+    default: string;
+    options: PluginConfigurationRadioSettingOption[];
+}
+
+export type PluginConfigurationSetting = PluginConfigurationRadioSetting
+
+export type PluginConfigurationSection = {
+    settings: PluginConfigurationSetting[];
+    title: string;
+    onSubmit?: (changes: {[name: string]: string}) => void;
```
manojmalik20, line 25:
```suggestion
    onSubmit?: (changes: Record<string, string>) => void;
```

webapp/src/types/mattermost-webapp/index.d.ts
```diff
@@ -1,7 +1,42 @@
...
+
 export interface PluginRegistry {
     registerPostTypeComponent(typeName: string, component: React.ElementType)
     registerRootComponent(component: React.ElementType)
     registerAdminConsoleCustomSetting(key: string, component: React.ElementType)
+    registerUserSettings?(setting: PluginConfiguration)
 
     // Add more if needed from https://developers.mattermost.com/extend/plugins/webapp/reference
-}
+}
```
manojmalik20, line 42: Add EOF.

larkox:
@abhijit-singh adding you to the review so you can get me the final texts and links for the settings.

webapp/src/index.tsx
```diff
@@ -16,13 +16,40 @@ export default class Plugin {
...
 
         // @see https://developers.mattermost.com/extend/plugins/webapp/reference/
         this.enforceConnectedAccountId = registry.registerRootComponent(EnforceConnectedAccountModal);
 
         registry.registerAdminConsoleCustomSetting('appManifestDownload', MSTeamsAppManifestSetting);
         registry.registerAdminConsoleCustomSetting('ConnectedUsersReportDownload', ListConnectedUsers);
+        registry.registerUserSettings?.({
+            id: manifest.id,
+            icon: `${serverRoute}/plugins/${manifest.id}/public/msteams-sync-icon.svg`,
+            uiName: 'MS Teams Sync',
```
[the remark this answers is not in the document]
larkox: There are two ways to handle this. Either we add here `manifest.name`, and let it be the same as the manifest, or we change it in the client side, so the client gets the manifest and use the name there.

The first option still leaves liberty to the plugin to use a different team if they desire, but at the same time reduces the maintenance cost because we would need to change the name only in one place.

The second option removes the option of changing the name from the plugin. It will always use the one in the manifest. I don't dislike it, but I thought we wanted to give this piece of freedom to the plugin. Worst case scenario, it can be added later.

Regarding the 32 char limit, that one is about the id and creating the preference category. The UI limit is not really about number of characters, but space (it is affected by many variables, like the width of each letter (a `W` is wider than a `i`)). I don't expect often the plugin developer to have an overcomplicated name they want to simplify in the settings. 0/5 on this.

abhijit-singh (commented):
Thanks @larkox! A couple of small requests:
1. Can we update the text to match [this design](https://www.figma.com/file/BDolYCzc6CMCgVMa3U1p9B/MS-Teams-Sync---Choose-Primary-Platform?type=design&node-id=42-12857&mode=design&t=vvyCXjtkgHu0nzlb-4)? The help text seems different right now. 
image
2. The icon in the left sidebar seems vertically misaligned. Can we make the spacing above and below it equal?

webapp/src/index.tsx
```diff
@@ -27,20 +27,20 @@ export default class Plugin {
...
                             value: 'mm',
-                            helpText: 'You will need to disable notifications in Microsoft Teams to avoid duplicates. **[Learn more](http://google.com)**',
+                            helpText: 'You will get notifications in Mattermost for synced messages and channels. You will need to disable notifications in Microsoft Teams to avoid duplicates. **[Learn more](http://google.com)**',
                         },
                         {
-                            text: 'Microsoft Teams will be my primary platform',
+                            text: 'Microsoft Teams',
                             value: 'teams',
-                            helpText: 'Notifications in Mattermost will be muted for linked channels and DMs to prevent duplicates. Unread statuses in linked channels and DMs will also be disabled in Mattermost. **[Learn more](http://google.com)**',
+                            helpText: 'Notifications in Mattermost will be muted for linked channels and DMs to prevent duplicates. You can unmute any linked channel or DM/GM if you with to receive notifications. **[Learn more](http://google.com)**',
```
manojmalik20, line 43:
```suggestion
                            helpText: 'Notifications in Mattermost will be muted for linked channels and DMs to prevent duplicates. You can unmute any linked channel or DM/GM if you wish to receive notifications. **[Learn more](http://google.com)**',
```

manojmalik20 (approved):

lieut-data (approved):
Nice work, @larkox!

webapp/src/index.tsx
```diff
@@ -16,13 +16,40 @@ export default class Plugin {
...
+            icon: `${serverRoute}/plugins/${manifest.id}/public/msteams-sync-icon.svg`,
+            uiName: manifest.name,
+            sections: [{
+                settings: [{
+                    name: 'primary_platform',
+                    options: [
+                        {
+                            text: 'Mattermost',
+                            value: 'mm',
+                            helpText: 'You will get notifications in Mattermost for synced messages and channels. You will need to disable notifications in Microsoft Teams to avoid duplicates. **[Learn more](http://google.com)**',
```
marianunez, line 38: @abhijit-singh @johndavidlugtu what plans do we have for this link out to `Learn More`?

webapp/src/index.tsx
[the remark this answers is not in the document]
marianunez: Thanks for the explanation here @larkox! I see you went with your first option here which still leaves the flexibility for plugin authors to use anything else if they would like. 👍

marianunez (approved):
Thanks @larkox!

abhijit-singh (approved):
@larkox We don't have the `Learn more` links yet. I'd recommend merging this without them and we can add them in later if we have to.

larkox:
@yasserfaraazkhan You have probably tested this already, but just to be thorough I added you as a reviewer here. Let me know if you have any questions.

yasserfaraazkhan (approved):

lieut-data:
@larkox, I can't remember where this sits in terms of merging -- we're aiming to cut v1.6.0 next week which we'll ship to ServiceNow thereafter. Is this PR ok to include?

larkox:
@lieut-data The final server piece for this feature will arrive on 9.6 if I am not mistaken (@hmhealey can you confirm?). Because of that, I would say let's wait for the next version, so it arrives with the correct features. Sounds good?

lieut-data:
@larkox, understood: I'll remove the v1.6.0 milestone for now.

hmhealey:
No, the server changes are in 9.5. I forget the exact deadline we had for this (and the automuting stuff that I'm working on), but I was under the impression that the deadline for them is in the first week of February. I don't know if that corresponded to 1.6.0 or something different.

CC @johndavidlugtu here

johndavidlugtu:
@hmhealey just getting to this. It looks like you and the team has clarified it in the [channel](https://community.mattermost.com/core/pl/qjp8nqca7pgu9dc8ithtue9yrh) 👍

webapp/src/types/mattermost-webapp/index.d.ts
larkox: I feel this is more intention revealing, since you can name the key. But 0/5.

PR #422, opened 2023-11-28
Store the most recent changes time

jespino:
This fixes this issue: https://mattermost.atlassian.net/browse/MM-55561


server/handlers/handlers.go
```diff
@@ -101,6 +102,36 @@ func (ah *ActivityHandler) Start() {
...
+	// doStart is the meat of the activity handler worker
+	doStartLastActivityAt := func() {
+		for {
+			timer := time.NewTimer(5 * time.Minute)
+			select {
+			case <-timer.C:
+				ah.lastUpdateAtMap.Range(func(subscriptionID, lastUpdateAt any) bool {
+					if lastUpdateAt.(time.Time).Add(5 * time.Minute).After(time.Now()) {
+						if err := ah.plugin.GetStore().UpdateSubscriptionLastActivityAt(subscriptionID.(string), lastUpdateAt.(time.Time)); err != nil {
+							ah.plugin.GetAPI().LogWarn("Error storing the subscription last activity at", "error", err)
```
lieut-data, line 114: Can we include the `subscription_id` and `last_update_at` in the log?

server/handlers/handlers.go
```diff
@@ -101,6 +102,36 @@ func (ah *ActivityHandler) Start() {
...
 	}
 
+	// doStart is the meat of the activity handler worker
+	doStartLastActivityAt := func() {
+		for {
+			timer := time.NewTimer(5 * time.Minute)
+			select {
+			case <-timer.C:
+				ah.lastUpdateAtMap.Range(func(subscriptionID, lastUpdateAt any) bool {
+					if lastUpdateAt.(time.Time).Add(5 * time.Minute).After(time.Now()) {
```
lieut-data, line 112: nit: I had to sit down and think about what this meant. Maybe if we expressed it as the following instead?
```go
if time.Now().Sub(lastUpdateAt) <= 5 * time.Minute
```

server/handlers/handlers.go
```diff
@@ -101,6 +102,36 @@ func (ah *ActivityHandler) Start() {
...
+		}
+	}
+
+	if lastActivityAtMap, err := ah.plugin.GetStore().GetSubscriptionsLastActivityAt(); err != nil {
+		ah.plugin.GetAPI().LogError("Unable to load last activity subscription", "error", err)
+	} else {
+		for subscriptionID, lastActivityAt := range lastActivityAtMap {
+			ah.lastUpdateAtMap.Store(subscriptionID, lastActivityAt)
+		}
+	}
```
lieut-data, line 133: Can we get away with not populating the map in the first place? I'm thinking about the error case here where something goes wrong and the map starts out empty. The `Range` above will just no-op until the first event, after which the subscription will get updated the very next time the job runs.

server/handlers/handlers.go
```diff
@@ -101,6 +102,36 @@ func (ah *ActivityHandler) Start() {
...
+						if err := ah.plugin.GetStore().UpdateSubscriptionLastActivityAt(subscriptionID.(string), lastUpdateAt.(time.Time)); err != nil {
+							ah.plugin.GetAPI().LogWarn("Error storing the subscription last activity at", "error", err)
+						}
+					}
+					return true
+				})
+			case <-ah.quit:
+				timer.Stop()
+				// we have received a signal to stop
+				return
```
lieut-data, line 122: Should we flush our cache to the database when the worker exits?

server/store/sqlstore/store.go
```diff
@@ -730,6 +734,35 @@ func (s *SQLStore) UpdateSubscriptionExpiresOn(subscriptionID string, expiresOn
 	return nil
 }
 
+func (s *SQLStore) UpdateSubscriptionLastActivityAt(subscriptionID string, lastActivityAt time.Time) error {
+	query := s.getQueryBuilder().Update(subscriptionsTableName).Set("lastActivityAt", lastActivityAt.UnixMicro()).Where(sq.Eq{"subscriptionID": subscriptionID}, sq.Lt{"lastActivityAt": lastActivityAt.Add(-10 * time.Second).UnixMicro()})
+	_, err := query.Exec()
+	if err != nil {
+		return err
```
lieut-data, line 741: For the global subscription, I expect this `UPDATE` to fail periodically due to concurrent writes. Propose we wrap this update in a `READ COMMITTED` transaction that will have Postgres seamlessly retry (re-evaluating the `WHERE` conditions). (Also, I'm separately proposing that we make this plugin Postgres only.)

server/store/sqlstore/store.go
```diff
@@ -730,6 +734,35 @@ func (s *SQLStore) UpdateSubscriptionExpiresOn(subscriptionID string, expiresOn
...
+	}
+	return nil
+}
+
+func (s *SQLStore) GetSubscriptionsLastActivityAt() (map[string]time.Time, error) {
+	query := s.getQueryBuilder().Select("subscriptionID, lastActivityAt").From(subscriptionsTableName).Where(sq.NotEq{"lastActivityAt": nil}, sq.NotEq{"lastActivityAt": 0})
```
lieut-data, line 747: nit: I find it easier to parse and later extend these when they're written as:
```go
	query := s.getQueryBuilder().
		Select("subscriptionID, lastActivityAt").
		From(subscriptionsTableName).
		Where(sq.NotEq{
			"lastActivityAt": nil, 
			"lastActivityAt": 0,
		})
```

lieut-data (changes requested):
Nice! Laying the foundation for the next steps of using this data. A few comments below.

server/store/sqlstore/store.go
```diff
@@ -730,6 +734,35 @@ func (s *SQLStore) UpdateSubscriptionExpiresOn(subscriptionID string, expiresOn
 	return nil
 }
 
+func (s *SQLStore) UpdateSubscriptionLastActivityAt(subscriptionID string, lastActivityAt time.Time) error {
```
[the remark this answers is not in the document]
jespino: Extend is, in my opinion a bit more confusing. We are not "extending" it, maybe "AdvanceSubscriptionLastActivityAt"? But in that case I would pass a "duration", and we don't pass a duration. I think the `Update` one ist he one that defines better the behavior in terms of parameters and internal behavior. I could consider acceptable "AdvanceSubscriptionLastActivityAtTo", and pass the date that we want to put there.

server/handlers/handlers.go
jespino: sure

server/handlers/handlers.go
jespino: Something to think about, I think the problem of not having that initialized is about not syncing properly the messages later.

server/handlers/handlers.go
jespino: We could.

server/store/sqlstore/store.go
jespino: This shouldn't happen concurrently for the same subscription.

server/handlers/handlers.go
lieut-data: I'm assuming when we go to synchronize, we'll want to get the latest values from the database and not rely on the map, as we won't necessarily "see" changes that only get to other plugin instances.

server/store/sqlstore/store.go
```diff
@@ -730,6 +734,35 @@ func (s *SQLStore) UpdateSubscriptionExpiresOn(subscriptionID string, expiresOn
...
+	query := s.getQueryBuilder().Update(subscriptionsTableName).Set("lastActivityAt", lastActivityAt.UnixMicro()).Where(sq.Eq{"subscriptionID": subscriptionID}, sq.Lt{"lastActivityAt": lastActivityAt.Add(-10 * time.Second).UnixMicro()})
```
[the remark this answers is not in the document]
jespino: I think is not even needed, I'm removing it. That comes from an intermediate state of the code where I followed a different approach to store the the last activity at.

server/handlers/handlers.go
jespino: Yes, that's true, we can relay on database content, and that should be enough.

server/store/sqlstore/store.go
jespino: Still, they are idempotent and 1 transaction per query, so I don't see any problem here. We can talk it later today.

lieut-data: As discussed offline, the default transaction isolation level is already `READ COMMITTED`, so we're all good 🎉

server/handlers/handlers.go
lieut-data: Discussed offline, but agreed we can just remove this and keep it simpler 🎉

wiggin77 (approved):
LGTM :+1:

jespino:
@lindalumitchell There is nothing user facing to test here. The idea is that any activity coming from MSTeams side should be generating updates on the database (storing the last activity date). For testing this I would suggest some smoke testing around creation channel subscriptions and verifying that messages flows correctly in both directions. Feel free to defer the testing after the merge, to do it as part of the regular testing. I can give it a quick round myself after or before the merge if you want.

lindalumitchell (approved):
Sounds like a good plan to me, thanks @jespino!

PR #423, opened 2023-11-28
MM-55491: ensure handlers stop

lieut-data:
#### Summary
When the plugin stops normally – e.g. as part of rolling the k8s pods – it’s imperative that all outstanding requests be given a chance to terminate normally. Accordingly, let’s make `ActivityHandler.Stop` blocking.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-55491

### DIFF
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -9,7 +9,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/enescakir/emoji"
@@ -56,7 +55,7 @@ type ActivityHandler struct {
 	plugin               PluginIface
 	queue                chan msteams.Activity
 	quit                 chan bool
-	quitting             atomic.Bool
+	workersWaitGroup     sync.WaitGroup
 	IgnorePluginHooksMap sync.Map
 }
 
@@ -82,7 +81,7 @@ func New(plugin PluginIface) *ActivityHandler {
 }
 
 func (ah *ActivityHandler) Start() {
-	ah.quitting.Store(false)
+	ah.quit = make(chan bool)
 
 	// This is constant for now, but report it as a metric to future proof dashboards.
 	ah.plugin.GetMetrics().ObserveChangeEventQueueCapacity(activityQueueSize)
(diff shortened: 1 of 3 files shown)


server/handlers/handlers.go
```diff
@@ -82,7 +81,7 @@ func New(plugin PluginIface) *ActivityHandler {
 }
 
 func (ah *ActivityHandler) Start() {
-	ah.quitting.Store(false)
+	ah.quit = make(chan bool)
```
lieut-data, line 84: Previously discussed with @agnivade to move away from a queue that has to be sent a message per worker to just a `close(ah.quit)` and recreating on `Start`.

server/handlers/handlers.go
```diff
@@ -101,25 +100,32 @@ func (ah *ActivityHandler) Start() {
 		}
 	}
 
+	// doQuit is called when the worker quits intentionally
+	doQuit := func() {
+		ah.workersWaitGroup.Done()
+	}
```
lieut-data, line 106: This additional complexity helps me know when a particular worker is actually quitting. I contemplated putting this inside `isQuitting`, but the semantics felt weird since it's an API that is meant to be called repeatedly. Thoughts?

server/handlers/handlers.go
```diff
@@ -101,25 +100,32 @@ func (ah *ActivityHandler) Start() {
...
 
 func (ah *ActivityHandler) Stop() {
-	ah.quitting.Store(true)
-	go func() {
-		for i := 0; i < numberOfWorkers; i++ {
-			ah.quit <- true
-		}
-	}()
+	close(ah.quit)
+	ah.workersWaitGroup.Wait()
```
lieut-data, line 128: This is ultimately what makes `Stop` blocking.

server/handlers/worker.go
```diff
@@ -10,12 +10,13 @@ type workerMetrics interface {
 
 // startWorker wraps and invokes the given callback in a goroutine, automatically restarting in a
 // new goroutine on any unrecovered panic or unexpected termination.
-func startWorker(logError func(msg string, keyValuePairs ...any), metrics workerMetrics, isQuitting func() bool, callback func()) {
+func startWorker(logError func(msg string, keyValuePairs ...any), metrics workerMetrics, isQuitting func() bool, doStart, doQuit func()) {
```
lieut-data, line 13: Renamed to `doStart` to match with `doQuit`.

jespino (approved):
LGTM

mgdelacroix (approved):
LGTM! :+1:

PR #426, opened 2023-11-29
Increase subscription duration to 2 hours

jespino:
This fixes the ticket https://mattermost.atlassian.net/browse/MM-55564

### DIFF
--- a/server/msteams/client.go
+++ b/server/msteams/client.go
@@ -38,6 +38,10 @@ import (
 	"golang.org/x/oauth2"
 )
 
+const (
+	subscriptionExpirationTime = 2 * time.Hour
+)
+
 type ConcurrentGraphRequestAdapter struct {
 	msgraphsdk.GraphRequestAdapter
 	mutex sync.Mutex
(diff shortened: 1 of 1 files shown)


wiggin77 (approved):
LGTM :+1:

lieut-data (approved):
Nice! Quick question, in the ticket you mention:

> The only drawback that I could see here is the fact that we can end up keep receiving notifications even if we are stopped without clearing the subscription.

I couldn't quite understand the drawback here -- can you elaborate?

jespino:
@lieut-data if for whatever reason we don't properly remove the subscription, it can end up keep that subscription active for more time, so we are going to receive more messages without needing it. In any case, it is corner case scenario and is not specially bad.

jespino:
We are not changing much here, only the amount of time of the subscription, do you want to QA it, or do you prefer to just merge it?

jespino:
@lindalumitchell :point_up:

lindalumitchell (approved):
I think this looks fine to approve without a test build etc. 👍

PR #428, opened 2023-11-30
[MI-3789] Add plugin API to check if user is present in whitelist

ayusht2810:
#### Summary
Add plugin API to check if user is present in whitelist


ayusht2810:
closing in favour of https://github.com/Brightscout/mattermost-plugin-msteams-sync/pull/30

PR #429, opened 2023-11-30
Removing transactions outside the store

jespino:
This PR removes the exposed Begin, Commit and Rollaback for managing transactions from outside the store, that was used to generate some "transactions" grouping MSTeams client calls and database calls. This removes that pattern and replace, where it is needed, with a cluster lock (per Post). That way the only contention generated is over the same post operations.


server/store/sqlstore/store.go
```diff
@@ -710,15 +698,24 @@ func (s *SQLStore) SaveChatSubscription(subscription storemodels.ChatSubscriptio
...
 }
 
-func (s *SQLStore) SaveChannelSubscription(tx *sql.Tx, subscription storemodels.ChannelSubscription) error {
+func (s *SQLStore) SaveChannelSubscription(subscription storemodels.ChannelSubscription) error {
+	tx, err := s.db.Begin()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = tx.Rollback()
```
lieut-data, line 707: I /think/ this is fine, but wondering if we should be explicit and copy the pattern in the server to not rollback unless necessary?

```go
// finalizeTransactionX ensures a transaction is closed after use, rolling back if not already committed.
func finalizeTransactionX(transaction *sqlxTxWrapper, perr *error) {
	// Rollback returns sql.ErrTxDone if the transaction was already closed.
	if err := transaction.Rollback(); err != nil && err != sql.ErrTxDone {
		*perr = merror.Append(*perr, err)
	}
}
```

(Haven't looked at the `merror` stuff, 0/5 on that.)

server/store/sqlstore/store.go
```diff
@@ -1085,36 +1082,6 @@ func (s *SQLStore) IsUserPresentInWhitelist(userID string) (bool, error) {
...
 }
 
-func (s *SQLStore) LockPostByMSTeamsPostID(tx *sql.Tx, messageID string) error {
-	query := s.getQueryBuilder().Select("*").From(postsTableName).Where(sq.Eq{"msTeamsPostID": messageID}).Suffix("FOR UPDATE").RunWith(tx)
-	if _, err := query.Exec(); err != nil {
-		return err
-	}
-
-	return nil
-}
```
lieut-data, line 1095: I'm really glad this is gong away :)

server/message_hooks.go
```diff
@@ -238,47 +239,32 @@ func (p *Plugin) SetChatReaction(teamsMessageID, srcUser, channelID, emojiName s
 	}
 
 	var teamsMessage *clientmodels.Message
-	tx, err := p.store.BeginTx()
+
+	mutex, err := cluster.NewMutex(p.API, "post_mutex_"+chatID+teamsMessageID)
```
lieut-data, line 243: This seems slightly better than locking the Posts table, but slightly worse in that we're effectively writing out new rows to the KV store all the time.

When we pivot to driving this from shared channels, there will only be one instance processing these hooks, so the lock won't be necessary at all.

What are the risks if we simply don't lock at all here? I'm not certain of the "critical section" being guarded here.

lieut-data (changes requested):
Love the simplifications! Just one uncertainty about `cluster.NewMutex` -- inline below.

server/message_hooks.go
jespino: That was trying to fix a problem of reentrance of updates on reactions/post updates, that had a race condition and endup generating some inconsistencies. The point here is to avoid that.

I think you are right, the shared channels should solve this problem.

About the risk, if we don't lock at all there, probably it is going to bring back some problem with reactions, getting inconsistencies there, and even entering in some "looping" there updating a high number of times the reactions.

jespino: I would pass this through QA, and see if everything goes as expected, if that is the case, merge this and wait until Shared Channels machinery is used, and remove the mutexes from there.

server/store/sqlstore/store.go
jespino: That pattern is not "rollback unless is necessary", that patten ensure that if there is a problem with the rollback, it is added to a list of errors, to be handled outside of the function. In this case, I think we can't do much with the error there, we can maybe log it, but I don't think we have much to do. If commit went wrong, and rollback went wrong... what it means? If commit went ok, and rollback went wrong, what it means? I think just swallowing that error, or logging it is enough in this case.

lieut-data: Ah, you're right: I forgot that the `transaction.Rollback()` is called unconditionally in both cases. Never mind :D

server/message_hooks.go
lieut-data: > About the risk, if we don't lock at all there, probably it is going to bring back some problem with reactions, getting inconsistencies there, and even entering in some "looping" there updating a high number of times the reactions.

Can you elaborate a bit on this? I'd like to understand what went wrong before we had the transactional locks. It would be great to test this somehow too -- I'm guessing the point of contention was to have a whole bunch of emojis added to a post... but then what was it that "went wrong"? The loop is especially concerning, but I'm wondering how that would happen.

jespino: that reintroduces the the bugs #327 and #330

jespino: I mean, it reintroduces the bug when we create the transactions, and the mutex in theory should fix that

jespino: The reality, I'm not 100% sure if we also need to apply the mutex at the MSTeams activity handler to avoid properly the race conditions, because "locking the whole table" was a very intrusive approach.

lieut-data: Discussed offline: we agreed to preserve the semantics here for now and revisit more holistically once we're leveraging shared channels as the substrate.

lieut-data (approved):

wiggin77 (approved):
This looks great!  :+1:

jespino:
@lindalumitchell for testing this it should be some "smoke testing" around the regular behavior of the app, and also verify that this two bugs (#327 and #330) aren't introduced again.

lindalumitchell (approved):
Given the trickiness of only being able to truly test the plugin functionality on the one test server that is connected to the MS Teams instance, and given that that server currently has the version we want to be testing, I'm going to defer QA on this one as well (rather than upload it to our test server).

PR #436, opened 2023-12-04
MM-55507: adopt ServeMetrics hook

lieut-data:
#### Summary
Adopt the new [`ServeMetrics`](https://github.com/mattermost/mattermost/pull/24249) hook to expose Prometheus metrics on `:8067/plugins/com.mattermost.msteams-sync/metrics`. This /duplicates/ the existing `:9094/metrics` for now, with the latter to be removed once we finish migrating to the new endpoint in Mattermost Cloud. This is not a breaking change.

As part of this change, we unconditionally initialize the `metricsServer` so it's never `nil`. We also now only stop the metrics server `OnDeactivate`.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-55507

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/gosimple/slug v1.13.1
 	github.com/jmoiron/sqlx v1.3.5
 	github.com/lib/pq v1.10.9
-	github.com/mattermost/mattermost/server/public v0.0.10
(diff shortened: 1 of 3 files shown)


streamer45 (approved):
Nice :rocket:

cpoile (approved):
Simple enough, nice. :)

PR #438, opened 2023-12-08
Added functionality for user to connect and disconnect their account through RHS UI

ayusht2810:
#### Summary

- Added functionality for users to connect and disconnect their accounts through the RHS UI.
- Added functionality to hide/show RHS on the basis of config.

#### Screenshots
image
image
image
image


larkox:
@ayusht2810 Just to verify before getting into the review. I see changes that at a first glance do not seem related, like the APIs for getting the channels or teams from MSTeams. Are these changes related to this PR and should be reviewed? Or there has been some issue with the PR?

ayusht2810:
@larkox The server changes are due to the addition of new plugin APIs and their related changes required for the UI integration.
Also, the base branch of this PR is `epic-ui`, which is synced with the `main` branch. We can merge all the phases of UI into the `epic-ui` branch, then merge the `epic-ui` branch into the main.
Let me know your thoughts on this.

server/get_teams_data.go
```diff
@@ -0,0 +1,281 @@
...
+			p.API.LogInfo("User does not have the permissions for the requested channel", "UserID", userID, "ChannelID", link.MattermostChannelID)
+			continue
+		}
+
+		msTeamsTeamIDsVsNames[link.MSTeamsTeamID] = ""
+		msTeamsChannelIDsVsNames[link.MSTeamsChannelID] = ""
+
+		// Build the channels query for each team
+		if msTeamsTeamIDsVsChannelsQuery[link.MSTeamsTeamID] == "" {
+			msTeamsTeamIDsVsChannelsQuery[link.MSTeamsTeamID] = "id in ("
```
larkox, line 37: I see this parenthesis is closed in line 78, but doesn't it make more sense to close it in this function (around line 44 I guess)? Why is it that we close it so far away?

larkox (changes requested):
Great work! I don't remember anything that needs to be fixed, but since I left many comments, I still marked this as `request changes`. Feel free to dismiss any comment you don't think is valid, and re-request my review.

server/get_teams_data.go
ayusht2810: We need to run an extra loop on the map to add the closing bracket, which can be avoided by closing it in the [line](https://github.com/mattermost/mattermost-plugin-msteams-sync/blob/620e6723c8215ccb864580e26cb47a1e95ddb597/server/get_teams_data.go#L77).

webapp/src/containers/Rhs/views/LinkedChannels/LinkedChannels.container.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    // Increase the page number by 1
+    const handlePagination = () => {
+        setPaginationQueryParams({...paginationQueryParams, page: paginationQueryParams.page + 1,
+        });
+    };
+
+    const hasMoreLinkedChannels = useMemo<boolean>(() => (
+        (totalLinkedChannels.length - (paginationQueryParams.page * defaultPerPage) === defaultPerPage)
+    ), [totalLinkedChannels]);
```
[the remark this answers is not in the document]
ayusht2810: Yes, you are right we have added this to update it when we receive more linked channels

webapp/src/containers/Rhs/views/LinkedChannels/LinkedChannels.container.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const getChannels = () => new Promise<void>((res) => {
+        setTimeout(() => res(), 2000);
+    });
+
+    useEffect(() => {
+        getChannels().then(() => {
+            const linkedChannels = mockLinkedChannels.slice((paginationQueryParams.page * paginationQueryParams.per_page), (paginationQueryParams.page + 1) * paginationQueryParams.per_page);
+            setTotalLinkedChannels([...totalLinkedChannels, ...(linkedChannels as ChannelLinkData[])]);
+        });
+    }, [paginationQueryParams]);
```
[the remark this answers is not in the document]
ayusht2810: This is used for testing. We have added a TODO comment for the same

webapp/src/components/LinkedChannelCard/LinkedChannelCard.component.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import {Icon} from 'components/Icon';
+
+import {LinkedChannelCardProps} from './LinkedChannelCard.types';
+
+import './LinkedChannelCard.styles.scss';
+
+export const LinkedChannelCard = ({msTeamsChannelName, msTeamsTeamName, mattermostChannelName, mattermostTeamName}: LinkedChannelCardProps) => (
```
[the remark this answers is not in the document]
ayusht2810: This can be fixed after we update the lint rules present in the webapp setup automatically. This was also help us in fixing the import order to avoid fixing it manually)

webapp/src/components/Icon/Icon.map.tsx
```diff
@@ -0,0 +1,450 @@
+import React from 'react';
+
+import {IconName} from './Icon.types';
+
+export const IconMap : Record<IconName, JSX.Element> = {
```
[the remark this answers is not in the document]
ayusht2810: We have used various [MM UI library](https://github.com/Brightscout/mm-ui-library/tree/mm_v6) components that have support for icons to handle different themes. After moving the icons to the above library, we can remove this component and use them from there itself.

ayusht2810:
@larkox fixed and resolved the review comments you gave and replied to some of the comments. Please re-review the PR.

webapp/src/containers/Rhs/views/ConnectedAccount/ConnectedAccount.container.tsx
```diff
@@ -1,41 +1,42 @@
...
+import {ReduxState} from 'types/common/store.d';
+
 import utils from 'utils';
 
 export const ConnectedAccount = () => {
     const dispatch = useDispatch();
+    const [showDialog, setShowDialog] = useState(false);
     const {makeApiRequestWithCompletionStatus, state} = usePluginApi();
     const {username, isAlreadyConnected, msteamsUserId, connected} = getConnectedState(state);
+    const globalState = useSelector((reduxState: ReduxState) => reduxState);
```
larkox, line 27: This makes so any change in the state of the app causes a re-render for this component. Is it intended? Shouldn't we get just the part of the state we need?

webapp/src/containers/Rhs/views/LinkedChannels/LinkedChannels.container.tsx
```diff
@@ -43,48 +44,50 @@ export const LinkedChannels = () => {
     }, []);
 
     // Increase the page number by 1
-    const handlePagination = () => {
-        setPaginationQueryParams({...paginationQueryParams, page: paginationQueryParams.page + 1,
-        });
-    };
+    const handlePagination = useCallback(() => {
+        setPaginationQueryParams((prev) => ({...prev, page: paginationQueryParams.page + 1,
```
larkox, line 48: any strong reason not to use `prev.page + 1` instead of `paginationQueryParams.page + 1`?

larkox (approved):
LGTM, just a couple comments.

lieut-data:
@larkox & @hmhealey, please ping me once reviews are "done" and if there's anything I can do to help complete this PR.

webapp/src/components/WarningCard/WarningCard.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+                border: '1px solid #F7434329',
+            }}
+        >
+            <div className='d-flex gap-12'>
+                <Icon iconName='warning'/>
+                <div>
+                    <div className='d-flex align-items-start justify-between'>
+                        <h5 className='wt-600 mt-0'>{'Please Connect your MS Teams account.'}</h5>
+                    </div>
+                    <p>{'You are not connected to your MS Teams account yet, please connect to your account to continue using MS Teams sync.'}
```
hmhealey, line 28: This isn't something we're doing in this PR, but all this UI text should be translatable eventually

webapp/src/components/enforceConnectedAccountModal.tsx
```diff
@@ -1,49 +1,73 @@
...
+        makeApiRequestWithCompletionStatus(Constants.pluginApiServiceConfigs.needsConnect.apiServiceName);
     }, []);
 
-    const checkConnected = useCallback(async () => {
-        const result = await Client.needsConnect();
-        if (!result.needsConnect) {
-            setOpen(false);
-            setConnecting(false);
-        }
+    const {data: needsConnectData} = useMemo(() => getApiState(Constants.pluginApiServiceConfigs.needsConnect.apiServiceName), [getApiState]);
```
hmhealey, line 29: I have similar comments here about potentially not needing a `useMemo` around `getApiState` and being able to use `useEffect` instead of `useApiRequestCompletionState` for these cases.

The `useApiRequestCompletionState` one might be different in this case though since it's making repeated requests for that

webapp/src/containers/Rhs/Rhs.container.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    const {isOpen} = getSnackbarState(state);
+
+    const {data: linkedChannels} = getApiState(pluginApiServiceConfigs.getLinkedChannels.apiServiceName);
+
+    // NOTE: Commented out on purpose.This is part of Phase-II
+    // const isAnyChannelLinked = useMemo(() => Boolean((linkedChannels as ChannelLinkData[])?.length), [linkedChannels]);
+    const isAnyChannelLinked = false;
+
+    const getRhsView = useCallback(() => {
```
hmhealey, line 30: I don't think the `useCallback` is actually needed here since we're calling the function immediately, and the work done by the function is simple. Tracking if the dependencies of this have changed is likely just as much work for the browser as it would to write this without the hook

webapp/src/containers/Rhs/views/LinkedChannels/LinkedChannels.mock.ts
```diff
@@ -0,0 +1,219 @@
+export const mockLinkedChannels = [
+
+    {
+        msTeamsTeamID: 'string',
```
hmhealey, line 4: Nitpicking a bit, but I feel like repeating the same string for all these risks missing bugs because arguments are passed in the wrong order since they all have the same value. For example, if we swap the team and channel name by mistake, the snapshot won't show any differences

webapp/src/hooks/useApiRequestCompletionState.ts
```diff
@@ -0,0 +1,49 @@
...
+            }
+
+            if (!isUninitialized) {
+                dispatch(resetApiRequestCompletionState(serviceName));
+            }
+        }
+    }, [
+        requests.includes(serviceName),
+        apiState,
+        resetApiRequestCompletionState(serviceName),
```
hmhealey, line 45: I'm not super familiar with Redux Toolkit so I might be incorrect here, but this seems like `resetApiRequestCompletionState(serviceName)` is going to return a new object every time it's called which would cause the `useEffect` to re-run on every render cycle. While the `exhaustive-deps` ESLint rule would probably complain, I feel like we shouldn't include this as a dependency here

webapp/src/hooks/usePluginApi.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    const makeApiRequestWithCompletionStatus = useCallback(async (serviceName: PluginApiServiceName, payload: APIRequestPayload) => {
+        const apiRequest = await makeApiRequest(serviceName, payload);
+
+        if (apiRequest) {
+            dispatch(setApiRequestCompletionState(serviceName));
+        }
+    }, [dispatch, makeApiRequest, setApiRequestCompletionState]);
+
+    // Pass payload in POST requests only. For GET requests, there is no need to pass a payload argument
```
hmhealey, line 27: This comment seems like it was unintentionally duplicated from above
```suggestion
```

webapp/src/hooks/usePluginApi.ts
```diff
@@ -0,0 +1,36 @@
...
+    const getApiState = useCallback((serviceName: PluginApiServiceName, payload: APIRequestPayload) => {
```
hmhealey, line 28: I'm not certain, but this feels like this `useCallback` may not be actually helping here much since it returns a new callback whenever anything in the plugin state changes, even if it's unrelated to the specific API that we want. I''m not sure if a good way to do this if we want `getApiState` to still be a callback, but if we just return `apiState` from `usePluginApi`, I think we should just be able to return `msTeamsPluginApi.endpoints[serviceName].select(payload)(state)` without wrapping it in any extra hooks

hmhealey (changes requested):
I'm dumping a lot here, so let me know if there's anything I can help with

PR #439, opened 2023-12-11
restore metrics server stop

lieut-data:
#### Summary
In https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/436, I incorrectly removed the `p.metricsServer.Shutdown()` call during `plugin.stop()`, believing it would incorrectly impact the new server managed `ServeMetrics` hook, but this wasn't true: we should indeed stop listening on the requested port if the configuration so changes, and we can continue handling `ServeMetrics` without issue.

#### Ticket Link
Relates-to: https://mattermost.atlassian.net/browse/MM-55507

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -280,6 +280,11 @@ func (p *Plugin) stop() {
 	if p.monitor != nil {
 		p.monitor.Stop()
 	}
+	if p.metricsServer != nil {
+		if err := p.metricsServer.Shutdown(); err != nil {
+			p.API.LogWarn("Error shutting down metrics server", "error", err)
+		}
+	}
 	if p.stopSubscriptions != nil {
 		p.stopSubscriptions()
 		time.Sleep(1 * time.Second)


wiggin77 (approved):

server/plugin.go
```diff
@@ -280,6 +280,11 @@ func (p *Plugin) stop() {
 	if p.monitor != nil {
 		p.monitor.Stop()
 	}
+	if p.metricsServer != nil {
+		if err := p.metricsServer.Shutdown(); err != nil {
+			p.API.LogWarn("Error shutting down metrics server", "error", err)
```
streamer45, line 285: Funny how we log an error as a warning. I sort of understand the reasoning, we are shutting down after all, but find it interesting nonetheless.

streamer45 (approved):
Makes sense

server/plugin.go
lieut-data: My own rationale is ERROR = "a human must take some action". Despite an error having occurred in this case, given we're shutting down anyway, no human need get involved. BUT where it gets interesting is also that a "spike" in WARN logs is also something a human should investigate. Next up for MS Teams is a holistic review of all logging to align to this philosophy, and then enable Grafana alerts for all ERROR logs and spikes in WARN logs. (Ambitious, I know! We'll see where we land :p)

PR #440, opened 2023-12-11
guard msteamsAppClient

lieut-data:
#### Summary
We had a mutex guarding msteamsAppClient, but it wasn't being used consistently. Fix that, and change it to a `RWMutex` to minimize the critical sections. Note that we never hold the mutex while using client: we're effectively just protecting the pointer.

Finally, I note we currently don't ever change the `msteamsAppClient`, even if the configuration changes. So this mutex is also technically unneeded. But it probably makes sense to support reconnecting the client on a real configuration change in the future without having to restart the server.

#### Ticket Link
Relates-to: https://mattermost.atlassian.net/browse/MM-55490


wiggin77 (approved):
LGTM :+1:

crspeller (approved):

PR #442, opened 2023-12-11
Refresh oauth tokens if needed

jespino:
### Summary
We rely on https://pkg.go.dev/golang.org/x/oauth2  to oversee handling of the oauth2 tokens used to authenticate as individual users. The library automatically refreshes the tokens, but provides no built-in way to notify the caller when this refresh occurs. For refresh tokens that never expire, this is “fine”, but in the case of MS Teams, the refresh tokens expires after 90 days and we lose the ability to refresh the access token required to make the request.

This pull request adds support for detecting changed refresh or access tokens and persisting them back to the database as this occurs. 

### Ticket
Fixes: https://mattermost.atlassian.net/browse/MM-56063


jespino:
I think this is going to be really challenging to test properly.

server/plugin.go
```diff
@@ -142,6 +142,22 @@ func (p *Plugin) GetClientForUser(userID string) (msteams.Client, error) {
 	if token == nil {
 		return nil, errors.New("not connected user")
 	}
+
+	if token.Expiry.Before(time.Now()) {
```
lieut-data, line 146: Should we more aggressive with our window handling? For example, if the token hasn't expired yet, but does so in 1 second, by the time I use this client it might have expired, the client will silently renew it, and we won't ever save that back to the database.

server/plugin.go
```diff
@@ -142,6 +142,22 @@ func (p *Plugin) GetClientForUser(userID string) (msteams.Client, error) {
...
+		var err error
+		token, err = msteams.RefreshToken(p.GetURL()+"/oauth-redirect", p.getConfiguration().TenantID, p.getConfiguration().ClientID, p.getConfiguration().ClientSecret, token)
+		if err != nil {
+			return nil, err
+		}
```
lieut-data, line 151: I wonder if we can move this to a `client.Refresh`? Or does the client get "used" by the builder before we can refresh it ourselves?

server/plugin.go
```diff
@@ -152,6 +168,21 @@ func (p *Plugin) GetClientForTeamsUser(teamsUserID string) (msteams.Client, erro
 		return nil, errors.New("not connected user")
 	}
 
+	if token.Expiry.Before(time.Now()) {
+		var err error
+		token, err = msteams.RefreshToken(p.GetURL()+"/oauth-redirect", p.getConfiguration().TenantID, p.getConfiguration().ClientID, p.getConfiguration().ClientSecret, token)
+		if err != nil {
+			return nil, err
+		}
+		userID, err := p.store.TeamsToMattermostUserID(teamsUserID)
```
lieut-data, line 177: I understand that this is the "inverse" of what we do in `GetClientForUser`, but I wonder if we can simplify and reduce code duplication here if we just reimplement this method as something like the following?
```go
func (p *Plugin) GetClientForTeamsUser(teamsUserID string) (msteams.Client, error) {
        userID, err := p.store.TeamsToMattermostUserID(teamsUserID)
        if err != nil {
                return nil, err
        }
        
        return p.GetClientForUser(userID)
}
```

This does add an extra database lookup for this code path (unless we also happen to refresh), but feels more resilient. And maybe we can optimize that later by just returning more from the store in the first place allowing us to do the conversion instantly?

lieut-data (changes requested):
Thanks for moving fast on this one, @jespino! A few comments below, but also one higher level question.

At the core, we're relying on the fact that by refreshing preemptively, we avoid the client doing it "transparently" and without saving back to the database. I guess, though, in principle it could still happen -- one case below, where our expiry window is near to the current time, but maybe for other reasons I can't imagine.

One potentially bullet proof approach would be to persist the changes /after/ we're done using the client, checking if the refresh or access token changed from initialization and using that as the trigger to write back to the database. There's always the chance that this fails too, e.g. we crash earlier, but I wonder if it's conceptually simpler and more bullet proof to unexpected events?

server/plugin.go
jespino: This can happen, but whenever we get the token again from the database is going to be obsolete enough, and eventually we are going to reach the point where the background job that keeps them alive is going to detect that the session is not properly updated and refresh it anyway. We need to have very bad luck to reach that point, but even with bad luck, that point should take care of it.

server/plugin.go
jespino: I can check it, but my concern is that the whole process of creating the client can lead to refreshing the token without noticing, but let me check it, and I'll see if we can move that as a method.

server/plugin.go
jespino: Yes, that makes sense, let me change that.

server/plugin.go
jespino: Done

server/plugin.go

wiggin77 (approved):
LGTM :+1:

lieut-data (approved):

server/msteams/client.go
```diff
@@ -252,6 +253,20 @@ func NewTokenClient(redirectURL, tenantID, clientID, clientSecret string, token
 	return client
 }
 
+func (tc *ClientImpl) RefreshToken(redirectURL string, token *oauth2.Token) (*oauth2.Token, error) {
```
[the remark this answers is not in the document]
jespino: Done

server/plugin.go
```diff
@@ -142,18 +142,33 @@ func (p *Plugin) GetClientForUser(userID string) (msteams.Client, error) {
...
+	if token.Expiry.Before(time.Now()) {
+		newToken, err := timerClient.RefreshToken(p.GetURL()+"/oauth-redirect", token)
+		if err != nil {
+			return nil, err
+		}
+		teamsUserID, err := p.store.MattermostToTeamsUserID(userID)
+		if err != nil {
+			return nil, err
+		}
+		if err := p.store.SetUserInfo(userID, teamsUserID, newToken); err != nil {
```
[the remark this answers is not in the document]
jespino: I going to defer this change until later

jespino:
@lindalumitchell I going to merge this without QA pass because is really hard to test it, so I going to build this branch and upload it to the test server that we have, and smoke test it, after that, I going to merge the PR, and we keep testing it later. Does it sounds ok?

lindalumitchell:
Thanks @jespino, yes, that sounds fine to me.

PR #444, opened 2023-12-11
Store initial preferences

larkox:
#### Summary
Store initial preference in the preferences.

Any side effect will be handled by the hook.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55638

### DIFF
--- a/server/constants/preferences.go
+++ b/server/constants/preferences.go
@@ -1,0 +1,5 @@
+package constants
+
+const PREFERENCE_NAME_PLATFORM = "platform"
+const PREFERENCE_VALUE_PLATFORM_MM = "mattermost"
+const PREFERENCE_VALUE_PLATFORM_MSTEAMS = "msteams"
--- a/server/api.go
+++ b/server/api.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-plugin-msteams-sync/server/constants"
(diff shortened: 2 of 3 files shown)


marianunez (commented):
Not sure if you want to include in this PR or separate but part of the work that was needed to finish the initial screen was also to get the UX to be up to date to the latest designs: https://www.figma.com/file/BDolYCzc6CMCgVMa3U1p9B/MS-Teams-Sync---Choose-Primary-Platform?type=design&node-id=253-25533&mode=design&t=Lvisli7vIP3Vc6tB-4

larkox:
@marianunez I feel both tasks are completely separated, so we can handle them in different PRs.

marianunez:
> @marianunez I feel both tasks are completely separated, so we can handle them in different PRs.

Sounds good. Feel free to create a ticket to cover the UX changes

server/constants/preferences.go
```diff
@@ -0,0 +1,7 @@
+package constants
```
hmhealey, line 1: 0/5 since I haven't looked around this code enough to know where we'll need these, but I don't think we want a separate package just for constants. We can always move it when we find a home for it though

hmhealey (approved):

server/constants/preferences.go
larkox: I agree we can move it to the right place when we find it. For now I prefer to have it here, just as a way to keep things organized. Moving this around won't be a breaking change :stuck_out_tongue:

marianunez (approved):
Thanks @larkox!

PR #445, opened 2023-12-11
MM-55490: dont reset on configuration changes

lieut-data:
#### Summary
When I originally drafted the linked ticket, I imagined we would completely eliminate the `p.restart()` method and just keep all the jobs running all the time, with any configuration changes being resolved at run time. This turned out to be trickier than I expected, so I constrained to the primary focus of the overarching epic: avoid anything that disrupts the correct functioning of the application.

With that constrained focus, I'm only preventing the reset of two things: the main app client, and the activity handlers. The jobs being reset are "fine" because they aren't directly user facing. But imagine for a moment that we stop the activity handlers and then go to stop the users sync job, but it's running, so it takes a while: all the time the activity handlers are shut down and the change event queue is theoretically backing up. So we avoid this altogether.

Not reconnecting the app client is just a recognition of the fact that we never supported a "new configuration" in this method in the first place. In principle, we could extend this to only reconnect if we detect a change in tenant id, etc., but it's out of scope for this epic.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-55490

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -176,7 +176,7 @@ func (p *Plugin) connectTeamsAppClient() error {
 	return nil
 }
 
-func (p *Plugin) start() {
+func (p *Plugin) start(isRestart bool) {
 	enableMetrics := p.API.GetConfig().MetricsSettings.Enable
 
 	if enableMetrics != nil && *enableMetrics {
@@ -186,7 +186,10 @@ func (p *Plugin) start() {
 		p.runMetricsUpdaterTask(p.store, updateMetricsTaskFrequency)
 	}
 
-	p.activityHandler.Start()
+	// We don't restart the activity handler since it's stateless.
+	if !isRestart {
+		p.activityHandler.Start()
+	}
 
 	err := p.connectTeamsAppClient()
 	if err != nil {
@@ -266,7 +269,7 @@ func (p *Plugin) getPrivateKey() (*rsa.PrivateKey, error) {
 	return privateKey, nil
 }
 
-func (p *Plugin) stop() {
+func (p *Plugin) stop(isRestart bool) {
 	if p.monitor != nil {
 		p.monitor.Stop()
 	}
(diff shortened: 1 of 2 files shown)


server/plugin.go
```diff
@@ -158,16 +158,20 @@ func (p *Plugin) connectTeamsAppClient() error {
...
-			p.getConfiguration().ClientSecret,
-			&p.apiClient.Log,
-		)
-
-		p.msteamsAppClient = client_timerlayer.New(msteamsAppClient, p.GetMetrics())
+	// We don't currently support reconnecting with a new configuration: a plugin restart is
+	// required.
+	if p.msteamsAppClient != nil {
+		return nil
 	}
```
lieut-data, line 165: Note that, before, we only ever created `p.msteamsAppClient` once: so calling `connectTeamsAppClient` multiple times just amounted to calling `p.msteamsAppClient.Connect()` multiple times. There's no reason to do that on a configuration change if we're not going to reflect an updated configuration.

server/plugin.go
```diff
@@ -158,16 +158,20 @@ func (p *Plugin) connectTeamsAppClient() error {
...
-	if p.msteamsAppClient == nil {
-		msteamsAppClient := msteams.NewApp(
-			p.getConfiguration().TenantID,
-			p.getConfiguration().ClientID,
-			p.getConfiguration().ClientSecret,
-			&p.apiClient.Log,
-		)
-
-		p.msteamsAppClient = client_timerlayer.New(msteamsAppClient, p.GetMetrics())
+	// We don't currently support reconnecting with a new configuration: a plugin restart is
```
lieut-data, line 161: This just makes explicit what is already true.

server/plugin_test.go
```diff
@@ -423,67 +423,38 @@ func TestSyncUsers(t *testing.T) {
 	}
 }
 
-func TestConnectTeamsAppClient(t *testing.T) {
```
lieut-data, line 426: This wasn't really testing anything other than calling order, and there's no way to mock it now that the implementation is simpler. I'll revisit this when we get to the integration testing epic.

jespino (approved):
LGTM

wiggin77 (approved):
LGTM :+1:

PR #446, opened 2023-12-12
only target linux-* architectures

lieut-data:
#### Summary
Change our plugin build strategy to target only `linux-amd64` and `linux-arm64`. Previously, we listed all of `linux-*`, `darwin-*` and `windows-*` in our manifest, but then only built `linux-amd64`.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-56175


wiggin77 (approved):
LGTM :+1:     Not blocking, but I don't quite see what is driving the decision regarding what builds we "support".  Why is this Linux only?

lieut-data:
@wiggin77, good question: I don't know the backstory there. Kept the restriction for now, but let's definitely revisit.

jespino:
@lieut-data there is not much back story there, was faster to just compile for 1 architecture and also was making the binary smaller and easier to deploy in environments where the upload size was limited. If that restrictions no longer apply, we can re-enable everything.

jespino (approved):

PR #450, opened 2023-12-18
Restoring the generate make command

jespino:
After the update of the starter template, the `make generate` command disappeared, and it was convenient to generate layers and mocks. This PR restores it.


wiggin77 (approved):

lieut-data (approved):
Oops, my bad! Thank you :)

PR #453, opened 2023-12-20
Notify user when reconnection is needed

jespino:
Whenever we detect a user connection is not working (because we receive a 401 error from the API or we receive a "oauth2" token error), we remove the token for the user and notify him to connect the account again. This fixes: [MM-56152](https://mattermost.atlassian.net/browse/MM-56152)


jespino:
@lieut-data this is the approach that I followed finally for the invalid token invalidation and notification.

server/msteams/client.go
```diff
@@ -140,14 +141,20 @@ func NormalizeGraphAPIError(err error) error {
 				message = *terr.GetMessage()
 			}
 			return &GraphAPIError{
-				Code:    code,
-				Message: message,
+				Code:       code,
+				Message:    message,
+				StatusCode: e.ResponseStatusCode,
```
jespino, line 146: I'm getting the status code to detect the 401 cases that means the token is not valid.

server/msteams/client.go
```diff
@@ -140,14 +141,20 @@ func NormalizeGraphAPIError(err error) error {
...
 			}
 		}
 	default:
+		statusCode := 0
+		if strings.HasPrefix(err.Error(), "oauth2: ") {
```
jespino, line 151: Also, in some cases we are returining directly that the token was unable to be properly processed, I have to find a better way of doing this, but the idea is to capture oauth2 related problems to consider them a 401 error too.

server/msteams/client_disconnectionlayer/disconnectionlayer.go
```diff
@@ -0,0 +1,622 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
jespino, line 1: This is the generated layer for the disconnection detection. The main point is that you can create the layer on top of the client and if the returned error is a 401 error we call the onDisconnect callback with the user id provided in the constructor.

server/plugin.go
```diff
@@ -138,12 +139,40 @@ func (p *Plugin) GetURL() string {
...
+}
+
 func (p *Plugin) GetClientForUser(userID string) (msteams.Client, error) {
 	token, _ := p.store.GetTokenForMattermostUser(userID)
 	if token == nil {
 		return nil, errors.New("not connected user")
 	}
+
 	client := p.clientBuilderWithToken(p.GetURL()+"/oauth-redirect", p.getConfiguration().TenantID, p.getConfiguration().ClientID, p.getConfiguration().ClientSecret, token, &p.apiClient.Log)
+	client = client_disconnectionlayer.New(client, userID, p.OnDisconnectedTokenHandler)
```
jespino, line 175: Then, every time I create user client I add the `client_disconnectionlayer` there with the userID of the client and the disconnection handler.

server/plugin.go
```diff
@@ -138,12 +139,40 @@ func (p *Plugin) GetURL() string {
 	return *config.ServiceSettings.SiteURL + "/plugins/" + pluginID
 }
 
+func (p *Plugin) OnDisconnectedTokenHandler(userID string) {
```
jespino, line 142: Whenever a 401 is detected in the client, we get rid of the user token from the database and send him a message telling that he is disconnected and he should reconnect using the slash command.

server/msteams/client.go
jespino: After investigating the source code, that there are some errors that are returned as `errors.New("oauth2: ...")`, so not sure if there is any better way to not use the string based approach. What I saw is that all the errors that I've seen use the `oauth2:` prefix in the library, so I think is pretty safe to use it.

wiggin77: I would recommend an IsOAuthError(err) method that can handle string prefixes and error instances, and perhaps be less brittle.

server/plugin.go
```diff
@@ -138,17 +139,44 @@ func (p *Plugin) GetURL() string {
...
+	}
+	channel, appErr := p.API.GetDirectChannel(userID, p.GetBotUserID())
+	if appErr != nil {
+		p.API.LogWarn("Unable to get direct channel for send message to user", "userID", userID, "error", appErr.Error())
+		return
+	}
+	_, appErr = p.API.CreatePost(&model.Post{
+		UserId:    p.GetBotUserID(),
+		ChannelId: channel.Id,
+		Message:   "Your connection to MS Teams has been lost. Please reconnect to using /msteams-sync connect slash command.",
```
wiggin77, line 161:
```suggestion
		Message:   "Your connection to MS Teams has been lost. Please reconnect to using '/msteams-sync connect' slash command.",
```

wiggin77 (approved):
LGTM :+1:

server/msteams/client.go
jespino: Good point, at least we encapsulate the approach in an specific point to avoid propagation of this bad pattern.

server/msteams/client.go
lieut-data: If this is true:

> The 401 (Unauthorized) status code indicates that the request has not been applied because it lacks valid authentication credentials for the target resource.

Then we can't just trigger `OnDisconnect` simply because a particular resource is unavailable, as a reconnection wouldn't help anyway, and this might lead to a stream of "random" disconnects.

Can we constrain our disconnection logic more narrowly?

server/plugin.go
```diff
@@ -138,17 +139,44 @@ func (p *Plugin) GetURL() string {
...
+	p.API.LogDebug("OnDisconnectedTokenHandler", "userID", userID)
+	teamsUserID, err := p.store.MattermostToTeamsUserID(userID)
+	if err != nil {
+		p.API.LogWarn("Unable to get teams user id from mattermost to user", "userID", userID, "error", err.Error())
+		return
+	}
+	if err2 := p.store.SetUserInfo(userID, teamsUserID, nil); err2 != nil {
+		p.API.LogWarn("Unable clean invalid token for the user", "userID", userID, "error", err2.Error())
+		return
+	}
```
lieut-data, line 152: Can we explicitly track this disconnection in a new column `DisconnectedAt`? This will allow us to enumerate the users that were disconnected and, for example, produce a report.

server/plugin.go
```diff
@@ -138,17 +139,44 @@ func (p *Plugin) GetURL() string {
...
+	_, appErr = p.API.CreatePost(&model.Post{
+		UserId:    p.GetBotUserID(),
+		ChannelId: channel.Id,
+		Message:   "Your connection to MS Teams has been lost. Please reconnect to using '/msteams-sync connect' slash command.",
+	})
+	if appErr != nil {
+		p.API.LogWarn("Unable to send direct message to user", "userID", userID, "error", appErr.Error())
+	}
+}
```
lieut-data, line 166: I'm hesitant to /immediately/ send a message to the enduser. There's a chance that events come into multiple plugins (or even activity handlers) and we fire off a rash of messages to the same user.

One approach might be to introduce a new job whose responsibility it is to notify these users, and add a column `DisconnectNotifiedAt` to track when last done. I'm leaning towards being super robust here, especially if there's a case where we "miss" sending a message -- the job would allow us to seamlessly catch up and notify users as such.

lieut-data (changes requested):
Thanks, @jespino! One major concern below re: 401, but also some ideation on how we can make this more bullet proof.

server/msteams/client.go
jespino: We can, but take into account that 401 (Unathorized) means actually, "Unathenticated", so... the token that you are using is not valid. Is not "you don't have permission", that would be 403. We can avoid this approach for now, and just observe it for now, if we see a lot of 401 happening, then, that would lead to random disconnections, if there are no 401, or they are anetdotical, that means that this is a valid approach. What do you think?

server/plugin.go
jespino: Fair enough, I like the idea of having it persisted and ensure it is sent.

jespino:
@lieut-data if I recall correctly, I think we decided on keeping it as "remove the session and send the message", and do some monitoring around the frequency of this happening to see if makes sense to make it more resilient.

lieut-data (approved):

jespino:
@lindalumitchell If you can test it, it would be great. For testing this I would expect that any "event" that invalidates your token should lead to a message to the user saying that the user is disconnected, and needs to reconnect his account. An example of that could be go to MSTeams account and invalidate your tokens.

lindalumitchell:
@jespino I uploaded the build artifact from this PR to https://msteams-sync-test.test.mattermost.cloud, but I'm not seeing the expected behavior after revoking a user's sessions from the Azure portal. Were you referring to a different type of revoking?

Testing notes so far:

I first tested on the older version of the plugin, just to get a baseline:
1. Revoked all of a user's sessions through Azure portal
2. Verified auto-logged out of Teams side
3. Verified *before updating plugin* in Mm that I was not notified, my posts were not syncing, and I had to use the slash command to disconnect then reconnect before my posts synced again

After updating the plugin:
1. Revoked all of a user's sessions through Azure portal
2. Verified auto-logged out of Teams side
3. Not seeing any effect on the Mm side; the user is still connected and posts are still syncing

jespino:
@lindalumitchell that is what makes me thing something is wrong with the Azure side. I tried the same and I had the same result, but in theory, the account should be invalidated and I shouldn't be sending messages to the other side. If I'm still able to send the messages to the other side is because the session is still working, so we are not testing this feature. We need a way to invalidate that token from msteams side. Let me investigate a bit more about this.

jespino:
@lindalumitchell I found this here: https://learn.microsoft.com/en-us/answers/questions/812245/invalidate-a-specific-token-from-an-application-in
imagen

jespino:
@lindalumitchell my theory now is that you have to do the revoke all sessions, wait an hour and try to send a message. I did some testing myself, and I had to fix something in the code, so get the latest code to test it. Now should be working (if you wait until the session is expired).

lindalumitchell:
Thanks so much! Updating the plugin and taking another look.

lindalumitchell:
Verified working as described! 🎉 Testing details below.

One question @jespino @lieut-data : Do we want suggestions about the wording of the message at this point? I had a couple thoughts about it. 

Currently says, on both Mm and MS Teams side:
`Your connection to MS Teams has been lost. Please reconnect to using '/msteams-sync connect' slash command.`

1. Delete "to" in "reconnect to using", or perhaps is something missing after ("reconnect to" what?)
2. Are we allowed to abbreviate to "MS Teams" or should we spell out "Microsoft"?
3. Should we use backticks instead of single quotes?
4. It might be confusing reading that message on the MS Teams side, if the user doesn't immediately know or remember how to use the slash command in Mm? 
     - Could reword a tiny bit to something like "Your connection to Microsoft Teams has been lost. Please reconnect using `/msteams-sync connect` slash command in any Mattermost channel.

Testing notes with successful disconnect notification:
- 10:26 revoked test.two through Azure portal
- Two minutes later: Still syncing posts from Mm to MS Teams
- 10:44 still syncing test.two's posts from Mm to MST
- 10:58 still syncing test.two's posts from Mm to MST
- 11:21 still syncing test.two's posts from Mm to MST
- 11:41 As soon as test.two posted in Mm, they rec'd DMs on Mm and MST side letting them know to reconnect, and post on MST side showed as not connected (Matterbridge bot)
- Reconnected using slash command in Mm and posts are syncing again as expected

Lastly: Only thing I'm not sure of is whether I had to post to trigger the DMs to appear, or whether if I had waited longer they would have appeared. I'll try again now and not touch the revoked account for longer.

lieut-data:
Yay, thank you for the detailed coverage @lindalumitchell! @cwarnermm, might you be able to comment on the textual considerations raised above? (Not urgent: I think we can iterate on this after this PR and even this release.)

lindalumitchell (approved):
Sounds good to me about potentially updating the text later. Shall I create a ticket for it?

In the meantime, I ran a second round of testing and verified that the user whose sessions have been revoked must post, specifically in a linked channel or the MS Teams bot DM, in order to trigger the notification DMs.

- Revoked test.two at 11:57am
- No notifications by 2:10pm
- Didn't notify me until I posted in a linked channel or the MS Teams bot DM, then notified me immediately

jespino:
@lindalumitchell I is kinda weird that we receive the bot message also in the MSTeams, my theory here is that MS Teams message is because the bot account is in Mattermost is connected to the bot account to the bot account in MSTeams. Maybe we should avoid propagating bot messages through platforms (or not). I going to make some of the proposed text changes, but I still get the @cwarnermm's feedback after the merge.

jespino:
@lindalumitchell also, another not. The behavior is, on any usage of the user token we get the disconnection message, that means that it waits until the token is unusable (something that we don't need to know upfront) to notify you. There is another change that will try to refresh the token of the user periodically, that also will discover this situation and send you a message. This should be exceptional behavior (We are going to add a metric to see how often this happen, but should be very uncommon).

PR #454, opened 2023-12-21
MM-56364: fix metrics stop on configuration change

lieut-data:
#### Summary
Revert https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/439, as it's incorrect to `.Shutdown()` and try to `.ServeAndListen()` on the same `http.Server`. Whenever the server configuration changes, we `reset()` and effectively kill the metrics server. In the past, we completely reset the `http.Server` on `reset()`, so it wasn't an issue, but with the new `ServeMetrics` hook, we (currently) keep around `handler` for reuse there.

Since all of this is going away once we can switch over exclusively to `ServeMetrics` in v1.6.0, ignore the fact that the revert means we no longer respond to configuration changes re: metrics. It's not an issue where we currently intend to deploy the plugin, and won't be an issue for long.

While I'm in here, update the `cloud.json` dashboard snapshot to reflect the renamed metrics.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-56364

### DIFF
--- a/server/metrics/dashboards/cloud.json
+++ b/server/metrics/dashboards/cloud.json
@@ -25,7 +25,7 @@
           "uid": "${datasource}"
         },
         "enable": true,
-        "expr": "msteams_connect_system_plugin_start_time{namespace=\"$namespace\"} * 1000",
+        "expr": "msteams_connect_system_plugin_start_timestamp_seconds{namespace=\"$namespace\"} * 1000",
         "iconColor": "green",
         "name": "Plugin Start",
         "target": {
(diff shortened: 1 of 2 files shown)


wiggin77 (approved):
Code change LGTM

lieut-data:
@lindalumitchell, I've been testing this one by hand on our test server so I don't think it needs direct QA review. We'll want to cut a v1.5.3 with these changes before releasing in January.

lindalumitchell:
Thanks @lieut-data, sounds good to me! 👍

jespino (approved):

PR #455, opened 2023-12-22
build only linux* but support devs

lieut-data:
#### Summary
Copy the changes from https://github.com/mattermost/mattermost-plugin-calls/pull/597 to slim down to just building `plugin-linux-amd64` and `plugin-linux-arm64` while still allowing developers to build for an arbitrary local platform when `MM_SERVICESETTINGS_ENABLEDEVELOPER` is true.

This will slim down the releases and make it easier to upload to test servers without bumping the max upload size.

#### Ticket Link
None.


wiggin77 (approved):

jespino (approved):
LGTM

PR #456, opened 2023-12-27
Invalid token detected metric

jespino:


server/metrics/metrics.go
```diff
@@ -227,6 +229,15 @@ func NewMetrics(info InstanceInfo) Metrics {
 	})
 	m.registry.MustRegister(m.httpErrorsTotal)
 
+	m.invalidTokenDetected = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemMSGraph,
+		Name:        "invalid_token_detected",
+		Help:        "The total number of times invalid token has been detected.",
```
lieut-data, line 236: Naming suggestion: `_total` to match Promethes best practices for counters, but also proposing we embed the idea of `oAuth` and `invalidated` for clarity and distinction from other potential tokens. Thoughts?

```suggestion
		Name:        "invalid_oauth_token_invalidated_total",
		Help: "The total number of times an invalid oAuth token has been invalidated.",
```

server/plugin.go
```diff
@@ -145,6 +145,8 @@ func (p *Plugin) GetURL() string {
 
 func (p *Plugin) OnDisconnectedTokenHandler(userID string) {
 	p.API.LogDebug("OnDisconnectedTokenHandler", "userID", userID)
+	p.metricsService.IncrementInvalidTokenDetected()
```
lieut-data, line 148: To confirm, because we set the token to `nil` via `SetUserInfo` below, we won't continuously "spam count" here? Is there a chance we end up double or triple counting in some cases, skewing the metric? Not necessary for this PR, but I wonder if we should only increment the metric if `SetUserInfo` actually changes something (using the database as a mutex for counting once).

server/plugin.go
```diff
@@ -182,9 +184,6 @@ func (p *Plugin) GetClientForUser(userID string) (msteams.Client, error) {
 	if token.Expiry.Before(time.Now()) {
 		newToken, err := client.RefreshToken(token)
 		if err != nil {
-			if msteams.IsOAuthError(err) {
-				p.OnDisconnectedTokenHandler(userID)
-			}
```
lieut-data, line 187: Is this related to the metrics changes, or a fix for something else?

lieut-data (approved):
A couple of questions / comments, but not strictly blocking.

server/metrics/metrics.go
jespino: `invalid_oauth_token_invalidated_total` sounds like a lot of "invalidation", maybe `oauth_token_invalidated_total

server/plugin.go
jespino: No, it fix the duplication of the message because after the `make generate` it adds the `Refresh` to the disconnect layer and this special case is no longer needed.

server/plugin.go
jespino: That's a good point, I don't think it is going to happen that often, but yes, we can find the way to make this run only once.

server/metrics/metrics.go
jespino: lets use this `invalid_oauth_token_invalidate_total` I going to change all the code around to make it consistent.

jespino: Done

lieut-data (approved):

lieut-data:
@jespino, I realized I typo'ed in my metric name suggestion -- heads up I just pushed a small change to avoid stuttering on `invalid`.

server/msteams/layer_generators/disconnection_layer.go.tmpl
```diff
@@ -9,31 +9,32 @@ package client_disconnectionlayer
 import (
 	"errors"
 	"io"
+	"net/http"
 	"time"
-    "net/http"
 
 	"github.com/microsoftgraph/msgraph-sdk-go/models"
+	"golang.org/x/oauth2"
```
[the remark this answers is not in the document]
jespino: It is needed for the `RefreshToken`, which is added automatically to the result file.

wiggin77 (approved):
LGTM :+1:    Left one question about a package import.

PR #457, opened 2023-12-27
Mark subscriptions as sync needed

jespino:


server/command.go
```diff
@@ -236,11 +236,16 @@ func (p *Plugin) executeLinkCommand(args *model.CommandArgs, parameters []string
 		return p.cmdError(args.UserId, args.ChannelId, "Error occurred while saving the subscription")
 	}
 
-	if err := p.store.CommitTx(tx); err != nil {
+	if err = p.store.CommitTx(tx); err != nil {
 		p.API.LogError("Unable to commit database transaction", "error", err.Error())
 		return p.cmdError(args.UserId, args.ChannelId, "Something went wrong")
 	}
 
+	err = p.store.UpdateSubscriptionSyncNeeded(channelsSubscription.ID, false)
```
wiggin77, line 244:
```suggestion
	if err = p.store.UpdateSubscriptionSyncNeeded(channelsSubscription.ID, false); err != nil {
```

server/command.go
```diff
@@ -236,11 +236,16 @@ func (p *Plugin) executeLinkCommand(args *model.CommandArgs, parameters []string
...
+	if err != nil {
```
wiggin77, line 245:
```suggestion

```

server/monitor/subscriptions.go
```diff
@@ -205,41 +219,43 @@ func (m *Monitor) recreateChannelSubscription(subscriptionID, teamID, channelID,
 
 	m.metrics.ObserveSubscription(metrics.SubscriptionReconnected)
 
-	if subscriptionID != "" {
-		if err = m.store.DeleteSubscription(subscriptionID); err != nil {
-			m.api.LogDebug("Unable to delete old channel subscription from DB", "subscriptionID", subscriptionID, "error", err.Error())
+	if subscriptionID == "" {
+		tx, err := m.store.BeginTx()
```
wiggin77, line 223: Isn't this adding to the transaction logic outside the store?  I thought we were going to undo that.

wiggin77: Ah, I see this is just continuing with the existing code.  I guess a different PR will fix it.

wiggin77 (approved):
LGTM :+1:

server/command.go
```diff
@@ -236,11 +236,16 @@ func (p *Plugin) executeLinkCommand(args *model.CommandArgs, parameters []string
...
+		p.API.LogWarn("Unable to mark the subscription as no sync needed", "subscriptionID", channelsSubscription.ID, "error", err.Error())
+	}
```
lieut-data, line 247: I would have expected this to be set to `false` in the call to `SaveChannelSubscription`. Is there a reason we don't do that?

server/monitor/subscriptions.go
```diff
@@ -178,16 +178,30 @@ func (m *Monitor) CreateAndSaveChatSubscription(mmSubscription *storemodels.Glob
...
+		}
 	}
 
 	newSubscription, err := m.client.SubscribeToUserChats(userID, m.baseURL, m.webhookSecret, !m.useEvaluationAPI, m.certificate)
 	if err != nil {
 		return err
 	}
 
-	return m.store.SaveChatSubscription(storemodels.ChatSubscription{SubscriptionID: newSubscription.ID, UserID: userID, Secret: secret, ExpiresOn: newSubscription.ExpiresOn, Certificate: m.certificate})
+	if subscriptionID == "" {
```
lieut-data, line 193: I don't understand this new check: if I pass `subscriptionID != ""` then we delete the old one and don't save the new one?

server/monitor/subscriptions.go
lieut-data: @jespino, would we consider fixing this as part of this PR? I'm finding it hard to understand the net effect of these changes by virtue of having to preserve the transaction juggling outside the store.

server/monitor/subscriptions.go
```diff
@@ -249,10 +265,19 @@ func (m *Monitor) recreateGlobalSubscription(subscriptionID, secret string) erro
 
 	m.metrics.ObserveSubscription(metrics.SubscriptionReconnected)
 
-	if err = m.store.DeleteSubscription(subscriptionID); err != nil {
-		m.api.LogDebug("Unable to delete old global subscription from DB", "subscriptionID", subscriptionID, "error", err.Error())
+	if subscriptionID == "" {
```
lieut-data, line 268: Same concern as above: seems like we don't save the new subscription if `subscriptionID == ""`?

server/store/sqlstore/store.go
```diff
@@ -233,6 +233,10 @@ func (s *SQLStore) Init() error {
 		return err
 	}
 
+	if err := s.addColumn(subscriptionsTableName, "syncNeeded", "BOOLEAN"); err != nil {
```
lieut-data, line 236: To confirm, is this `NULLABLE`? Do we handle this with the requisite `COALESCE` when querying this table? Are all existing queries explicit in which columns they return, i.e. no `SELECT *` anywhere that would break a previous version of the plugin?

server/store/sqlstore/store.go
lieut-data, line 1: I was expecting to see more of the `List*` methods updated to return this column -- is the omission intentional?

lieut-data (changes requested):
A few comments inline, but one larger proposal: I'd suggest that instead of just a boolean, we model a `syncNeededAt BIGINT` that represents the timestamp when this sync was marked as being needed.

Besides being useful debugging information, this would also allow us to choose how we prioritize in which order we sync those subscriptions.

server/monitor/subscriptions.go
jespino: This is gone after the transactions exposure is removed from the API

server/monitor/subscriptions.go
jespino: Yes, that's wrong, I don't know if was a brain fart or a merge problem, but clearly the idea is not that.

jespino: ohh, no, wait, what is going on here, is that we are deleting the "MS Teams subscription" and creating it again, but the local stored subscription is created brand new or updated based on the existing one. To ensure we preserve things like the last update at, and things like that.

server/monitor/subscriptions.go
jespino: Exact same situation, first is "client.DeleteSubscription", and later is `store.SaveGlobalSubscription`, so we are talking about different delete/save places.

server/store/sqlstore/store.go
jespino: Yes, all the queries are explicit about what they return. Probably we should make it not null and default false, let me try to fix that.

server/command.go
jespino: This was in fact something that we can remove all together, the whole function from the store even.

server/store/sqlstore/store.go
jespino: I change it to BOOLEAN NOT NULL DEFAULT False

lieut-data:
@jespino, propose we pause on these changes while we define next steps for the project.

lieut-data:
Moving in a different direction.

PR #458, opened 2023-12-27
Adding test containers support for E2E tests (using mocks only for msteams client)

jespino:


server/api.go
```diff
@@ -74,6 +85,10 @@ func NewAPI(p *Plugin, store store.Store) *API {
 	router.HandleFunc("/connected-users/download", api.getConnectedUsersFile).Methods(http.MethodGet)
 	router.HandleFunc(APIChoosePrimaryPlatform, api.choosePrimaryPlatform).Methods(http.MethodGet)
 
+	if os.Getenv("MM_MSTEAMSSYNC_MOCK_CLIENT") == "true" {
```
jespino, line 88: Adding this new mock API based on environment variable, the implementation is bellow, and the idea is that you can decide what to mock using API calls.

server/plugin.go
```diff
@@ -33,9 +35,29 @@ import (
 	"github.com/mattermost/mattermost/server/public/pluginapi/cluster"
 	"github.com/pborman/uuid"
 	"github.com/pkg/errors"
+	"github.com/stretchr/testify/mock"
 	"golang.org/x/oauth2"
 )
 
+var clientMock *mocks.Client = nil
+
+func getClientMock(p *Plugin) *mocks.Client {
```
jespino, line 44: This is the initial mock needed for the plugin to start, and also the one that is going to be used when the environment variable is there.

server/plugin.go
```diff
@@ -132,6 +154,9 @@ func (p *Plugin) GetClientForApp() msteams.Client {
 
 func (p *Plugin) GetURL() string {
 	config := p.API.GetConfig()
+	if config.ServiceSettings.SiteURL == nil {
```
jespino, line 157: The system was already panicking if there was no SiteURL in the config. Probably we should make it somehow better.

server/plugin_test.go
```diff
@@ -107,101 +200,124 @@ func getPluginPathForTest() string {
 	return path
 }
 
-func TestMessageHasBeenPostedNewMessage(t *testing.T) {
-	plugin := newTestPlugin(t)
+func TestMessageHasBeenPostedNewMessageE2E(t *testing.T) {
```
jespino, line 203: This are examples of the new tests migrated to the test-containers approach.

server/testutils/mmcontainer/testscontainer.go
```diff
@@ -0,0 +1,359 @@
+package mmcontainer
```
jespino, line 1: And ths is the implementation of the testcontainers module for mattermost, this should be usable for other plugins, maybe it should be as part of the started template or the public mattermost libraries, or another repository. I don't know, but this can be very useful for other plugins testing.

server/testutils/containers_e2e.go
```diff
@@ -0,0 +1,62 @@
+package testutils
```
jespino, line 1: Here are the helper functions to initialize the environment for this plugin, this is specific to this plugin and uses the "testcontainer mattermost module" created. Also I added the function for mocking things using the client library.

server/ce2e/plugin_test.go
```diff
@@ -0,0 +1,142 @@
+package ce2e
```
jespino, line 1: Here are the migrated examples of e2e tests for some cases.

server/iframe_test.go
```diff
@@ -110,5 +109,5 @@ func readZipFile(zipFile *zip.File) ([]byte, error) {
 	}
 	defer rc.Close()
 
-	return ioutil.ReadAll(rc)
+	return io.ReadAll(rc)
```
jespino, line 112: This is because I upgraded to 1.20 and now ioutil is deprecated

jespino:
@crspeller I added you here to review this because it is related to our discussion about how it was done in playbooks, maybe you like this approach and can be replicated there. It is not perfect, but I think for playbooks is a really good fit (in case we want to apply that in the future).

server/ce2e/iframe_test.go
```diff
@@ -0,0 +1,94 @@
+package ce2e
```
jespino, line 1: This is another example of tests migrated from mocks to using the containerized approach.

server/ce2e/plugin_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	err = store.SetUserInfo(user.Id, "ms-user-id", &oauth2.Token{})
+	require.NoError(t, err)
+
+	t.Run("Without Channel Link", func(t *testing.T) {
+		var newPost *model.Post
+		newPost, _, err = client.CreatePost(context.Background(), &post)
+		require.NoError(t, err)
+
+		time.Sleep(50 * time.Millisecond)
```
crspeller, line 49: These sleeps are an unfortunate consequence of using this method. I wonder if it would make sense to add a flag to be able to run the HasBeen hooks without a goroutine.

server/api.go
```diff
@@ -665,6 +676,283 @@ func (a *API) getConnectedUsersFile(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (a *API) resetMocks(w http.ResponseWriter, r *http.Request) {
```
crspeller, line 679: Can we add system admin checks to these? I know that environment variable should only be set when testing, but just in case.

server/api.go
```diff
@@ -665,6 +676,283 @@ func (a *API) getConnectedUsersFile(w http.ResponseWriter, r *http.Request) {
...
+	if mockCall.Err != "" {
+		returnErr = errors.New(mockCall.Err)
+	}
+
+	a.p.API.LogDebug("mocking method", "method", method, "returns", mockCall.Returns, "returnType", mockCall.ReturnType, "returnErr", returnErr)
+
+	var output any
+
+	returns := method.Type.NumOut()
+	switch returns {
```
crspeller, line 721: There is a lot of duplication here. Can we clean this up?

server/api.go
```diff
@@ -665,6 +676,283 @@ func (a *API) getConnectedUsersFile(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func (a *API) resetMocks(w http.ResponseWriter, r *http.Request) {
+	clientMock = nil
+	newMock := getClientMock(a.p)
+	a.p.msteamsAppClient = newMock
+	w.WriteHeader(http.StatusOK)
+}
+
+func (a *API) addMockCall(w http.ResponseWriter, r *http.Request) {
```
crspeller, line 686: Maybe a comment here explaining what this is doing would be nice. Also naming. Maybe `resetMSTeamsClientMocks` and `addMSTeamsClientMock`?

crspeller (changes requested):
Nice! This is great! The test containers look really easy to use. This seems to make more sense for plugins whit lighter interactions with the MM core as we discussed.

Just a couple nits.

server/ce2e/plugin_test.go
jespino: I removed the other 2 sleeps, but this one, specifically is about something "not-happening", so I prefer to keep the Sleep here anyway. Having some flag to make the hooks Sync is adding more changes in the behavior between testing and real environment, and I prefer to avoid that, at least for now.

server/api.go
jespino: Done, will be available whenever I push myt local changes.

server/api.go
jespino: Done

server/api.go
jespino: Done

crspeller (approved):

server/testutils/containere2e/containere2e.go
```diff
@@ -0,0 +1,77 @@
...
+	if len(matches) == 0 {
+		t.Fatal("Unable to find plugin tar.gz file")
+	}
+	filename := matches[0]
+
+	pluginConfig := map[string]any{
+		"clientid":                   "client-id",
+		"clientsecret":               "client-secret",
+		"connectedusersallowed":      1000,
+		"encryptionkey":              "eyPBz0mBhwfGGwce9hp4TWaYzgY7MdIB",
```
lieut-data, line 32: This looks "real" -- I'm guessing it's just random? Should we replace it with something like `encryption-key` or would that break things?

server/ce2e/iframe_test.go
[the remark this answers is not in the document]
jespino: I moved that in a separated package to have clear separation of where test are "containerized e2e tests" and what is regular tests. We can use "tags" instead, or something else. But I found useful that separation because e2e tests are way slower, so maybe you don't want to use them always.

server/testutils/containere2e/containere2e.go
jespino: Exactly, I need an "encryption key" that is valid.

server/plugin.go
jespino: Yes, we can, for the API is a bit more challenging, but we can do it, for sure.

server/ce2e/iframe_test.go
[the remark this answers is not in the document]
lieut-data: @mgdelacroix pioneered a useful approach with Go tags that let the tests live together, but easily filter down to running only some tests and not others. But we can iterate here either way :)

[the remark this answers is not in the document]
jespino: Yes, lets get this in, and we can decide later if we want to move that to flags, it shouldn't be a big deal.

lieut-data (approved):

jespino:
@lieut-data I added some changes to fix the CI execution with the new tag based compilation. PTAL

server/plugin.go
```diff
@@ -132,10 +132,14 @@ func (p *Plugin) GetClientForApp() msteams.Client {
...
-		return *config.ServiceSettings.SiteURL + "plugins/" + pluginID
+	siteURL := ""
+	if config.ServiceSettings.SiteURL != nil {
+		siteURL = *config.ServiceSettings.SiteURL
 	}
-	return *config.ServiceSettings.SiteURL + "/plugins/" + pluginID
+	if strings.HasSuffix(siteURL, "/") {
+		return siteURL + "plugins/" + pluginID
+	}
+	return siteURL + "/plugins/" + pluginID
```
lieut-data, line 142:
```suggestion
	return url.JoinPath(siteURL, "plugins", pluginID)
```

lieut-data (approved):

PR #460, opened 2023-12-29
Adding whitelist limit metric

jespino:
This adds the limit of allowed connected users to the metrics to easily have control over it through grafana.

This closes the ticket [MM-56091](https://mattermost.atlassian.net/browse/MM-56091)


wiggin77 (approved):
LGTM :+1:

server/metrics/metrics.go
```diff
@@ -148,6 +151,16 @@ func NewMetrics(info InstanceInfo) Metrics {
 	m.pluginStartTime.SetToCurrentTime()
 	m.registry.MustRegister(m.pluginStartTime)
 
+	m.whitelistLimit = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemApp,
+		Name:        "whitelist_limit",
+		Help:        "The number of allowed connected users by configuration.",
```
lieut-data, line 158:
```suggestion
		Help:        "The maximum number of users allowed to connect.",
```

lieut-data (approved):
Just one wording proposal. LGTM!

jespino:
@lindalumitchell this is a new metric added for showing number of whitelist limit on grafana metrics.

To test this you only need to see if the new metric msteams_connect_app_whitelist_limit is in the exposed metrics (port :9094, for example http://localhost:9094).

server/configuration.go
```diff
@@ -166,7 +166,10 @@ func (p *Plugin) OnConfigurationChange() error {
 	}
 
 	p.setConfiguration(configuration)
-	p.metricsService.ObserveWhitelistLimit(configuration.ConnectedUsersAllowed)
+
+	if p.metricsService != nil {
+		p.metricsService.ObserveWhitelistLimit(configuration.ConnectedUsersAllowed)
+	}
```
lieut-data, line 172: Should we just put this in `p.start()` which will get called both on startup and on a configuration change by virtue of `p.restart()` below?

jespino: Good call, let me change it.

jespino: Done

lindalumitchell:
Thanks @jespino! I found the compiled plugin in the pipeline for this PR and uploaded it to a test server, but I'm not familiar with viewing the metrics. I'm poking around in that test server's logs in Grafana, the server logs in-app, and also checked in dev tools in the browser, just making stabs at where I might verify this. So.... where should I be able to verify this? Does it require a valid connection to an MS Teams instance, or should just enabling the plugin be enough to cause the relevant elements to be present?

lieut-data:
> Thanks @jespino! I found the compiled plugin in the pipeline for this PR and uploaded it to a test server, but I'm not familiar with viewing the metrics. I'm poking around in that test server's logs in Grafana, the server logs in-app, and also checked in dev tools in the browser, just making stabs at where I might verify this. So.... where should I be able to verify this? Does it require a valid connection to an MS Teams instance, or should just enabling the plugin be enough to cause the relevant elements to be present?

@lindalumitchell, do you have access to grafana.internal.mattermost.com? If we upload the plugin to our msteams-sync instance, it should be possible to verify there... but this is one of those tickets where I'd recommend we skip QA testing of the metric itself. Provided the plugin starts up without issue, I think we can better evaluate the "correctness" of this metric once we update our Grafana dashboard to reflect same.

jespino:
@lindalumitchell if you are ok with defering this until is merged, I'm totally fine too.

lindalumitchell (approved):
Yeah, let's do that.

PR #467, opened 2024-01-04
Initial usage of share channels api (Only triggering metrics)

jespino:
This PR adds the usage of the shared channels API in the plugin and start using the metrics for tracking how many "events" are generated from each side (hooks and shared channels).

This PR doesn't include the subscription to the direct messages based on "any connected user on it", that would be a follow up PR.


server/plugin.go
```diff
@@ -426,6 +428,36 @@ func (p *Plugin) OnActivate() error {
 		}
 	}
 
+	remoteID, err := p.API.RegisterPluginForSharedChannels(model.RegisterPluginOpts{
```
jespino, line 431: It is important to notice that I'm registering everything here, and adding the hooks, and subscribing to channels, so the flag is just enabling and disabling the metric? Should we enable/disable the whole thing based on the configuration setting for monitoring?

wiggin77: There is already a setting to turn of shared channels, in which case the registration will fail. IMO that is enough should we want to disable it.  However, that means OnActivate will fail and the plugin will not run.  I think we need to decide if disabled shared channels is a fatal plugin error.  Today, probably not a fatal error, but perhaps in the future.   We could also look at the plugin enabling shared channels as needed.

server/plugin.go
```diff
@@ -426,6 +428,36 @@ func (p *Plugin) OnActivate() error {
...
+	}
+	p.remoteID = remoteID
+
+	linkedChannels, err := p.store.ListChannelLinks()
+	if err != nil {
+		return err
+	}
+	for _, linkedChannel := range linkedChannels {
+		p.API.LogDebug("LINKED CHANNEL", "channelID", linkedChannel.MattermostChannelID, "teamID", linkedChannel.MattermostTeamID, "remoteID", p.remoteID)
+		// TODO: Handling error here or using a better policy about re-share channels
```
wiggin77, line 449: Sharing an already shared channel was made idempotent with no error.  See https://github.com/mattermost/mattermost/blob/eee5a71e3d18d04b1b136e44cd8d848fd6e3bea9/server/channels/app/plugin_api.go#L1309

wiggin77: ... meaning, if there is an error here, it is a real error such as invalid channel_id or database error.

server/plugin.go
```diff
@@ -757,3 +789,40 @@ func (p *Plugin) runMetricsUpdaterTask(store store.Store, updateMetricsTaskFrequ
 	metricsUpdater()
 	model.CreateRecurringTask("metricsUpdater", metricsUpdater, updateMetricsTaskFrequency)
 }
+
+func (p *Plugin) OnSharedChannelsPing(rc *model.RemoteCluster) bool {
+	p.API.LogDebug("Ping received", "remote", rc.DisplayName, "pluginID", pluginID)
+	return true
+}
+
+func (p *Plugin) OnSharedChannelsSyncMsg(msg *model.SyncMsg, rc *model.RemoteCluster) (model.SyncResponse, error) {
```
wiggin77, line 798: Should we start counting file attachments even though we won't implement until after June?

wiggin77:
```
func (p *Plugin) OnSharedChannelsAttachmentSyncMsg(fi *model.FileInfo, post *model.Post, rc *model.RemoteCluster) error {

}
```

plugin.json
```diff
@@ -128,6 +128,12 @@
             "type": "bool",
             "help_text": "When true, synthetic users will be converted to members when they login for the first time.",
             "default": false
+        },{
+            "key": "metricsForSharedChannelsInfrastructure",
+            "display_name": "Add metrics Shared Channels Infrastructure (Not using it yet)",
```
wiggin77, line 133:
```suggestion
            "display_name": "Add metrics Shared Channels Service (Not used yet)",
```

plugin.json
```diff
@@ -128,6 +128,12 @@
...
+            "type": "bool",
+            "help_text": "When true, the system collect metrics about Shared Channels infrastructure usage by the plugin without using it.",
```
wiggin77, line 135:
```suggestion
            "help_text": "When true, the system collect metrics about Shared Channels Service usage by the plugin without using it.",
```

server/plugin.go
wiggin77: Checking the code, registration will succeed regardless of whether shared channels service is enabled.  If disabled you simply won't get any sync messages.

wiggin77 (approved):
LGTM :+1:   Few questions/comments.

server/plugin.go
jespino: Let me add it

server/plugin.go
jespino: Then I going to add a "warning" log in case of error there.

server/plugin.go
lieut-data: > shared channels service is enabled

Ah, this is interesting @wiggin77. Can you elaborate on what's required to enable? Does enabling force the user to enable the full shared channels experience, or can we selectively enable this just for plugins?

go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/gosimple/slug v1.13.1
 	github.com/jmoiron/sqlx v1.3.5
 	github.com/lib/pq v1.10.9
-	github.com/mattermost/mattermost/server/public v0.0.12
+	github.com/mattermost/mattermost/server/public v0.0.13-0.20240123143303-e98aa55ad739
```
lieut-data, line 16: Let's be sure to tag a public module and bump the version before merging this.

plugin.json
```diff
@@ -128,6 +128,12 @@
             "type": "bool",
             "help_text": "When true, synthetic users will be converted to members when they login for the first time.",
             "default": false
+        },{
+            "key": "metricsForSharedChannelsInfrastructure",
+            "display_name": "Add metrics Shared Channels Service (Not used yet)",
+            "type": "bool",
+            "help_text": "When true, the system collect metrics about Shared Channels Service usage by the plugin without using it.",
+            "default": false
```
lieut-data, line 136: Can we default this to `true`? We ran into an issue recently where "new" config didn't get written to `config.json`, and then we couldn't manipulate it via `mmctl` until it had been set at least once. In this case, I think it's fine to default on and monitor as we rollout, and only turn it off if need be.

```suggestion
            "default": true
```

plugin.json
```diff
@@ -128,6 +128,12 @@
             "type": "bool",
             "help_text": "When true, synthetic users will be converted to members when they login for the first time.",
             "default": false
+        },{
+            "key": "metricsForSharedChannelsInfrastructure",
+            "display_name": "Add metrics Shared Channels Service (Not used yet)",
+            "type": "bool",
+            "help_text": "When true, the system collect metrics about Shared Channels Service usage by the plugin without using it.",
```
lieut-data, line 135: Propose avoiding the term "shared channels" in any user-facing setting. 0/5 on the actual term used, just trying to steer clear of the feature with the same name. I guess this configuration isn't long lived anyway, so we don't have to worry about a long term name.

```suggestion
            "key": "metricsForSharedChannelsInfrastructure",
            "display_name": "Initialize and monitor the durable queue service",
            "type": "bool",
            "help_text": "When true, the the durable queue service is enabled with metrics to evaluate the overhead. Message delivery is unaffected.",
```

server/command.go
```diff
@@ -221,6 +221,22 @@ func (p *Plugin) executeLinkCommand(args *model.CommandArgs, parameters []string
 		return p.cmdError(args.UserId, args.ChannelId, "Error occurred while saving the subscription")
 	}
 
+	if _, err = p.API.ShareChannel(&model.SharedChannel{
```
lieut-data, line 224: Should this be conditional on the configuration setting?

server/command.go
```diff
@@ -221,6 +221,22 @@ func (p *Plugin) executeLinkCommand(args *model.CommandArgs, parameters []string
...
+		Home:      true,
+		ReadOnly:  false,
+		CreatorId: p.userID,
+		RemoteId:  p.remoteID,
+		ShareName: channelLink.MattermostChannelID,
+	}); err != nil {
+		if err2 := p.store.DeleteSubscription(channelsSubscription.ID); err2 != nil {
+			p.API.LogDebug("Unable to rollback the subscription creation", "channelID", channelLink.MattermostChannelID, "error", err2.Error())
+		}
+		p.API.LogDebug("Unable to share the channel", "channelID", channelLink.MattermostChannelID, "error", err.Error())
```
lieut-data, line 236: Let's log this as a warning:
```suggestion
		p.API.LogWarn("Unable to share the channel", "channelID", channelLink.MattermostChannelID, "error", err.Error())
```

server/command.go
```diff
@@ -221,6 +221,22 @@ func (p *Plugin) executeLinkCommand(args *model.CommandArgs, parameters []string
...
+		TeamId:    channelLink.MattermostTeamID,
+		Home:      true,
+		ReadOnly:  false,
+		CreatorId: p.userID,
+		RemoteId:  p.remoteID,
+		ShareName: channelLink.MattermostChannelID,
+	}); err != nil {
+		if err2 := p.store.DeleteSubscription(channelsSubscription.ID); err2 != nil {
+			p.API.LogDebug("Unable to rollback the subscription creation", "channelID", channelLink.MattermostChannelID, "error", err2.Error())
+		}
```
lieut-data, line 235: Given we're not relying on this shared channels feature yet, I'm wondering if we shouldn't be destructive here and deleting the subscription?

```suggestion
```

server/command.go
```diff
@@ -264,10 +278,20 @@ func (p *Plugin) executeUnlinkCommand(args *model.CommandArgs) (*model.CommandRe
 		return &model.CommandResponse{}, nil
 	}
 
+	if _, err = p.API.UnshareChannel(link.MattermostChannelID); err != nil {
```
lieut-data, line 281: Should this be conditional on the configuration setting?

server/command.go
```diff
@@ -264,10 +278,20 @@ func (p *Plugin) executeUnlinkCommand(args *model.CommandArgs) (*model.CommandRe
...
+		if err2 := p.store.SaveChannelSubscription(*subscription); err2 != nil {
+			p.API.LogDebug("Unable to rollback the subscription deletion", "subscriptionID", subscription.SubscriptionID, "error", err2.Error())
+		}
```
lieut-data, line 284: See above for probably doing nothing in this case, but even if this was "live", I'm wondering why we would try to rollback the subscription here? Presumably we should still delete the subscription?

server/command.go
```diff
@@ -264,10 +278,20 @@ func (p *Plugin) executeUnlinkCommand(args *model.CommandArgs) (*model.CommandRe
...
+		p.API.LogDebug("Unable to unshare the channel", "channelID", link.MattermostChannelID, "error", err.Error())
```
lieut-data, line 285: If we do keep this, let's log as a warning.
```suggestion
		p.API.LogWarn("Unable to unshare the channel", "channelID", link.MattermostChannelID, "error", err.Error())
```

server/message_hooks.go
```diff
@@ -38,27 +38,32 @@ func (p *Plugin) UserWillLogIn(_ *plugin.Context, user *model.User) string {
 }
 
 func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
+	p.GetMetrics().ObserveMessageHooksEvent("message_create")
+	_ = p.syncMessage(post)
```
lieut-data, line 42: Can we handle this error and at least log it?
```suggestion
	if err := p.syncMessage(post); err != nil {
		p.API.LogWarn("Failed to handle MessageHasBeenPosted", "postID", post.ID, "error", err.Error())
	}
```

server/message_hooks.go
```diff
@@ -38,27 +38,32 @@ func (p *Plugin) UserWillLogIn(_ *plugin.Context, user *model.User) string {
 }
 
 func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
+	p.GetMetrics().ObserveMessageHooksEvent("message_create")
```
lieut-data, line 41: Counting here will always be out of sync with the shared channels system, since we get the event for all messages whether or not we process them, whereas shared channels is only for things to which we subscribe. Should we try to push this metric down further?

server/message_hooks.go
```diff
@@ -69,106 +74,127 @@ func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
...
 	_, err = p.Send(link.MSTeamsTeam, link.MSTeamsChannel, user, post)
 	if err != nil {
 		p.API.LogWarn("Unable to handle message sent", "error", err.Error())
+		return err
 	}
+	return nil
 }
 
 func (p *Plugin) ReactionHasBeenAdded(c *plugin.Context, reaction *model.Reaction) {
+	p.GetMetrics().ObserveMessageHooksEvent("reaction_added")
```
lieut-data, line 91: See above re: where to measure this. (I won't comment everywhere this applies.)

server/message_hooks.go
```diff
@@ -69,106 +74,127 @@ func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
...
 	updateRequired := true
 	if c.RequestId == "" {
 		_, ignoreHookForReaction := p.activityHandler.IgnorePluginHooksMap.LoadAndDelete(fmt.Sprintf("%s_%s_%s", reaction.PostId, reaction.UserId, reaction.EmojiName))
 		updateRequired = !ignoreHookForReaction
 	}
+	_ = p.syncReactionAdded(reaction, updateRequired)
```
lieut-data, line 97: See above for proposal on handling this error. (I won't comment everywhere this applies.)

server/metrics/metrics.go
```diff
@@ -291,6 +298,24 @@ func NewMetrics(info InstanceInfo) Metrics {
 	}, []string{"action"})
 	m.registry.MustRegister(m.subscriptionsTotal)
 
+	m.messageHooksEvents = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemHooks,
+		Name:        "events_total",
+		Help:        "The total number of message hooks events (message creation and update, reactions creation and update).",
```
lieut-data, line 305: nit: propose letting the labels expand on the help here
```suggestion
		Help:        "The total number of hook events related to message processing.",
```

server/metrics/metrics.go
```diff
@@ -291,6 +298,24 @@ func NewMetrics(info InstanceInfo) Metrics {
...
+		ConstLabels: additionalLabels,
+	}, []string{"event"})
+	m.registry.MustRegister(m.messageHooksEvents)
+
+	m.messageSharedChannelsEvents = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemSharedChannels,
+		Name:        "events_total",
+		Help:        "The total number of message events through shared channels (message creation and update, reactions creation and update).",
```
lieut-data, line 314:
```suggestion
		Help: "The total number of shared channel events related to message processing.",
```

server/message_hooks.go
lieut-data, line 1: Given we're not actually reusing the new helper methods yet, I wonder if we could defer those changes vs. including them here?

server/plugin.go
```diff
@@ -426,6 +428,38 @@ func (p *Plugin) OnActivate() error {
 		}
 	}
 
+	remoteID, err := p.API.RegisterPluginForSharedChannels(model.RegisterPluginOpts{
+		Displayname:  "ms_teams_connect",
```
lieut-data, line 432: Propose matching the pending name:
```suggestion
		Displayname: "MS Teams Plugin",
```

server/plugin.go
```diff
@@ -426,6 +428,38 @@ func (p *Plugin) OnActivate() error {
...
+		PluginID:     pluginID,
+		CreatorID:    botID,
+		AutoShareDMs: true,
+		AutoInvited:  true,
```
lieut-data, line 436: Can you remind me what this does?

server/plugin.go
```diff
@@ -757,3 +791,49 @@ func (p *Plugin) runMetricsUpdaterTask(store store.Store, updateMetricsTaskFrequ
...
 	model.CreateRecurringTask("metricsUpdater", metricsUpdater, updateMetricsTaskFrequency)
 }
+
+func (p *Plugin) OnSharedChannelsPing(rc *model.RemoteCluster) bool {
+	p.API.LogDebug("Ping received", "remote", rc.DisplayName, "pluginID", pluginID)
+	return true
+}
+
+func (p *Plugin) OnSharedChannelsAttachmentSyncMsg(fi *model.FileInfo, post *model.Post, rc *model.RemoteCluster) error {
+	p.API.LogDebug("SyncMsg attachment", "fileInfo", fi, "post", post, "remote", rc.DisplayName, "pluginID", pluginID)
```
lieut-data, line 801: Propose not logging this: will be too spammy and not useful.

server/plugin.go
```diff
@@ -757,3 +791,49 @@ func (p *Plugin) runMetricsUpdaterTask(store store.Store, updateMetricsTaskFrequ
...
+	if !p.configuration.MetricsForSharedChannelsInfrastructure {
+		return nil
+	}
+	p.GetMetrics().ObserveMessageSharedChannelsEvent("attachment_created")
```
lieut-data, line 805: Wondering what the point of the "feature flag" here is? Feels like we should be turning it on or off depending on the feature flag, and always measuring metrics. If we can't reliably "turn it off", maybe the feature flag is itself not useful?

server/command.go
lieut-data, line 1: I'm wondering if we should consider skipping shared channels for channels, given our focus on DMs and GMs. There are edge cases here like what to do with existing subscriptions that we're not solving for, and I worry that a half-baked approach might be more confusing than adding support for shared channels explicitly later. Thoughts?

lieut-data (changes requested):
Thanks, @jespino! A few thoughts as we figure out how to roll this out and start measuring impact.

@wiggin77, could you also refresh me as to what metrics shared channels itself already exposes and how we might want to start monitoring that from a Grafana dashboard?

server/command.go
jespino: Not necesarilly, if you enable the shared channels it would work, if not it shouldn't be a problem. This way we exercise more machinery during this changes.

server/command.go
jespino: The amount of complexity of including shared channels for channels is minimal, I don't see much value on that for this specific case and for now. Let's see how it behaves, and as soon as we see a rock in the road we can disable channels, but for now, I don't feel the need.

server/command.go
jespino: This is not "deleting the subscription", this is "undoing the channel linking done before in case of error"

server/command.go
jespino: I don't think we need to make it conditional.

server/command.go
jespino: This is because it is consider that the whole operation has fail, I think this is the correct behavior. You are trying to "unshare" the channel, and because you were unable, you share it again.

server/message_hooks.go
jespino: Yes, I'll take a look

server/message_hooks.go
jespino: Good point

server/plugin.go
jespino: This doesn't require to go through the Invite process that shared channels require.

server/plugin.go
jespino: We are deciding to execute or not the metrics based on the config setting, but as you said, if there is not messages flowing because the shred channels features is off that maybe is not needed.

server/message_hooks.go
jespino: I guess so

server/command.go
lieut-data: If shared channels was a subsystem we were relying on, I agree this code makes sense (though I'd suggest we invert and start sharing the channel first). But for the purpose of this PR -- an experiment to measure the overhead of shared channels -- we're directly tying the functionality of that core to the enduser's experience, breaking channel linking if shared channels doesn't work for some reason. Should we narrow the scope of this PR to "just an experiment" with 0 expected impact on the enduser?

cc @jespino / @wiggin77 as I sense this is a broader question for what we want to achieve here.

server/message_hooks.go
jespino: Done

server/message_hooks.go
jespino: Done

server/message_hooks.go
jespino: Done

server/plugin.go

server/command.go
lieut-data: I'm not sure I'd expect those semantics as an enduser. I agree that in a purely transactional sense, this is ideal -- but the inability to stop sharing a channel in a subsystem internal to us shouldn't preclude disconnecting the MS Teams side. That succeeded and severs the link, even if we have to fix issue internally.

(In other words, I'd rather we harden the code to deal with shared channels that have no matching subscription, than force this error onto the enduser and prevent them from unlinking in the first place.)

server/command.go
lieut-data: So to be clear, we can't really disable shared channels with this PR -- we can disable collecting metrics, and we can disable the shared channels subsystem, but all the other side effects (e.g. failure to unshare a channel above) aren't feature flagged and can't be "rolled back" if they go wrong.

I'm advocating for not adding partial code unless we're going to spend the effort to productize it -- as it stands right now, we have no plan to figure out what to do with these older linked channels, and probably won't revisit until after June. It would be great if we could keep the total surface area of complexity as low as possible.

jespino: I think we are more or less align on allowing to share channels but disabling that feature optionally, giving us less surface exposure to the early adopters but giving the possibility to some of them to enable channels and start testing it.

server/command.go
jespino: I going to put this behind the config flag to be able to reduce surface here while the metrics are off. But that can leave the system in a weird state where only after restart some channels are really shared with Shared Channels (not very important for now).

lieut-data:
Heads up I'm looking into the build failures here, @amyblais, and may try to include a narrower scope of this PR into tomorrow's v1.7.0 build.

lieut-data:
@amyblais, I've patched the build issues, but am running into other things that make me think we should push this to v1.8 instead.

lieut-data:
@jespino & @wiggin77, I made some updates to this pull request to focus the scope, mostly around measuring latency bewteen when we process the sync msg relative to when the post was created. I also updated the unit tests, opting to effect some of my simplifications above to make the job of testing easier.

I note that this won't "do" anything in environments where we aren't supporting linked channels, but should still be good fodder for our testing infrastructure.

lieut-data:
One oddity is that I don't seem to be getting calls to `OnSharedChannelsAttachmentSyncMsg` -- or at least the metrics aren't capturing anything. I haven't debugged yet, but wondering if you have any suggestions @wiggin77?

CleanShot 2024-02-09 at 23 06 26@2x

wiggin77:
> I don't seem to be getting calls to OnSharedChannelsAttachmentSyncMsg

I just double checked my test plugin that logs all the hooks calls and it is getting called.  Perhaps metrics are not hooked up there.

lieut-data (approved):

wiggin77 (approved):
LGTM :+1:

PR #469, opened 2024-01-09
fix release notes url generation

lieut-data:
#### Summary
Fix this auto generation:
```diff
-    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-msteams-syncreleases/tag/v1.5.3",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-msteams-sync/releases/tag/v1.5.3",
```

#### Ticket Link
None.

### DIFF
--- a/build/manifest/main.go
+++ b/build/manifest/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"os"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


mickmister (approved):

build/manifest/main.go
```diff
@@ -145,7 +146,10 @@ func findManifest() (*model.Manifest, error) {
 
 	// Update the release notes url to point at the latest tag, if present.
 	if BuildTagLatest != "" {
-		manifest.ReleaseNotesURL = manifest.HomepageURL + "releases/tag/" + BuildTagLatest
+		manifest.ReleaseNotesURL, err = url.JoinPath(manifest.HomepageURL, "releases/tag", BuildTagLatest)
```
jespino, line 149: I'm not 100% sure, but maybe `"release/tag"` can be `"release", "tag", and that way it would work even in windows (I don't think is needed, but there is no harm in it.

jespino (approved):
LGTM. Just a nit comment, but I don't think is needed.

build/manifest/main.go
lieut-data: Thanks, @jespino! Can you expand on the Windows perspective to help me understand -- this is a URL, so was thinking it should always be `/`, but you're right that we could just have `url.JoinPath` "do that" instead of manually composing the two.

jespino: Totally right, this is a URL, so my comments makes no sense.

PR #471, opened 2024-01-17
Adding client secret expire metric

jespino:
This PR adds a metric for keep monitor when the different secrets for the Microsoft Teams app Expires. This will help to detect client secrets that are about to expire.

Fixes: https://mattermost.atlassian.net/browse/MM-56492


jespino:
This is a draft, I want to add testing and polish some things, like the way I check the system_admin role, or how I decide to send or not a message, also we should add the metrics for this.

jespino:
@lieut-data still a WIP but if you want you to take a look to the approach, to pay too much attention to the details because I have to change things.

wiggin77 (approved):
LGTM

server/monitor/credentials.go
```diff
@@ -0,0 +1,21 @@
+package monitor
+
+import (
+	"time"
+)
+
+func (m *Monitor) checkCredentials(force bool) {
+	if !force && (time.Now().Hour() != 0 || time.Now().Minute() != 0) {
+		return
+	}
```
lieut-data, line 10: This feels a bit kludgy: would we consider using a discrete [job](https://github.com/mattermost/mattermost/blob/master/server/public/pluginapi/cluster/job.go) that runs once an hour vs. adding it to the `monitor`?

server/monitor/credentials.go
```diff
@@ -0,0 +1,21 @@
...
+
+	credentials, err := m.client.GetAppCredentials(m.applicationID)
+	if err != nil {
+		m.api.LogDebug("Error getting client secret expire date", "error", err.Error())
```
lieut-data, line 14: Let's log this as a warning if indeed the request is failing:
```suggestion
		m.api.LogWarn("Error getting client secret expire date", "error", err.Error())
```

server/monitor/credentials.go
```diff
@@ -0,0 +1,21 @@
...
+		return
+	}
+
+	for _, credential := range credentials {
+		m.metrics.ObserveClientSecretExpireDate(credential.ID, credential.ExpireDate)
+	}
```
lieut-data, line 20: I realize this is likely to be finite and very small in practice, but now that I see the idea we discussed in practice, it feels potentially problematic to create a discrete time series for each `credential.ID`. 

Would we consider sorting the `credentials` by `ExpireDate` (ascending, so we see the expired / next to expire first), and then reporting on at most `5`? We could then name the labels programmatically as `credential_0`, bounding the total number of time series. As credentials are removed, we would see the times shift, but this is probably sufficient fidelity for our purposes.

server/msteams/client.go
```diff
@@ -280,6 +280,23 @@ func (tc *ClientImpl) RefreshToken(token *oauth2.Token) (*oauth2.Token, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	credentials := []clientmodels.Credential{}
+	credentialsList := application.GetPasswordCredentials()
+	for _, credential := range credentialsList {
+		credentials = append(credentials, clientmodels.Credential{
+			ID:         credential.GetKeyId().String(),
+			Name:       *credential.GetDisplayName(),
+			ExpireDate: *credential.GetEndDateTime(),
```
lieut-data, line 294: nit: instead of `ExpireDate`, should we translate the field (and related APIs in the metrics) more directly as `EndDateTime`? I always like grepping that "finds everything" vs. having a translation mid-way that obscures things.

server/monitor/credentials.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"time"
+)
+
+func (m *Monitor) checkCredentials(force bool) {
+	if !force && (time.Now().Hour() != 0 || time.Now().Minute() != 0) {
+		return
+	}
+
+	credentials, err := m.client.GetAppCredentials(m.applicationID)
```
lieut-data, line 12: Can we skip trying if `applicationID` is empty? (With a debug noting same.) It would also be ideal if we could distinguish a generic error from one that relates to not having the credentials to make this request. Is this possible?

server/monitor/credentials.go
jespino: I can move it to an independent job.

server/monitor/credentials.go
jespino: That sounds like some extra complexity that I don't see the need for. From my perspective you should only have 1 credential, and in the worst case scenario, you are going to renew that credentials every 3 months. So I don't think is going to be that many metrics. Can you elaborate more what is the problem of having multiple metrics? from my perspective, your are not even querying for the whole history, you are probably query for the current credentials, we can even stop reporting credentials that are expired for more than a month, or something like that (I don't think is needed anyway).

server/msteams/client.go
jespino: Done

server/monitor/credentials.go
jespino: Done

server/monitor/credentials.go
jespino: Done

server/monitor/credentials.go
jespino: Adding it to a job now requires extra complexity and I don't see any problem on hooking it as part of the monitoring. If you feel strong about that, we can discuss it, but I don't think it necessarily belongs to a different job.

server/monitor/credentials.go
lieut-data: Every new `credential.ID` creates a new Prometheus timeseries. As I understand it, this is like a brand new `CREATE TABLE`. Even if you don't add new data after a while, I understand it to be best practice to avoid a potentially unbounded input set.

You are right, @jespino, that in practice, this is unlikely to be problematic. I also wonder if maybe our idea of a metric for this was altogether unnecessary in the first place, given we have to ask customers to give the application extra permissions.

Thoughts about just abandoning this in favour of a plan to contact specific customers re: credential expirations, and do more battle testing on the plugin when those credentials stop working (plus resiliency to recover) so we can help customers who forget to transition across this threshold?

jespino: I'm ok with both approaches, but I don't think that approach is very "forgiving", and is not very scalable. Having this properly monitored gives you a clear way to act on that. Not having that monitored makes this error-prone. And can end up leading to down times in the plugin.

I see the drawback on adding an extra permission requirement, and I see the drawback on adding one credential ID for certificate. But, I don't see the point on removing that. If you don't give the permission I'm not going to create any metric, so, that has no impact on prometheus, if you have the permission, the metric is going to be there and isgoing to be providing value.

Another option is to only show the one that is more "expired" or more near to expire. That way we can directly communicate that and they have to remove that to ensure that we are showing the right expire date.

lieut-data (approved):

wiggin77 (approved):
LGTM :+1:

PR #472, opened 2024-01-17
initialize API only once

lieut-data:
#### Summary
Don't initialize the API on every single HTTP request. Besides the performance overhead, this is spamming the plugin API to `GetBundlePath` after upgrading to v1.5.3.:

image

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-55489

### DIFF
--- a/server/api_test.go
+++ b/server/api_test.go
@@ -914,7 +914,6 @@ func TestConnect(t *testing.T) {
 			}
 
 			mockAPI := &plugintest.API{}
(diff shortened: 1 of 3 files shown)


jespino (approved):
Good Catch.

sbishel (approved):

lieut-data:
@lindalumitchell, I merged this one a bit too quick before syncing with you, but proposing that unit tests should cover this and we'll want to test further on our testing instance where metrics are enabled.

PR #477, opened 2024-01-27
MM-56496/MM-56497 Automute linked channels when a user sets their primary platform to Teams

hmhealey:
#### Summary
This is the first chunk of the automuting logic. When the user sets their primary platform to Teams, it mutes all their linked channels and DMs/GMs, and when it changes back to Mattermost, it unmutes them all. It doesn't yet support muting channels that become linked, channels the user is added to, or new DM channels, and it also doesn't change the muting when a user connects or disconnects.

Some implementation details:

1. When the users' primary platform is set AND they're connected, all the channels are muted and we set the an `automute_enabled` preference to track that we're in that case. Tracking that as a separate preference does mean we have a second source of truth for if automuting is enabled (as opposed to just checking the conditions separately), but it means that we can more easily tell when that condition has changed to avoid muting and unmuting channels which were already muted or unmuted.
2. Automuted channel members have an extra `msteams_automute` notify prop which can be used to identify those. It's kind of extra data, but it lets us identify those channel members if we need to in the future.
3. I spent a ton of time fighting with how to test this code because I kept having to mock every plugin API method which made it feel like I wasn't testing anything worthwhile. For example, having to mock both the methods to set a preference and get a preference made the tests feel really untrustworthy, so I instead ended up implementing a more functional server mock for what I needed. 

    So far, that mock keeps track of preferences, channels, and channel members for the APIs that I needed. I'm happy with how it works since it feels more authentic when I can call the plugin API to set up test state, I don't need to repeatedly mock the same lower level method calls, and I can confirm the results by calling the same plugin API. That said, I don't like writing something that works completely different for my first commit on this repo, and I'm fully ready to switch back to manual mocking if this causes a bunch of issues just by being different. Let me know what you think about that part.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56496
https://mattermost.atlassian.net/browse/MM-56497


server/automute.go
```diff
@@ -0,0 +1,214 @@
...
+
+	var teamsPrimary bool
+	if skipPrimaryPlatformCheck {
+		teamsPrimary = true
+	} else {
+		teamsPrimary = p.isUsersPrimaryPlatformTeams(userID)
+	}
+
+	if !connected || !teamsPrimary {
+		return false, nil
```
larkox, line 55: 0/5 on the performance value of it (not sure how often will this function be called) but we can early return once we have the `connected` value.

server/automute.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	var membersToMute []*model.ChannelMemberIdentifier
+
+	if channels, appErr := p.API.GetChannelsForTeamForUser("", userID, true); appErr != nil {
+		return false, errors.Wrap(appErr, fmt.Sprintf("Unable to get channels for user %s to automute them", userID))
+	} else {
+		for _, channel := range channels {
+			if linked, err := p.canAutomuteChannel(channel); err != nil {
+				return false, err
```
larkox, line 78: Do we want to return potentially half way through? may make sense, but double checking.

server/automute.go
```diff
@@ -0,0 +1,214 @@
...
+		return false, nil
+	}
+
+	var membersToMute []*model.ChannelMemberIdentifier
+
+	if channels, appErr := p.API.GetChannelsForTeamForUser("", userID, true); appErr != nil {
+		return false, errors.Wrap(appErr, fmt.Sprintf("Unable to get channels for user %s to automute them", userID))
+	} else {
+		for _, channel := range channels {
+			if linked, err := p.canAutomuteChannel(channel); err != nil {
```
larkox, line 77: This can be expensive since (if I am not mistaken), potentially it does one call to the KV store for each channel. Probably nothing we have to worry about now, but let's keep it in mind in case it becomes a hotspot.

server/automute.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+	}
+
+	if err := p.setChannelMembersAutomuted(membersToMute, automuteEnabled); err != nil {
+		return false, errors.Wrap(err, fmt.Sprintf("Unable to automute all channels for linked user %s", userID))
+	}
+
+	if err := p.setAutomuteIsEnabledForUser(userID, automuteEnabled); err != nil {
+		return false, err
+	}
```
larkox, line 93: Any potential race condition because of doing this at the end? Hypothetical scenario that comes to mind:
- Change from false to true. The process (A) start.
- Change back to false. The new process (B) will stop at `p.getAutomuteIsEnabledForUser`
- (A) finishes, and set this to true, even if the user thought they put it to false

May be too corner case to consider, but wanted to raise it either way.

larkox: I have been thinking that moving this piece of code somewhere else would only create a different race condition (potentially worse). If we were to address this at any point, we would have to use some kind of HA mutex (write into the KV store, for example). And even then, we would still have problems :sweat_smile:

server/automute_test.go
```diff
@@ -0,0 +1,451 @@
...
+		preferences:    make(map[string]model.Preference),
+		t:              t,
+	}
+	t.Cleanup(func() { mockAPI.AssertExpectations(t) })
+
+	p := newTestPlugin(t)
+	p.SetAPI(mockAPI)
+
+	return p
+}
```
larkox, line 133: I would love to not need this, but I guess is the state we are at with plugin testing.

server/automute_test.go
```diff
@@ -0,0 +1,451 @@
...
+
+		setPrimaryPlatform(p, user.Id, PreferenceValuePlatformMM)
+
+		result, err := p.shouldEnableAutomuteForUser(user.Id, false, false)
+
+		assert.Equal(t, false, result)
+		assert.NoError(t, err)
+	})
+
+	t.Run("should return false when a user is not connected and has their primary platform set to MM", func(t *testing.T) {
```
larkox, line 177: Maybe not needed, but we could test in case no preference is set.

server/automute_test.go
```diff
@@ -0,0 +1,451 @@
...
+	directChannel, _ := p.API.GetDirectChannel(user.Id, model.NewId())
+
+	mockLinkedChannel(p, channel)
+
+	t.Run("initial conditions", func(t *testing.T) {
+		assertChannelNotAutomuted(t, p, channel.Id, user.Id)
+		assertChannelNotAutomuted(t, p, directChannel.Id, user.Id)
+	})
+
+	t.Run("should do nothing when false is passed and automuting has never been enabled", func(t *testing.T) {
```
larkox, line 275: Wonder if this set of tests could be refactor to do it in a table testing way (i.e. define the test cases, and then run them in a for loop).

larkox (approved):
A few comments, but no changes needed really. Just double check a few things.

server/automute.go
hmhealey: That's a good point about returning early. I think I forgot to do that because I was super focused on performance which is why `skipConnectedCheck` and `skipPrimaryPlatform` exist. I expected this method to be called whenever anything happened that might trigger automuting, but since I added the `automute_enabled` preference, most code can just check that instead which means performance is less of an issue here.

I'm going to return early when `!connected` like you suggested, but I might also remove the `skipX` parameters to simplify this logic as well. It'll make performance slightly worse overall, but I think that would be overoptimization overall

server/automute.go
hmhealey: There's some potential raciness here as well if the user joins/leaves a channel at the same time this logic is running. Using a transaction or a more purpose-built SQL query could help that, but it'd definitely need some sort of locking either on the app level or DB level.

It does seem like something good to be aware of, but like you said, I'm not sure how possible it would be to prevent this even with something really complicated. At least this is just about muting and isn't super critical feature, so I think it's okay even if it's not 100% correct in weird cases 😅

server/automute_test.go
hmhealey: Yeah, same. Unless we run the tests with an actual server or restructure our code in some way that we use more pure functions and less side effects, we're stuck mocking something.

Thinking about this more since last week though, I do prefer this style of mocking where you make a fake server once and then can use it all over instead of faking return values for individual API methods like we've had to do elsewhere

server/automute_test.go
hmhealey: Maybe I'm not treating this as enough of a black box, but I figured I'd handle that case when testing `isUsersPrimaryPlatformTeams` and then I could just assume that it works here. The other alternative would be to  rewrite `shouldEnableAutomute` so that it's called like 
```go
shouldEnableAutomute(p.isUserConnected(userID), p.isUsersPrimaryPlatformTeams(userID))
```
but then testing this function would be just testing that `return !connected || !teamsPrimary` returned true which is as pointless as it sounds.

Maybe that would be a better way to rewrite this, but I'm not going to rewrite it now

server/automute_test.go
hmhealey: Table testing wouldn't work in this case since it's actually relying on the state from the previous test for each step of this. The reason it uses `t.Run` is to group the assertions together so that it's more clear what each set of assertions does. I could instead write it like
```go
	p := newAutomuteTestPlugin(t)

	user := &model.User{Id: model.NewId()}

	channel, _ := p.API.CreateChannel(&model.Channel{Id: model.NewId(), Type: model.ChannelTypeOpen})
	p.API.AddUserToChannel(channel.Id, user.Id, "")
	directChannel, _ := p.API.GetDirectChannel(user.Id, model.NewId())

	mockLinkedChannel(p, channel)

	// Confirm initial conditions
		assertChannelNotAutomuted(t, p, channel.Id, user.Id)
		assertChannelNotAutomuted(t, p, directChannel.Id, user.Id)
	})

	// Confirm that this should do nothing when false is passed and automuting has never been enabled
	result, err = p.setAutomuteEnabledForUser(user.Id, false)

	assert.Equal(t, false, result)
	assert.NoError(t, err)

	assertChannelNotAutomuted(t, p, channel.Id, user.Id)
	assertChannelNotAutomuted(t, p, directChannel.Id, user.Id)
```
but I like brackets and indentation to help differentiate those sets.

Plus it makes the output look more interesting 😛

server/automute.go
lieut-data: I'd propose we log the errors, but keep working through it. If it did ever fail, is there anything that would trigger this to "catch up"?

server/automute.go
```diff
@@ -0,0 +1,182 @@
...
+
+// enableAutomute mutes all of the user's linked channels and DMs and sets a preference to track that automute is
+// enabled. It assumes that the caller has checked that the user is both connected and has their primary platform
+// set to MS Teams.
+func (p *Plugin) enableAutomute(userID string) (bool, error) {
+	return p.setAutomuteEnabledForUser(userID, true)
+}
+
+// disableAutomute mutes all of the user's linked channels and DMs and sets a preference to track that automute is
+// disabled. It assumes that the user is either not connected to MS Teams or has their primary platform set to MM.
```
lieut-data, line 27: The first part of the comment here sounds like it might be inaccurate? I'm guessing it **unmutes**, but does it also separately track that automute was disabled? Not sure why we'd need to remember that?

server/automute.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	var membersToMute []*model.ChannelMemberIdentifier
+
+	channels, appErr := p.API.GetChannelsForTeamForUser("", userID, true)
+	if appErr != nil {
+		return false, errors.Wrap(appErr, fmt.Sprintf("Unable to get channels for user %s to automute them", userID))
+	}
+
+	for _, channel := range channels {
```
lieut-data, line 45: I'm assuming this is simply unpaginated? Or does the server have some implicit pagination in play that might artificially limit the records returned?

server/automute.go
```diff
@@ -0,0 +1,182 @@
...
+
+	return true, nil
+}
+
+func (p *Plugin) setChannelMembersAutomuted(members []*model.ChannelMemberIdentifier, automuteEnabled bool) error {
+	notifyProps := getNotifyPropsForAutomute(automuteEnabled)
+
+	i := 0
+	perPage := 200
+	for i < len(members) {
```
lieut-data, line 71: Should we inline this?
```go
for i := 0; i < len(members); i += perPage {
    // ...
}
```

server/automute.go
```diff
@@ -0,0 +1,182 @@
...
+
+// func (p *Plugin) isUsersPrimaryPlatformTeams(userID string) bool {
+// 	pref, appErr := p.API.GetPreferenceForUser(userID, PreferenceCategoryPlugin, PreferenceNamePlatform)
+// 	if appErr != nil {
+// 		// GetPreferenceForUser returns an error when a preference is unset, so we default to MM being primary platform
+// 		return false
+// 	}
+
+// 	return pref.Value == PreferenceValuePlatformMSTeams
+// }
```
lieut-data, line 140: Is this used by a later PR?

server/automute_preferences.go
```diff
@@ -0,0 +1,59 @@
+package main
+
+import (
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/plugin"
+)
+
+func (p *Plugin) PreferencesHaveChanged(c *plugin.Context, preferences []model.Preference) {
+	p.updateAutomutingOnPreferencesChanged(c, preferences)
+}
```
lieut-data, line 10: Interesting, it looks like this can be called both for a single user's and multiple users' preferences?

server/automute_preferences.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, userID := range userIDsToEnable {
+		if connected, err := p.isUserConnected(userID); err != nil {
+			p.API.LogError(
+				"Unable to potentially enable automute for user",
+				"UserID", userID,
+				"Error", err.Error(),
+			)
+		} else if !connected {
+			continue
+		}
```
lieut-data, line 24: I'm wondering if we should remove this check: users who haven't connected will still get "spammed" if the counterparty in the DM is connected, and it might be desirable to effect this setting manually. Thoughts, @johndavidlugtu?

server/automute_preferences.go
```diff
@@ -0,0 +1,59 @@
...
+			)
+		}
+	}
+
+	for _, userID := range userIDsToDisable {
+		_, err := p.disableAutomute(userID)
+		if err != nil {
+			p.API.LogError(
+				"Unable to unmute channels for a user who set their primary platform to Mattermost",
+				"UserID", userID,
```
lieut-data, line 40:
```suggestion
				"user_id", userID,
```

server/automute_preferences.go
```diff
@@ -0,0 +1,59 @@
...
+				"Error", err.Error(),
```
lieut-data, line 41:
```suggestion
				"error", err.Error(),
```

server/automute.go
```diff
@@ -0,0 +1,182 @@
...
+	if appErr != nil {
+		return false, errors.Wrap(appErr, fmt.Sprintf("Unable to get channels for user %s to automute them", userID))
+	}
+
+	for _, channel := range channels {
+		if linked, err := p.canAutomuteChannel(channel); err != nil {
+			return false, err
+		} else if !linked {
+			continue
+		}
```
lieut-data, line 50: If a channel is unlinked, and the user is disabling automuting, might we inadvertently skip some channels? When setting to `false`, maybe we should do this for every channel?

server/automute.go
```diff
@@ -0,0 +1,182 @@
...
+	if automuteEnabled {
+		return map[string]string{
+			model.MarkUnreadNotifyProp: model.ChannelMarkUnreadMention,
+			NotifyPropAutomuted:        "true",
+		}
+	}
+
+	return map[string]string{
+		model.MarkUnreadNotifyProp: model.ChannelMarkUnreadAll,
+		NotifyPropAutomuted:        "false",
```
lieut-data, line 100: It's unfortunate that this persists "indefinitely" -- is it possible to return to the original state without this prop?

lieut-data (changes requested):
Thanks, @hmhealey! A few questions and comments inline. I like the simplicity of your testing strategy: hopefully we can consolidate on non-mocked tests in the future once our tooling is ready to use :)

server/automute.go
hmhealey: This does two things:
1. It unmutes all the users' connected channels
2. Sets the `automute_enabled` preference to `false`

`automute_enabled` will always equal "user is connected && primary platform is Teams", so it's kind of redundant, but it being set basically gives me confidence that all of the users' channels are correctly automuted and will stay correctly automuted down the line. It simplifies the followup logic by letting us just check `automute_enabled` when, for example, muting a channel that gets linked

server/automute.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (p *Plugin) setAutomuteEnabledForUser(userID string, automuteEnabled bool) (bool, error) {
+	if channelsAutomuted := p.getAutomuteIsEnabledForUser(userID); channelsAutomuted == automuteEnabled {
+		// We've already automuted all the users' channels, so there's nothing to do
+		return false, nil
+	}
+
+	var membersToMute []*model.ChannelMemberIdentifier
+
+	channels, appErr := p.API.GetChannelsForTeamForUser("", userID, true)
```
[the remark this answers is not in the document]
hmhealey: It returns all of those. I had to confirm that, but I added a test in https://github.com/mattermost/mattermost/pull/26057 to ensure it stays that way

server/automute.go
hmhealey: As far as I can tell, yeah, it's unpaged. At the very least, there's no paging mechanism I could use to handle if this was paged

server/automute.go
hmhealey: For some reason, that looks weird to me, but I can't think of a good reason not to do that. Maybe it's because it's rare we do a regular for loop in Go without it involving a `range`.

I'll change this regardless

server/automute.go
hmhealey: No, I don't think we can because of how this is a patch operation. Unless we were to add some special logic to the plugin API or the DB code, I don't think we could pass an undefined value here to make the DB remove this prop. If we find a way to do that though, I'd like to, but since we're reliant on server 9.5 at this point, we'd have to leave it as-is for now

server/automute.go
hmhealey: Yeah, it will be

server/automute.go
```diff
@@ -0,0 +1,182 @@
...
+// // canAutomuteChannelID returns true if the channel with the given ID is either explicitly linked to a channel in
+// // MS Teams or if it's a DM/GM channel that is implicitly linked to MS Teams.
+// func (p *Plugin) canAutomuteChannelID(channelID string) (bool, error) {
+// 	channel, appErr := p.API.GetChannel(channelID)
+// 	if appErr != nil {
+// 		return false, errors.Wrap(appErr, fmt.Sprintf("Unable to get channel %s to check if it's a DM/GM channel", channelID))
+// 	}
+
+// 	return p.canAutomuteChannel(channel)
+// }
```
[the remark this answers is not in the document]
hmhealey: Unlike the above, it doesn't end up getting used, but I've removed it in that later PR

server/automute_preferences.go
hmhealey: It's only called with a single user's preferences at the moment, but unfortunately, I wrote the signature in a way that it could be called with multiple users' preferences on the server side, so I decided to continue writing it this way. I thought that it might be possible to call that endpoint with multiple users' preferences at some point for the case where the customer wanted to set this for multiple users at once, but it turns out that wasn't the case.

server/automute_preferences.go
hmhealey: I think that would probably work as expected then since an unconnected MM-only user will still get notifications for their DMs. The only way they'd get extra notifications is if they're on Teams and MM without being connected which I can't see us supporting

server/automute_preferences.go
hmhealey: I'll go through and do a pass on these since this style will come up in all 3 of my PRs, but I saw examples of both naming conventions (and I swear there was one rogue `userId`), so it would be good to get the rest of those consistent at some point

lieut-data: See https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/480, but this should have mostly fixed everything (done after your branch was started, of course!)

server/automute.go
lieut-data: If the channel was previously linked and muted, then unlinked, I guess we're relying on the fact that we unmute at the the point of unlinking?

server/automute.go
lieut-data: Sounds good! Re: the first part of the comment, I wonder if this is meant to read:
```suggestion
// disableAutomute unmutes all of the user's linked channels and DMs and sets a preference to track that automute is
// disabled. It assumes that the user is either not connected to MS Teams or has their primary platform set to MM.
```

server/automute.go
hmhealey: The only catch up logic right now would be for the user to toggle their primary platform to disable and re-enable automuting which could run into the same error here

server/automute.go
hmhealey: Oh yeah, good catch. I was constantly mixing up words like mute/unmute/automute/un-automute and automute/automuting when writing this 😛

server/automute_preferences.go
hmhealey: Ah, that probably explains why it was so much harder for me to find an example that matched mine when I searched recently

server/automute.go
hmhealey: Yeah, that's right. As soon as one of the conditions for a channel being automuted (user is connected, user has Teams as primary platform, channel can be automuted) becomes false, the channel will no longer be muted so we don't need to or want to unmute it again when a second condition becomes false

calebroseland (approved):

server/automute_preferences.go
johndavidlugtu: Referencing the other PR where I left a comment to keep one thread: https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/479#issuecomment-1947350958

lieut-data:
Let's QA the whole feature as one.

PR #478, opened 2024-01-29
Fixing issue on reconnect required notification

jespino:
This fixes the problem of having not valid sessions for msteams and still not reciving the message for reconnect.

Now the user gets properly notified and the session is removed from the database.

### DIFF
--- a/server/msteams/layer_generators/disconnection_layer.go.tmpl
+++ b/server/msteams/layer_generators/disconnection_layer.go.tmpl
@@ -31,7 +31,7 @@ func (c *{{$.Name}}) {{$index}}({{$element.Params | joinParamsWithType}}) {{$ele
 	{{genResultsVars $element.Results false }} := c.Client.{{$index}}({{$element.Params | joinParams}})
 	if err != nil {
 		var graphErr *msteams.GraphAPIError
-		if errors.As(err, &graphErr) && graphErr.StatusCode == http.StatusUnauthorized {
+		if msteams.IsOAuthError(err) || (errors.As(err, &graphErr) && graphErr.StatusCode == http.StatusUnauthorized) {
 			c.onDisconnect(c.userID)
 		}
 	}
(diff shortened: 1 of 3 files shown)


wiggin77:
With this pattern repeated so many time would it not make sense to put that in a method so that logic be changed easily?
```
if err != nil {
		var graphErr *msteams.GraphAPIError
		if errors.As(err, &graphErr) && graphErr.StatusCode == http.StatusUnauthorized {
		if msteams.IsOAuthError(err) || (errors.As(err, &graphErr) && graphErr.StatusCode == http.StatusUnauthorized) {
			c.onDisconnect(c.userID)
		}
	}
```

wiggin77 (approved):
LGTM :+1:

server/msteams/layer_generators/disconnection_layer.go.tmpl
```diff
@@ -31,7 +31,7 @@ func (c *{{$.Name}}) {{$index}}({{$element.Params | joinParamsWithType}}) {{$ele
 	{{genResultsVars $element.Results false }} := c.Client.{{$index}}({{$element.Params | joinParams}})
 	if err != nil {
 		var graphErr *msteams.GraphAPIError
-		if errors.As(err, &graphErr) && graphErr.StatusCode == http.StatusUnauthorized {
+		if msteams.IsOAuthError(err) || (errors.As(err, &graphErr) && graphErr.StatusCode == http.StatusUnauthorized) {
```
lieut-data, line 34: Does this mean the current batch of errors when the user's session is invalidated /aren't/ returning `http.StatusUnauthorized` but either no status code or some other code instead? I'm wondering if that means our relying on status unauthorized is unnecessary, or worse: somehow incorrect.

Not blocking, but I'll push a small commit to add logging of the `StatusCode` to this PR so we can get visibility here.

lieut-data (approved):

lieut-data:
@lindalumitchell, thinking we might include this in a v1.6.1. Would it be ok to merge and test on our test server with an "official release"?

lindalumitchell (approved):
Yes, that sounds good to me, thanks. 👍

PR #479, opened 2024-01-29
MM-56501 Enable/disable automuting when a user connects/disconnects

hmhealey:
#### Summary
Continuing from https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/477, since automuting is enabled when a user is both connected and has their primary platform set to MS Teams, we need to also potentially enable or disable automuting when a user connects or disconnects from MS Teams.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56501

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -471,6 +471,8 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	a.p.updateAutomutingOnUserConnect(mmUserID)
+
 	bundlePath, err := a.p.API.GetBundlePath()
 	if err != nil {
 		a.p.API.LogWarn("Failed to get bundle path.", "error", err.Error())
--- a/server/api.go
+++ b/server/api.go
@@ -471,7 +471,7 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	a.p.updateAutomutingOnUserConnect(mmUserID)
+	_, _ = a.p.updateAutomutingOnUserConnect(mmUserID)
 
 	bundlePath, err := a.p.API.GetBundlePath()
 	if err != nil {
--- a/server/automute.go
+++ b/server/automute.go
@@ -126,15 +126,15 @@ func (p *Plugin) setAutomuteIsEnabledForUser(userID string, channelsAutomuted bo
 	return nil
 }
 
-// func (p *Plugin) isUsersPrimaryPlatformTeams(userID string) bool {
-// 	pref, appErr := p.API.GetPreferenceForUser(userID, PreferenceCategoryPlugin, PreferenceNamePlatform)
-// 	if appErr != nil {
-// 		// GetPreferenceForUser returns an error when a preference is unset, so we default to MM being primary platform
-// 		return false
-// 	}
+func (p *Plugin) isUsersPrimaryPlatformTeams(userID string) bool {
+	pref, appErr := p.API.GetPreferenceForUser(userID, PreferenceCategoryPlugin, PreferenceNamePlatform)
+	if appErr != nil && errors.Is(appErr.Unwrap(), sql.ErrNoRows) {
+		// GetPreferenceForUser returns an error when a preference is unset, so we default to MM being primary platform
+		return false
+	}
 
-// 	return pref.Value == PreferenceValuePlatformMSTeams
-// }
+	return pref.Value == PreferenceValuePlatformMSTeams
+}
 
 func (p *Plugin) isUserConnected(userID string) (bool, error) {
 	token, err := p.store.GetTokenForMattermostUser(userID)
--- a/server/automute.go
+++ b/server/automute.go
@@ -128,7 +128,7 @@ func (p *Plugin) setAutomuteIsEnabledForUser(userID string, channelsAutomuted bo
 
 func (p *Plugin) isUsersPrimaryPlatformTeams(userID string) bool {
 	pref, appErr := p.API.GetPreferenceForUser(userID, PreferenceCategoryPlugin, PreferenceNamePlatform)
-	if appErr != nil && errors.Is(appErr.Unwrap(), sql.ErrNoRows) {
+	if appErr != nil {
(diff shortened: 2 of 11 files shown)


lieut-data:
@johndavidlugtu, this relates to the comment on the previous PR, but I'm wondering if we want to restrict this functionality only for connected users. Maybe this setting should work for all users regardless of connection status?

lieut-data (changes requested):
Blocking only for feedback on enduser experience.

hmhealey:
If we were to make that change, then in other words, a non-connected user would be able to set their primary platform and have all their linked channels auto-muted.

There'd be a fair bit of this logic that would/could be rewritten then, but it would simplify some stuff since the multiple conditions for enabling automuting did make this rather complicated to reason about and make work

server/api.go
```diff
@@ -472,6 +472,8 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
 
 	_, _ = w.Write([]byte(fmt.Sprintf("<html><body><h1>%s</h1><p>You can close this window.</p></body></html>", connectionMessage)))
 
+	_, _ = a.p.updateAutomutingOnUserConnect(mmUserID)
```
lieut-data, line 475: Can we log `LogWarn` any error that's returned here to avoid burying it?

server/automute.go
```diff
@@ -126,15 +126,15 @@ func (p *Plugin) setAutomuteIsEnabledForUser(userID string, channelsAutomuted bo
...
-// func (p *Plugin) isUsersPrimaryPlatformTeams(userID string) bool {
-// 	pref, appErr := p.API.GetPreferenceForUser(userID, PreferenceCategoryPlugin, PreferenceNamePlatform)
-// 	if appErr != nil {
-// 		// GetPreferenceForUser returns an error when a preference is unset, so we default to MM being primary platform
-// 		return false
-// 	}
+func (p *Plugin) isUsersPrimaryPlatformTeams(userID string) bool {
+	pref, appErr := p.API.GetPreferenceForUser(userID, PreferenceCategoryPlugin, PreferenceNamePlatform)
+	if appErr != nil {
+		// GetPreferenceForUser returns an error when a preference is unset, so we default to MM being primary platform
```
lieut-data, line 132: Can we distinguish an error from being unset from a different kind of error?

server/automute_user_connect.go
```diff
@@ -0,0 +1,55 @@
...
+	if !p.isUsersPrimaryPlatformTeams(userID) {
+		return false, nil
+	}
+
+	automuteEnabled, err := p.enableAutomute(userID)
+	if err != nil {
+		p.API.LogError(
+			"Unable to enable automuting for user on connect",
+			"UserID", userID,
+			"Error", err,
```
lieut-data, line 13:
```suggestion
			"user_id", userID,
			"error", err,
```

server/automute_user_connect.go
```diff
@@ -0,0 +1,55 @@
+package main
+
+func (p *Plugin) updateAutomutingOnUserConnect(userID string) (bool, error) {
+	if !p.isUsersPrimaryPlatformTeams(userID) {
+		return false, nil
+	}
+
+	automuteEnabled, err := p.enableAutomute(userID)
+	if err != nil {
+		p.API.LogError(
```
lieut-data, line 10: We're about to start raising alerts to SRE for errors. Is this something that we think is human actionable? Or might this fail for other reasons that justifying just a `LogWarn`?

Alternatively, since we already return the error, can we skip the log altogether?

server/automute_user_connect.go
```diff
@@ -0,0 +1,55 @@
...
+		)
+	} else {
+		var message string
+		if automuteEnabled {
+			message = "Enabled automuting for user on connect"
+		} else {
+			message = "Not enabling automute for user on connect"
+		}
+
+		p.API.LogDebug(
```
lieut-data, line 23: Propose we log at `info` -- this is low volume, and helpful context for an event where debug logs might not be enabled.
```suggestion
		p.API.LogInfo(
```

server/automute_user_connect.go
```diff
@@ -0,0 +1,55 @@
...
+			message,
+			"UserID", userID,
```
lieut-data, line 25:
```suggestion
			"user_id", userID,
```

server/automute_user_connect.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	automuteEnabled, err := p.enableAutomute(userID)
+	if err != nil {
+		p.API.LogError(
+			"Unable to enable automuting for user on connect",
+			"UserID", userID,
+			"Error", err,
+		)
+	} else {
```
lieut-data, line 15: nit: maybe we can `return false, err` here and unindent the rest of the method?

lieut-data (approved):
This looks great, @hmhealey! Just a handful of comments, but nothing that requires a re-review from me.

server/api.go
hmhealey: This code is actually a bit misleading since we do log any errors inside `updateAutomutingOnUserConnect`. I wanted the method to log its own errors to keep the automute logic more self-contained, but I still wanted the method to return errors because that makes testing easier

server/automute.go
hmhealey: Yeah, looks like we can actually. I thought that wouldn't work because `GetPreferenceForUser` returns an error which wraps `sql.ErrNoRows`, but we can actually test for that with `errors.Is(err, sql.ErrNoRows)`

server/automute_user_connect.go
hmhealey: Yeah, I can change them to LogWarn since they're not really human actionable as you said.

And returning vs logging is the opposite of my previous comment. I can change that if you think it's a good idea if you think it's better to log these in `api.go` to be a bit more idiomatic

johndavidlugtu:
> Maybe this setting should work for all users regardless of connection status?

Good point @lieut-data . Correct in that receiving "unwanted" DM/GM notifications from Mattermost is dependent on the sender's connection status and not the recipient.

+1 on opening it up to non-connected users

cc: @hmhealey

calebroseland (approved):

PR #481, opened 2024-02-01
remove avatar handling

lieut-data:
#### Summary
Discussed with @jespino [here](https://hub.mattermost.com/private-core/pl/3z7aj48shiby589hh7jarza19a), but in short this PR removes support for resolving and caching MS Teams avatars. The primary win here is reducing scope and surface area.

For context, in situations where we can't create or don't have a corresponding synthetic user, we fallback to sending posts from the MS Teams bot in Mattermost. After this PR, we'll stop using the `override_username` and `override_icon_url` to simulate the message as coming from someone else: it will instead appear as coming from the MS Teams user. (We may revisit this latter functionality going forward).

This change unblocks https://mattermost.atlassian.net/browse/MM-55978.

#### Ticket Link
Relates-to: https://mattermost.atlassian.net/browse/MM-55978


server/handlers/converters_test.go
```diff
@@ -50,12 +49,10 @@ func TestMsgToPost(t *testing.T) {
 			},
 			setupPlugin: func(p *mocksPlugin.PluginIface, mockAPI *plugintest.API, client *mocksClient.Client, mockmetrics *mocksMetrics.Metrics) {
 				p.On("GetBotUserID").Return(testutils.GetSenderID())
-				p.On("GetURL").Return("https://example.com/")
 				p.On("GetClientForApp").Return(client)
 				p.On("GetMetrics").Return(mockmetrics).Maybe()
 			},
 			setupAPI: func(api *plugintest.API) {
-				api.On("LogDebug", "Unable to get user avatar", "Error", mock.Anything).Once()
```
wiggin77, line 58: Seems like this can be simplified further.  We range through a slice of test cases that only has one test, and the setAPI call does nothing.  Should we just remove that?

wiggin77 (approved):
LGTM.  Nothing blocking.

server/handlers/converters_test.go
lieut-data: Sigh, you're right -- this test is basically doing nothing. I've got a separate ticket to revisit all of our tests as e2e / integration variety, and this one will probably just disappear in the process.

sbishel (approved):

lieut-data:
@lindalumitchell, another proposed "check in the bug bash". Actually, this is behaviour that was "net new" to me last week, so I'm not entirely sure what checking it means 😆.

lindalumitchell (approved):
Sure, sounds good to me. 👍

PR #482, opened 2024-02-01
drop support for MySQL

lieut-data:
#### Summary
In addition to a hard-drop for MySQL (we're not testing this configuration, so let's benefit from the simplification), this PR also simplify store-related unit tests (easier to invoke a single test now) as well as speeds up test container initialization.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-55978


README.md
```diff
@@ -44,6 +44,8 @@ This plugin supports the following features:
 
 ## Installation
 
+These installation instructions assume you already have a [Mattermost instance](https://mattermost.com/download/) running [PostgreSQL](https://www.postgresql.org/):
```
wiggin77, line 47: This comment doesn't explicitly state that only Postgres is supported.  It suggests the installation instructions are for a Postgres installation but could be interpreted as just a limitation of the instructions.  

I suggest a more epxlicit "Postgres required" be included.

lieut-data: That's fair -- @crspeller, I copied this from the AI plugin. How have you been handling pushback on Postgres thus far? For both our plugins, perhaps we could tweak this language to be:

```suggestion
These installation instructions assume you already have a [Mattermost instance](https://mattermost.com/download/) running [PostgreSQL](https://www.postgresql.org/). Note that this plugin does not support MySQL.
```

crspeller: Pushback for support is documented in https://github.com/mattermost/mattermost-plugin-ai/issues/19

The tweak looks good for both plugins.

sbishel (approved):

PR #491, opened 2024-02-02
only expose metrics via ServeMetrics

lieut-data:
#### Summary
With https://mattermost.atlassian.net/browse/MM-55507 resolved and shipped, we no longer need to expose metrics on a discrete port and can exclusively rely on the `ServeMetrics` hooks. This resolves a class of metrics failures that still creates issues in some deployments.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-55508


cpoile (approved):
Looks good!

I'll take Claudio off, since he's out for another 2 weeks -- can you request someone else?

isacikgoz (commented):
Looks good but we should update the `min_server_version` requirement in the plugin manifest right?

lieut-data:
> Looks good but we should update the `min_server_version` requirement in the plugin manifest right?

We could, but as this is already a "breaking change" (dropping support) and the new metrics are a no-op if the containing server doesn't call the hook, I think we can skip the min version bump for this change, at least. We'll definitely need to bump it for some of the other incoming changes re: shared channels, though.

Thanks, @isacikgoz!

isacikgoz (approved):
Thanks for the explanation, LGTM!

PR #492, opened 2024-02-02
drop -sync, renaming to just msteams

lieut-data:
#### Summary
Note that this does **NOT** change the plugin id. It's not worth the cost to effect a plugin ID change at this time. (Too many dependencies on the KV Store and metrics scraping.)

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-55478


johndavidlugtu:
FYI on this @cwarnermm

johndavidlugtu (approved):
Thanks @lieut-data

cwarnermm:
@lieut-data - Is this PR intended to update the System Console UI text to remove ``sync`` terminology as well? Or will that update be addressed as a separate PR? CC @johndavidlugtu

crspeller (approved):

lieut-data:
@cwarnermm, this should remove the `sync` terminology everywhere, but I'll double check the system console to be sure.

@amyblais, heads up I'm punting this to v1.8.0 until https://mattermost.atlassian.net/browse/CLD-7049 is resolved.

lieut-data:
Alas, I thought @johndavidlugtu's review would "count", but seems like I need a second developer review -- @wiggin77, might I get your sign off?

wiggin77:
@lieut-data @johndavidlugtu has anyone checked whether we can just name this plugin "MS Teams" / "msteams"?  Isn't that trademark infringement?  Also see https://www.microsoft.com/en-us/legal/intellectualproperty/trademarks

server/message_hooks.go
```diff
@@ -542,7 +542,7 @@ func (p *Plugin) handlePromptForConnection(userID, channelID string) {
 	}
 
 	if time.Until(timestamp) < -time.Hour*time.Duration(promptInterval) {
-		p.sendBotEphemeralPost(userID, channelID, "Your Mattermost account is not connected to MS Teams so your activity will not be relayed to users on MS Teams. You can connect your account using the `/msteams-sync connect` slash command.")
+		p.sendBotEphemeralPost(userID, channelID, "Your Mattermost account is not connected to MS Teams so your activity will not be relayed to users on MS Teams. You can connect your account using the `/msteams connect` slash command.")
```
wiggin77, line 545: Are we standardizing on "MS Teams" versus "Microsoft Teams" or "MS-Teams" ?

wiggin77 (approved):
LGTM

server/message_hooks.go
lieut-data: Good question! I'm hoping @calebroseland and @asaadmahmood can help us with all the bot interactions as well as word smithing going forward.

PR #493, opened 2024-02-02
require enterprise license

lieut-data:
#### Summary
Require an enterprise license to start the plugin. This PR intentionally includes a test license for up to 100 users that only works in a test service environment.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-56618


server/plugin.go
```diff
@@ -410,6 +410,12 @@ func (p *Plugin) OnActivate() error {
 
 	p.apiClient = pluginapi.NewClient(p.API, p.Driver)
 
+	config := p.apiClient.Configuration.GetConfig()
+	license := p.apiClient.System.GetLicense()
+	if !pluginapi.IsE20LicensedOrDevelopment(config, license) {
```
marianunez, line 415: Out of scope of this PR but we may want to remane this function check in the plugin api as it was misleading that it was just checking E20 (after looking it checks for enterprise too)

server/ce2e/license_test.go
```diff
@@ -0,0 +1,34 @@
...
+		for _, plugin := range plugins.Active {
+			require.NotEqual(t, "com.mattermost.msteams-sync", plugin.Manifest.Id)
+		}
+
+		for _, plugin := range plugins.Inactive {
+			if plugin.Manifest.Id == "com.mattermost.msteams-sync" {
+				return
+			}
+		}
+		require.Fail(t, "failed to find plugin status at all")
```
marianunez, line 32: I'm not familiar with this test containers approach but seems that we are checking here that the plugin is inactive as it failed to start without the license. This works but wondering if there is another way, maybe at the plugin unit test level to check that it failed for the actual error of the license and not something else.

server/plugin_test.go
```diff
@@ -78,6 +78,7 @@ func newTestPlugin(t *testing.T) *Plugin {
 	plugin.API.(*plugintest.API).On("KVSetWithOptions", "mutex_subscriptions_cluster_mutex", []byte(nil), model.PluginKVSetOptions{ExpireInSeconds: 0}).Return(true, nil).Maybe()
 	plugin.API.(*plugintest.API).On("KVSetWithOptions", "mutex_mmi_bot_ensure", []byte{0x1}, model.PluginKVSetOptions{Atomic: true, ExpireInSeconds: 15}).Return(true, nil).Times(1)
 	plugin.API.(*plugintest.API).On("KVSetWithOptions", "mutex_mmi_bot_ensure", []byte(nil), model.PluginKVSetOptions{ExpireInSeconds: 0}).Return(true, nil).Times(1)
+	plugin.API.(*plugintest.API).On("GetLicense").Return(&model.License{SkuShortName: "E20"}).Maybe()
```
marianunez, line 81: nit: could we change this to `enterprise` ? E20 has reached EOL so we should be using the latest sku

marianunez (approved):

nickmisasi (approved):
Nothing to add on top of Maria's comments - LGTM. 

Will we need some sort of PluginHook from Mattermost itself so that the action of adding or changing a license performs this check again? I'm unfamiliar with whether the plugins are automatically restarted when a license changes. Thinking of a scenario where someone in Cloud upgrades from Professional -> Enterprise, but the plugin is not running because on load it was a Pro license.

lieut-data:
> Will we need some sort of PluginHook from Mattermost itself so that the action of adding or changing a license performs this check again? I'm unfamiliar with whether the plugins are automatically restarted when a license changes. Thinking of a scenario where someone in Cloud upgrades from Professional -> Enterprise, but the plugin is not running because on load it was a Pro license.

Good call out @nickmisasi! We don't have a hook like that today, though I'd wager `OnConfigurationChange` might get called "incidentally" when this happens. Either way, in this particular case, since the plugin refuses to start without a valid license, the customer would be expected to upgrade first and then enable the plugin.

PR #494, opened 2024-02-02
Remove enforce connected flow

lieut-data:
#### Summary
This feature was intended to force users to connect to MS Teams, but as this is no longer part of the rollout plan, let’s remove it to reduce surface area and reduce risk. It also has a non-trivial impact on plugin updates when all connected users smash the endpoint.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-56756


enahum (approved):
Lgtm, I have a high level idea of what this means, but still unfamiliar with the code

calebroseland (approved):
> Lgtm, I have a high level idea of what this means, but still unfamiliar with the code

^ Likewise

lieut-data:
@lindalumitchell, this should be as simple as verifying the setting is gone and nothing else bad occurs. Did you want to test this pre-emptively, or should we wait for the bug bash?

lindalumitchell (approved):
We can check it in the bug bash. 👍

PR #495, opened 2024-02-08
MM-56498/MM-56499/MM-56500 Automuting when users join channels or channels change

hmhealey:
#### Summary
The previous PRs (https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/477 and https://github.com/mattermost/mattermost-plugin-msteams-sync/pull/479) set up automuting and initially mute the required channels, and this PR make it so that channels are automatically muted or unmuted as needed while automute is enabled for a user. The cases that are handled are:
1. When a user with automuting joins a channel, it's muted for them
2. When a DM is created, if either member has automuting enabled, the channel will be muted
    1. This code supports GMs as well, but as I've mentioned elsewhere, this doesn't work due to https://mattermost.atlassian.net/browse/MM-56776 on the server side.
    2. This code doesn't need to handle public/private channels because I've assumed that they're never created in a state where they're both already linked and already have members inside of them.
3. When a channel is linked, it is muted for all members with automuting enabled. The opposite applies when a channel is unlinked.

Most of this follows the patterns set up in the previous PRs, but I wanted to mention:
1. For plugin hooks that aren't used elsewhere, I put them in the automuting code. If you want me to move them to somewhere central like `server/hooks.go`, I can do that, but no such file seems to exist currently.
2. I modified the `AutomuteAPIMock` to also call the needed plugin hooks to behave more like the server. From what I can tell, the server triggers those hooks when plugin API methods are called, so this matches server behaviour.
3. I copied over some changes that have been made to the `.golangci.yml` because I was getting frustrated at the linting rules forcing weird things we don't do on the server like preventing us from using if-elseif blocks (it wants you to use a switch block?) or from writing `page += 1`. Let me know if you want me to revert those, but they were really bothering me.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56498
https://mattermost.atlassian.net/browse/MM-56499
https://mattermost.atlassian.net/browse/MM-56500


server/automute_channel.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	if canAutomute, err := p.isChannelLinked(channelID); err != nil {
+		p.API.LogError(
+			"Unable to check if channel is linked to update automuting when a user has joined the channel",
+			"UserID", userID,
+			"ChannelID", channelID,
+			"Error", err.Error(),
+		)
+		return false, errors.Wrap(err, "Unable to update automuting when a user has joined a channel")
```
lieut-data, line 28: Should we just use `canAutomuteChannel` here? Maybe redundant for DMs and GMs, but feels less "surprising".

server/automute_channel.go
```diff
@@ -0,0 +1,94 @@
...
+	if automuteEnabled := p.getAutomuteIsEnabledForUser(userID); !automuteEnabled {
+		return false, nil
+	}
+
+	if canAutomute, err := p.isChannelLinked(channelID); err != nil {
+		p.API.LogError(
+			"Unable to check if channel is linked to update automuting when a user has joined the channel",
+			"UserID", userID,
+			"ChannelID", channelID,
+			"Error", err.Error(),
```
lieut-data, line 26: A recent change, but let's use snake case here:
```suggestion
			"user_id", userID,
			"channel_id", channelID,
			"error", err.Error(),
```

server/automute_channel.go
```diff
@@ -0,0 +1,94 @@
...
+func (p *Plugin) updateAutomutingOnChannelLinked(channelID string) error {
+	// This simply mutes the channel for all users with automuting enabled, regardless of their settings before. It
+	// doesn't pay attention to if the user manually muted the channel beforehand.
+	return p.updateAutomutingForChannelMembers(channelID, true)
+}
+
+func (p *Plugin) updateAutomutingOnChannelUnlinked(channelID string) error {
+	// This simply unmutes the channel for all users with automuting enabled, regardless of their settings before. It
+	// doesn't pay attention to if the user manually muted the channel beforehand to keep it muted.
+	return p.updateAutomutingForChannelMembers(channelID, false)
```
lieut-data, line 60: This is unfortunate, but not blocking. It would be really neat if we could somehow distinguish the two types of muting, maybe by remembering the timestamp for manual vs. auto events.

server/automute_channel.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (p *Plugin) updateAutomutingForChannelMembers(channelID string, enableAutomute bool) error {
+	var membersToUpdate []*model.ChannelMemberIdentifier
+
+	page := 0
+	perPage := 200
+	for {
```
lieut-data, line 68: nit: inline?
```
for page := 0; ; page += 1 {
    // ...
}
```

server/automute_channel.go
```diff
@@ -0,0 +1,94 @@
...
+		members, appErr := p.API.GetChannelMembers(channelID, page, 200)
```
lieut-data, line 69:
```suggestion
		members, appErr := p.API.GetChannelMembers(channelID, page, perPage)
```

server/automute_channel.go
```diff
@@ -0,0 +1,94 @@
...
+
+		for _, member := range members {
+			if p.getAutomuteIsEnabledForUser(member.UserId) {
+				membersToUpdate = append(membersToUpdate, &model.ChannelMemberIdentifier{ChannelId: channelID, UserId: member.UserId})
+			}
+		}
+
+		if len(members) < perPage {
+			break
+		}
```
lieut-data, line 82: nit: if `len(members) == perPage`, but we're at the end of the list, we'll go through the loop once more anyway and check this a second time. There's nothing wrong with that, it just make this terminating condition slightly unclear. I suppose we could always iterate until `len(members) == 0`. -1/5, just thinking out loud.

lieut-data (changes requested):
Looks great! Just a handful of thoughts.

server/automute_channel.go
hmhealey: I could do that, but it would require an additional DB call to load the channel from the DB since we'd need to check if get the channel to check if it's a DM/GM channel. Avoiding that DB call is likely premature optimization though

server/automute_channel.go
hmhealey: I could save the previous state in the `NotifyProps` and then restore it when the channel is no longer automuted, but I honestly don't think that will come up often enough to bother doing that. I'd assume channels won't go back and forth between connected often and that, if they are unlinked, it'll be long enough that the user wouldn't remember if they had it previously muted

server/automute_channel.go
hmhealey: In retrospect, it really feels like we should've done this paging on the server-side and let plugins just get an unpaginated list. I can't think of any cases where plugin code wouldn't want to get all the objects at once, so it seems like we're going to see this pattern a lot

server/automute_channel.go
hmhealey: Yeah, this is the same flaw in our paging API that we run into in frontend code as well where the caller has to guess if there's more results. It's a bit less expensive here than with an HTTP request, but it would be tremendously better if the APIs were explicit if there were more results or not. That's easy to implement because we just query the DB for `perPage + 1` results on the server side and use that extra result to know whether to set `hasMore` to true, but it would require changes to every paging API we have

server/automute_channel.go
lieut-data: I do think it's valuable to be able to request less than the whole payload, otherwise we'd run into potential bottlenecks with large requests. We can always abstract away the pagination in the pluginapi at some point :)

hmhealey: I'd rather leave this one as-is since it can't be written using the middle condition parameter without jumping through some hoops with the scope of `members`

larkox (approved):
LGTM

lieut-data (approved):
Thanks, @hmhealey! Re: detecting error type across the plugin boundary, I'd suggest testing this "for real" vs. mocking as there's something that makes me think the way we serialize errors may not preserve the `wrapped` error.

hmhealey:
That's a very good point. I just added a unit test on the server for that, and you're right that the serialization prevents that strategy from working. Unfortunately, that error doesn't have a unique message or status code we could use here, so I think we need to go back to treating any error as "preference not found"

lieut-data:
Merging this so we can test the whole feature as one.

PR #497, opened 2024-02-10
stop managing global chats subscription if disabled

lieut-data:
#### Summary
Honour the existing "Sync direct and group messages" setting, skipping the attempt to register a global chats subscription accordingly.

#### Ticket Link
None.

### DIFF
--- a/server/monitor/subscriptions.go
+++ b/server/monitor/subscriptions.go
@@ -110,13 +110,33 @@ func (m *Monitor) checkChannelsSubscriptions(msteamsSubscriptionsMap map[string]
 // 	}
 // }
 
-func (m *Monitor) checkGlobalSubscriptions(msteamsSubscriptionsMap map[string]*clientmodels.Subscription, allChatsSubscription *clientmodels.Subscription) {
+func (m *Monitor) checkGlobalChatsSubscription(msteamsSubscriptionsMap map[string]*clientmodels.Subscription, allChatsSubscription *clientmodels.Subscription) {
 	subscriptions, err := m.store.ListGlobalSubscriptions()
 	if err != nil {
 		m.api.LogWarn("Unable to get the chat subscriptions from store", "error", err.Error())
 		return
 	}
 
+	// Clean up if we're not syncing direct messages.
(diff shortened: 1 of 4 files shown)


server/monitor/subscriptions.go
```diff
@@ -110,13 +110,33 @@ func (m *Monitor) checkChannelsSubscriptions(msteamsSubscriptionsMap map[string]
...
 		m.api.LogWarn("Unable to get the chat subscriptions from store", "error", err.Error())
 		return
 	}
 
+	// Clean up if we're not syncing direct messages.
+	if !m.syncDirectMessages {
+		// Delete any MS Teams subscription for the global chats, if present.
+		if allChatsSubscription != nil {
+			if err := m.client.DeleteSubscription(allChatsSubscription.ID); err != nil {
+				m.api.LogWarn("Failed to delete old global chats subscriptions", "error", err.Error())
```
enahum, line 125: Other than logging, should we do something else like attempting to remove the subscription at a later time?

server/monitor/subscriptions.go
```diff
@@ -110,13 +110,33 @@ func (m *Monitor) checkChannelsSubscriptions(msteamsSubscriptionsMap map[string]
...
+			}
+		}
+
+		// Delete any global subscriptions (assume at most one here).
+		if len(subscriptions) > 0 {
+			if err := m.store.DeleteSubscription(subscriptions[0].SubscriptionID); err != nil {
+				m.api.LogWarn("Unable to delete the old all chats subscription", "error", err.Error())
```
enahum, line 132: Same here

enahum (approved):

server/monitor/subscriptions.go
jespino: If this fails, in theory should try it again in the next monitor iteration, It shouldn't be a big deal.

server/monitor/subscriptions.go
jespino: Same here, it should try again in the next monitor iteration (in 1 minute or so)

jespino (approved):
LGTM

PR #500, opened 2024-02-14
MM-56776 Update mock to fix ChannelHasBeenCreated not being called for GMs

hmhealey:
#### Summary
I made `AutomuteAPIMock` match the server behaviour in MM-56776, so once https://github.com/mattermost/mattermost/pull/26226 goes in, the mock can also be "fixed". At the moment, we're planning for that fix to ship in 9.6, so fixing the corresponding bug on the MS Teams plugin (newly created GMs won't have automuting applied) will simply require upgrading to MM 9.6 once that's out

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56776

### DIFF
--- a/server/automute_channel_test.go
+++ b/server/automute_channel_test.go
@@ -175,8 +175,6 @@ func TestUpdateAutomutingOnChannelCreated(t *testing.T) {
 	})
 
 	t.Run("when a GM is created, should mute it for users with automuting enabled", func(t *testing.T) {
-		t.Skip("not working due to MM-56776")
-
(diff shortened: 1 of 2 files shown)


lieut-data (approved):
@hmhealey, is there any harm in merging this now, with the "real fix" coming with v9.6?

hmhealey:
@lieut-data No, there's nothing blocking this. I wanted these to behave "realistically", but the bug fix for this issue is entirely in the MM server

enahum (approved):

PR #502, opened 2024-02-14
avoid burying errors during test setup

lieut-data:
#### Summary
I ran into a recent build error, but the details were buried. Surface them more explicitly if this happens again.

#### Ticket Link
None,.


JulienTant (approved):
Death to the `_`! :D Thanks @lieut-data

calebroseland (approved):

PR #504, opened 2024-02-15
MM-56778 - handle self chats

sbishel:
#### Summary
In Teams, single user (self) chats are group chats. Mattermost handles these as Direct Messages. Therefore, if only a single member is present, we need to convert the Teams Message Type from "GM" to "DM". Mattermost expects DMs to have 2 users as well, but will accept a duplicate UserID, so we need to add the UserID to the array as well.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-56778


lieut-data (approved):
Neat!

calebroseland (approved):

PR #506, opened 2024-02-15
Add plugin's remoteID to all posts

wiggin77:
#### Summary
Add the plugin's remoteID to all posts.  This is needed as Shared Channels uses that field to ensure it does not pync posts back to the remote from which it came.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56826

### DIFF
--- a/server/handlers/converters.go
+++ b/server/handlers/converters.go
@@ -5,10 +5,11 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/net/html"
+
 	"github.com/mattermost/mattermost-plugin-msteams-sync/server/markdown"
 	"github.com/mattermost/mattermost-plugin-msteams-sync/server/msteams/clientmodels"
 	"github.com/mattermost/mattermost/server/public/model"
-	"golang.org/x/net/html"
 )
 
 const hostedContentsStr = "hostedContents"
@@ -22,6 +23,7 @@ func (ah *ActivityHandler) msgToPost(channelID, senderID string, msg *clientmode
 	text = markdown.ConvertToMD(text)
 	props := make(map[string]interface{})
 	rootID := ""
+	remoteID := ah.plugin.GetRemoteID()
 
 	if msg.ReplyToID != "" {
 		rootInfo, _ := ah.plugin.GetStore().GetPostInfoByMSTeamsID(msg.ChatID+msg.ChannelID, msg.ReplyToID)
@@ -40,7 +42,15 @@ func (ah *ActivityHandler) msgToPost(channelID, senderID string, msg *clientmode
 		text = "## " + msg.Subject + "\n" + text
 	}
 
-	post := &model.Post{UserId: senderID, ChannelId: channelID, Message: text, Props: props, RootId: rootID, CreateAt: msg.CreateAt.UnixNano() / int64(time.Millisecond)}
+	post := &model.Post{
+		UserId:    senderID,
+		ChannelId: channelID,
+		Message:   text,
+		Props:     props,
+		RootId:    rootID,
+		RemoteId:  &remoteID,
+		CreateAt:  msg.CreateAt.UnixNano() / int64(time.Millisecond),
+	}
 	if !isUpdatedActivity {
 		post.FileIds = attachments
 	}
--- a/server/handlers/converters_test.go
+++ b/server/handlers/converters_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+
 	mocksPlugin "github.com/mattermost/mattermost-plugin-msteams-sync/server/handlers/mocks"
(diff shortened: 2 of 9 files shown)


wiggin77:
I will be testing before merge, but the code is ready for review.

lieut-data (approved):
Brilliant.

jespino (approved):
LGTM

lieut-data:
@lindalumitchell  / @lindy65, would you be able to dry-run this PR and confirm no regression in basic syncing functionality for DMs and GMs? 

@wiggin77 & @jespino, also worth noting that when we eventually start using shared channels, it will mean users can't edit or delete posts from before this transition. That's totally fine, but probably something to keep in mind in case we run into any surprises.

wiggin77:
> users can't edit or delete posts from before this transition

It's not quite as bad as that.  You can edit/delete posts, but you need to be the owner (per usual) and you have to edit/delete from the platform you created the post from.

lieut-data:
> > users can't edit or delete posts from before this transition
> 
> It's not quite as bad as that. You can edit/delete posts, but you need to be the owner (per usual) and you have to edit/delete from the platform you created the post from.

Sorry, yes, I misspoke -- edits or deletes from before the eventual transition will stop syncing between platforms.

wiggin77:
> edits or deletes from before the eventual transition will stop syncing between platforms.

@lieut-data I may have a blind-spot here.  I am expecting edits and deletions to continue to sync after this code is deployed, provided the edit/delete is not on the originating platform.  This PR reduces the amount of redundant sync'ing by not sync'ing back to the origin, but legacy posts will not have a remoteID and thus should sync.  We may need to add code to this PR to ensure the remoteID gets added on edits/deletes.  Currently it is only added on new posts.

lieut-data:
> @lieut-data I may have a blind-spot here. I am expecting edits and deletions to continue to sync after this code is deployed, provided the edit/delete is done on the originating platform. This PR reduces the amount of redundant sync'ing by not sync'ing back to the origin, but legacy posts will not have a remoteID and thus should sync. We may need to add code to this PR to ensure the remoteID gets added on edits/deletes. Currently it is only added on new posts.

You're absolutely right, @wiggin77 -- I had the impact backwards. 

For context, I was thinking through the following sequence of events:
```
// hooks being used, before this PR
post1 (remoteID == nil) created
// hooks being used, after this PR
post2 (remoteID != nil)
// shared channels used, after other PRs
post3 (remoteID != nil)
```

Once we cut over to shared channels, edits to `post2` and `post3` will sync normally, but I mistakenly thought we'd stop "seeing" changes to `post1` because of this PR. But as you correctly point out, `post1` won't get filtered out at all, so there's no meaningful change to the flow of events to the plugin.

Thank you for clarifying this!

wiggin77:
@lieut-data you raised a good point in that sync won't happen if the user edits/deletes the post from a different plaform that the post was created on.

User creates a post in Teams.  It gets a remoteID assigned when the post is sync'd to MM.  If the user modifies the post in MM, which I assume is possible since they are the considered the same user in MS Teams Plugin (?), but the remoteID check won't allow that to be sync'd back.  

In Shared Channels, there is no concept of the same user existing on both sides so really a post cannot be modified from a different platform.

lieut-data:
> @lieut-data you raised a good point in that sync won't happen if the user edits/deletes the post from a different plaform that the post was created on.
> 
> User creates a post in Teams. It gets a remoteID assigned when the post is sync'd to MM. If the user modifies the post in MM, which I assume is possible since they are the considered the same user in MS Teams Plugin (?), but the remoteID check won't allow that to be sync'd back.
> 
> In Shared Channels, there is no concept of the same user existing on both sides so really a post cannot be modified from a different platform.

Oh dear -- you're right again. Does this mean we abandon setting `RemoteId` at all, and decide in the plugin whether or not we need to sync the change?

wiggin77:
> Does this mean we abandon setting `RemoteId` at all, and decide in the plugin whether or not we need to sync the change?

Possibly, but we need to be careful that we don't create an endless loop.  Another option is to not allow editing of posts in MM that originate on Teams.   Yet another option is to continue to set remoteID, and filter, but make an exception for posts modified by the author (creatorID == the ID of the user making the change).

lindalumitchell:
Thanks @lieut-data and @wiggin77 -- are there likely to be more changes to this (related to the above discussion), or should we go ahead with testing?

lieut-data:
@lindalumitchell & @lindy65, let's pause for now while we sort out next steps -- sorry for the thrashing!

lieut-data:
> > and decide in the plugin whether or not we need to sync the change?
> 
> @lieut-data how will the plugin make that determination? We need to be absolutely sure we don't create an endless loop.

@wiggin77, my understanding is the plugin already has a mechanism to do this today, since we would otherwise get the same endless loop from hooks too. Is this something you can help make define more explicitly as part of https://mattermost.atlassian.net/browse/MM-56829?

wiggin77:
Closing based on team decision to filter posts in the plugin.

PR #507, opened 2024-02-16
fix Continue button during oAuth flow

lieut-data:
#### Summary
`widow.close` generally won't work, but even if it doesn't, fall back to showing the "You can now close this window" prompt. I've also added `target="_blank"` for the learn more links to avoid losing this page state.

My edit threw in some random whitespace fixes as well which seemed harmless.

#### Ticket Link
None.

### DIFF
--- a/assets/info-page/index.html
+++ b/assets/info-page/index.html
@@ -420,7 +420,7 @@
                   <img src="static/list-icon-info.png" alt="list-icon" class="list-icon">You receive notifications in Mattermost.
                 </li>
                 <li class="list-item">
-                  <img src="static/list-icon-warn.svg" alt="list-icon" class="list-icon">You must disable notifications in Microsoft Teams to avoid duplicate notifications. <a href="https://mattermost.com/pl/ms-teams-plugin-end-user-learn-more" class="list-link">Learn more</a>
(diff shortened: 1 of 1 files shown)


assets/info-page/index.html
```diff
@@ -50,9 +50,9 @@
       function choosePrimaryPlatform() {
         const serverURL = "{{.ServerURL}}";
         const apiEndpoint = "{{.APIEndPoint}}";
-        const queryParam = "{{.QueryParamPrimaryPlatform}}";        
+        const queryParam = "{{.QueryParamPrimaryPlatform}}";
         fetch(`${serverURL}${apiEndpoint}?${queryParam}=${primaryPlatform}`)
-        .then(() => window.close());
+        .then(closeInfoPage());
```
lieut-data, line 55: See above, but `closeInfoPage` also tries `window.close` (even though it likely won't work).

calebroseland (approved):
LGTM. Would be nice if we could update the ephemeral message after successful connection. ~Is there a story for that somewhere already?~ added https://mattermost.atlassian.net/browse/MM-56913

hmhealey (approved):

PR #510, opened 2024-02-21
fix closeInfoPage to call on promise resolution

lieut-data:
#### Summary
I regressed the behaviour, which closed the window before the promise was resolved. In my case, since the window never actually closed, the fetch had a chance to succeed. On browsers that honoured `window.close`, the `fetch` was interrupted and the settings never saved.

#### Ticket Link
None yet.


calebroseland (approved):

larkox (approved):
:facepalm:

PR #511, opened 2024-02-26
update google.com links

lieut-data:
#### Summary
Fix links hard-coded to google.com to use the permalinks from the info-page instead.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-56959


sbishel (approved):

lieut-data:
I've deployed and tested these changes locally, and will follow up once v1.8.1 is cut to do the same.

PR #512, opened 2024-02-27
Do not sync users when there is no need

enahum:
#### Summary
The sync users job was updating the activity of guests users every time even when there was nothing to be updated.

At the same time, if we have multiple users with the same `DisplayName` on MSTeams, we would trigger an `UpdateUser`  incrementing by one the suffix of the username even if the Mattermost username did not need to be updated, triggering unnecessary updates.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56942

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -694,7 +694,7 @@ func (p *Plugin) syncUsers() {
 		if msUser.Type == msteamsUserTypeGuest {
 			// Check if syncing of MS Teams guest users is disabled.
 			if !syncGuestUsers {
-				if isUserPresent && isRemoteUser(mmUser) {
+				if isUserPresent && isRemoteUser(mmUser) && mmUser.DeleteAt == 0 {
 					// Deactivate the Mattermost user corresponding to the MS Teams guest user.
 					p.API.LogInfo("Deactivating the guest user account", "user_id", mmUser.Id, "teams_user_id", msUser.ID)
 					if err := p.API.UpdateUserActive(mmUser.Id, false); err != nil {
@@ -707,6 +707,7 @@ func (p *Plugin) syncUsers() {
 		}
 
 		username := "msteams_" + slug.Make(msUser.DisplayName)
+		mmUsername := mmUser.Username // temporarily store the original username
 		if !isUserPresent {
 			userUUID := uuid.Parse(msUser.ID)
 			encoding := base32.NewEncoding("ybndrfg8ejkmcpqxot1uwisza345h769").WithPadding(base32.NoPadding)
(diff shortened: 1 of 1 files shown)


server/plugin.go
```diff
@@ -707,6 +707,7 @@ func (p *Plugin) syncUsers() {
 		}
 
 		username := "msteams_" + slug.Make(msUser.DisplayName)
+		mmUsername := mmUser.Username // temporarily store the original username
```
lieut-data, line 710: `mmUser` might be `nil` here.

server/plugin.go
```diff
@@ -774,9 +775,15 @@ func (p *Plugin) syncUsers() {
...
-						continue
+						mmUser.Username = mmUsername               // assign the original username
+						username += "-" + fmt.Sprint(userSuffixID) // add the sufix before checking again
+						if mmUser.Username != username {
+							mmUser.Username = username
+							userSuffixID++
+							continue
+						}
+
+						break
```
lieut-data, line 786: I wonder if there's more to address here:
* if we ever do have to rely on `userSuffixID` to generate the username, it sounds like the check on line 771 will always be false, and we'll always call `UpdateUser` at least once, perhaps more before we get "back" to the suffixed username. 
* `username` looks like it keeps accumulating instead of incrementing? `username`, `username-${userSuffixID}`, `username-${userSuffixID}-${userSuffixID + 1}`, `...`

Wondering if it makes sense to make the suffix deterministic, such as embedding some part of the user id? Or maybe don't even try to "fix" these users at all, log a warning, and leave them unsynced?

server/plugin.go
```diff
@@ -774,9 +775,15 @@ func (p *Plugin) syncUsers() {
 				_, err := p.API.UpdateUser(mmUser)
```
lieut-data, line 775: Would it be simpler if we split apart the checks for `username != mmUser.Username` and `msUser.DisplayName != mmUser.FirstName` so we could only generate usernames in one case?

server/plugin.go
```diff
@@ -774,9 +775,15 @@ func (p *Plugin) syncUsers() {
...
 				if err != nil {
 					if err.Id == "app.user.save.username_exists.app_error" {
-						mmUser.Username += "-" + fmt.Sprint(userSuffixID)
-						userSuffixID++
-						continue
+						mmUser.Username = mmUsername               // assign the original username
+						username += "-" + fmt.Sprint(userSuffixID) // add the sufix before checking again
+						if mmUser.Username != username {
```
lieut-data, line 780: I'm having trouble understanding this block, since it seems like it's guaranteed to always be true given the assignments prior?

lieut-data (changes requested):
Thanks, @enahum! The `mmUser.DeleteAt == 0` sounds like it's responsible for the bulk of the spurious extra `UpdateUser` calls. Given the complexity with the other cases discussed below, wondering if it makes sense to split that apart to a discrete PR that we could merge more readily?

enahum:
@lieut-data I kept both things in this PR as both scenarios will trigger the update WS event as mentioned in the ticket.

I came with another alternative that is checking that the user does not already contain the prefix `msteams_` meaning that it has not been updated before, the only issue I see in this case is that the username will not really update for users that have already been sync'ed once, but I guess that should be ok.

lieut-data (approved):
Thanks, @enahum! This is very much clearer to understand :)

PR #513, opened 2024-02-27
Using mock-server instead of the built-in mock for msteam

jespino:
This PR replaces the built-in mock server, that required a specific compilations and mocked the whole msteams client, with one that is external ([mock-server](https://www.mock-server.com/)). This allows to do a full test of 100% of our code, without testing the MSTeams side.


enahum:
@jespino I was looking into https://github.com/microsoft/dev-proxy for mocking.. it has many other things not just mocking responses. The idea was to use that for the load tests.. in your opinion what do you think is best
1. Keep this and use it for the load tests as well
2. Switch to dev proxy
3. Use both, this and load tests with dev proxy

Let me know

server/msteams/client.go
```diff
@@ -200,25 +200,40 @@ func NewManualClient(tenantID, clientID string, logService *pluginapi.LogService
...
+
+	adapter, err := msgraphsdk.NewGraphRequestAdapterWithParseNodeFactoryAndSerializationWriterFactoryAndHttpClient(authProvider, nil, nil, httpClient)
 	if err != nil {
-		fmt.Printf("Error creating client: %v\n", err)
+		fmt.Printf("Error creating credentials: %v\n", err)
 		return nil
 	}
-	c.client = client
+
+	clientMutex.Lock()
```
enahum, line 234: Just so that I can understand, why the mutex?

server/msteams/client_mock.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func getAuthClient() *http.Client {
+	proxyAddress := "http://mockserver:1080"
+	proxyUrl, _ := url.Parse(proxyAddress)
+	// Setup proxy for the token credential from azidentity
+	authClient := &http.Client{
+		Transport: &http.Transport{
+			Proxy:           http.ProxyURL(proxyUrl),
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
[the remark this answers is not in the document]
enahum: Do we want to pay attention to this?

enahum (approved):
A couple of comments but this looks great

server/msteams/client_mock.go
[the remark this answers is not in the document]
jespino: Actually, we don't, that is totally expected for testing :)

server/msteams/client.go
jespino: I was replicating the pre-existing behavior, but I don't really know if that is needed in all the cases. Let me review it, probably is only needed in the case of `Connect` call, that can happen outside the creating process.

jespino:
> @jespino I was looking into https://github.com/microsoft/dev-proxy for mocking.. it has many other things not just mocking responses. The idea was to use that for the load tests.. in your opinion what do you think is best
> 
>     1. Keep this and use it for the load tests as well
> 
>     2. Switch to dev proxy
> 
>     3. Use both, this and load tests with dev proxy
> 
> 
> Let me know

I don't know much about the dev-proxy, but one thing I like about the mock-server is that it is fully open-source and generalistic, so we can make it behave however we want. We can modify the mocks on the fly with the API, reset them, make assertions... I think it is very powerful. Maybe dev-proxy is equally powerful, but I don't know it. Also, mock-server is already an official testcontainers module. Also, depending on the dev-proxy capabilities, we can try to use dev-proxy instead of mock-server, I can create a testcontainers module for dev-proxy without much effort (I guess), and if that works as a proxy, probably it would be as easy to use as the one that we are using now.

server/plugin.go
```diff
@@ -314,6 +308,15 @@ func (p *Plugin) start(isRestart bool) {
...
 
+	// Unregister and re-register slash command to reflect any configuration changes.
+	if err = p.API.UnregisterCommand("", "msteams"); err != nil {
+		p.API.LogWarn("Failed to unregister command", "error", err)
+	}
+
+	if err = p.API.RegisterCommand(p.createCommand(p.getConfiguration().SyncLinkedChannels)); err != nil {
+		p.API.LogError("Failed to register command", "error", err)
+	}
+
```
lieut-data, line 319: I'm assuming relocating this relates to the testing changes, but do we understand what was preventing this from "running normally" before?

server/testutils/containere2e/containere2e.go
```diff
@@ -2,29 +2,38 @@ package containere2e
...
+
 var buildPluginOnce sync.Once
 
 func buildPlugin(t *testing.T) {
 	cmd := exec.Command("make", "-C", "../../", "dist")
 	cmd.Env = os.Environ()
 	cmd.Env = append(cmd.Env, "DEFAULT_GOOS=linux")
 	cmd.Env = append(cmd.Env, "DEFAULT_GOARCH=amd64")
-	cmd.Env = append(cmd.Env, "GO_BUILD_TAGS=clientMock")
+	cmd.Env = append(cmd.Env, "MM_SERVICESETTINGS_ENABLEDEVELOPER=true")
```
lieut-data, line 35: Note that this implies different semantics at runtime, i.e. if we had license checks they would "change" with this setting. Is this strictly necessary?

server/testutils/mmcontainer/testscontainer.go
```diff
@@ -25,7 +24,7 @@ const (
 	defaultPassword        = "admin"
 	defaultTeamName        = "test"
 	defaultTeamDisplayName = "Test"
-	defaultMattermostImage = "mattermost/mattermost-enterprise-edition"
+	defaultMattermostImage = "mattermost/mattermost-enterprise-edition:release-9.6"
```
lieut-data, line 27: Should we avoid hard-coding this, lest we forget and keep testing on what will be an old version?

server/msteams/client.go
```diff
@@ -200,25 +200,40 @@ func NewManualClient(tenantID, clientID string, logService *pluginapi.LogService
 		client:     nil,
 	}
 
+	// Create an HTTP client with the middleware
```
lieut-data, line 203: @jespino, can you help me understand the full set of changes to `NewManualClient`? I'm guessing we're lifting some things around here so we can have access to internals that we need for the mocking, but I'm wondering about the risk that our copy/paste here deviates from what the `msgraphsdk` does in the future.

Would it make sense to isolate these changes to creating a `NewManualTestingClient` or some such so that production is guaranteed to be unaffected?

lieut-data (approved):
Thanks, @jespino! A few questions below, but nothing strictly blocking.

server/plugin.go
jespino: I think the problem was a race condition (The check certificates was taking time, so the plugin wasn't registering it), I can go back to the other way because now I should be replying to the list of certificate really quickly because is mocked, and if not, it is at least not forwarded to Microsoft api, so it shouldn't be a problem now.

server/testutils/containere2e/containere2e.go
jespino: No, I was trying to speed up the compilation time, but I can remove it. For sure.

server/testutils/mmcontainer/testscontainer.go
jespino: That is going to happen, at least locally, anyway, because if you don't specify the version is going to use the one that you have already in your cache. Also, having it explicitly ensure that the plugin works in previous versions, if it fails, you detect that you need to bump the version of mattermost in the requirements of the plugin.json.

server/msteams/client.go
jespino: Actually, the NewManualClient was already there and the idea was to be used for creating a client manually (with user intervention) that allows to execute the E2E tests that hits the real MSTeams. Probably makes now sense to expose the mockserver to this, and probably we can move it into the tests. Let me review that, but think we doesn't need that changes here. Anyway, I'll do them.

server/msteams/client.go
jespino: removed where is not needed.

jespino:
@lieut-data I addressed all your concerns, I think, PTAL.

lieut-data (approved):
Thanks, @jespino!

PR #515, opened 2024-02-28
avoid assets error

lieut-data:
#### Summary
When this plugin is deployed, the `assets/` folder ends up being scanned within the monorepo as part of `make check-style` and spits out the following:

```
plugins/com.mattermost.msteams-sync/assets/assets.go:1:1: License not found or wrong
make: *** [vet] Error 1
```

That's because the `assets/` folder both gets copied with the plugin **and** is used to compile assets directly into the binary. Just add a license to fix this for now.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-57016


wiggin77:
@lieut-data I'm curious why not just add a license header to the file?

lieut-data:
> @lieut-data I'm curious why not just add a license header to the file?

Hah, yeah, I guess that would work too... thanks @wiggin77 :p

wiggin77 (approved):

enahum (approved):

PR #516, opened 2024-02-28
introduce selective sync

lieut-data:
#### Summary
If enabled, change the sync strategy for chats to only sync messages that “span platforms”. That means:
* local user → local user: no sync
* remote user → remote user: no sync
* local user → remote user: sync

A user is **remote** if they are synthetic (does not login, but managed by the MS Teams plugin) or if they have chosen MS Teams as their primary platform. Remote users are expected to read chats exclusively from MS Teams. DMs and GMs for users who have chosen MS Teams as their primary platform are automatically muted.

A user is **local** if they login normally and have kept the default of Mattermost as their primary platform. A local user is expected to read chats primarily from within Mattermost. Local users may or may not have connected their Mattermost account to MS Teams.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-57017


server/selective_sync_test.go
```diff
@@ -0,0 +1,444 @@
+//go:build exclude
```
lieut-data, line 1: See https://mattermost.atlassian.net/browse/MM-57018 for the remaining work here, but these tests pass locally on my machine.

server/selective_sync_test.go
```diff
@@ -0,0 +1,444 @@
...
+		user1 := setupUser(t, p, team, false)
+		user2 := setupUser(t, p, team, false)
+
+		channel, err := p.API.CreateDirectChannel(user1.Id, user2.Id)
+		require.Nil(t, err)
+
+		chatSpansPlatforms, appErr := p.ChatSpansPlatforms(channel.Id)
+		require.Nil(t, appErr)
+		assert.False(t, chatSpansPlatforms)
+	})
```
lieut-data, line 116: This test is "real" -- no mocks, not limited to just the external plugin API. It requires changes in the server to support "reattaching plugins" (see https://mattermost.atlassian.net/browse/MM-57018), but in short this will complement our testcontainers/e2e strategy but for unit tests where mocking remained the only option.

lieut-data:
Thank you, @enahum -- your thoughts re: performance are spot on. I too am hoping the existing cache will serve us well, and hopefully we can quantify the impact better during an upcoming loadtest.

jespino (approved):
LGTM

sbishel (approved):

lieut-data:
@lindy65, proposing we test this as part of bug bash and into this week.

lindy65:
Queued for bug bash @lieut-data 👍

PR #518, opened 2024-02-28
MM-56681 Use existing file Id if already uploaded

sbishel:
#### Summary
Updated to process file attachments on an update. However, code was added to find and use the existing FileID rather than upload the file again.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-56681


lieut-data:
@sbishel, I may not be able to review this until after the SKO -- @isacikgoz, wondering if you'd be willing to take a peek at this?

server/handlers/attachments.go
```diff
@@ -220,6 +222,23 @@ func (ah *ActivityHandler) handleAttachments(channelID, userID, text string, msg
 	return newText, attachments, parentID, errorFound
 }
 
+func (ah *ActivityHandler) getFileInfo(name string, fileIDs []string, fileInfos map[string]*model.FileInfo) string {
```
isacikgoz, line 225: To confirm, are we trying to see if an attachment is already present either in existing attachements or file store? I am a bit confused with the function here. (Sorry, it's my first time looking into this code)

isacikgoz (commented):
With the tests, functionality wise looks good. I have a couple of questions to understand the overall code further.

server/handlers/attachments.go
sbishel: @isacikgoz - Yeah...that was a complicated and hard to understand attempt at limiting db calls to absolute necessity. I rewrote this logic which is -
- if there are attachments, retrieve the existing attachment's file Infos. 
- this is because, from MS Teams we receive a "filename", but we have fileID.
- if an attachment is already present, just use that FileID. Otherwise, download...store, create, etc.

I think the new logic is way more readable.

sbishel:
There is one big flaw in this PR. That is we are simply matching on file name...it is probably possible to add another file with the same name on the MS Teams side. 

However, because we are not supporting attachments for June and this PR makes things...better. We should revisit the "same filename" problem in the future.

sbishel:
@jespino - gentle ping....

jespino (approved):
LGTM. For the same filename, would be great to keep it tracked in an issue to not forget to address it later.

sbishel:
> LGTM. For the same filename, would be great to keep it tracked in an issue to not forget to address it later.

Created https://mattermost.atlassian.net/browse/MM-57358

server/handlers/attachments.go
```diff
@@ -123,6 +123,17 @@ func (ah *ActivityHandler) handleAttachments(channelID, userID, text string, msg
...
 	}
 
+	fileNames := make(map[string]string)
+	if len(msg.Attachments) > 0 {
+		for _, fID := range existingFileIDs {
+			fileInfo, _ := ah.plugin.GetAPI().GetFileInfo(fID)
+			if fileInfo == nil {
+				continue
+			}
+			fileNames[fileInfo.Name] = fID
```
isacikgoz, line 133:
```suggestion
			if fileInfo != nil {
				fileNames[fileInfo.Name] = fID
			}
```

server/handlers/attachments.go
```diff
@@ -147,11 +158,15 @@ func (ah *ActivityHandler) handleAttachments(channelID, userID, text string, msg
...
-			continue
+		fileInfoID := fileNames[a.Name]
+		if fileInfoID != "" {
+			attachments = append(attachments, fileInfoID)
 		}
+		// // We don't support retroactively adding file attachments to posts.
+		// if isUpdatedActivity {
+		// 	skippedFileAttachments = true
+		// 	continue
+		// }
```
isacikgoz, line 169: Shall we remove the commented out lines.

server/handlers/attachments_test.go
```diff
@@ -497,7 +582,11 @@ func TestHandleAttachments(t *testing.T) {
 				ChannelID:   testutils.GetChannelID(),
 			}
 
+<<<<<<< HEAD
```
isacikgoz, line 585: Seems like a conflicted file found its way here.

isacikgoz (commented):
Thanks for clarification @sbishel I think it's good to go after fixing the conflicts and the builds.

isacikgoz (approved):

PR #520, opened 2024-02-29
Move LICENSE to bottom of README

lieut-data:
#### Summary
There's no good reason the LICENSE is the most prominent paragraph in the README. Let's move it to the bottom, like we do for https://github.com/mattermost/mattermost-plugin-ai.

#### Ticket Link
None.


lieut-data:
@jasonblais & @johndavidlugtu, tagging you just for visibility. No review required :)

PR #521, opened 2024-03-01
Fix a DB connection leak on restart

jespino:
This ensures that the db connection provided by the plugin API gets closed on plugin deactivation.

Fixes: https://mattermost.atlassian.net/browse/MM-57036


server/plugin.go
```diff
@@ -396,6 +396,10 @@ func (p *Plugin) stop(isRestart bool) {
 			p.API.LogError("failed to close metrics job", "error", err)
 		}
 	}
+
+	if err := p.store.Shutdown(); err != nil {
+		p.API.LogError("failed to db connection", "error", err)
```
wiggin77, line 401:
```suggestion
		p.API.LogError("failed to close db connection", "error", err)
```

server/plugin.go
```diff
@@ -506,6 +510,9 @@ func (p *Plugin) OnActivate() error {
 		)
 		p.store = timerlayer.New(store, p.GetMetrics())
 		if err = p.store.Init(); err != nil {
+			if err2 := p.store.Shutdown(); err2 != nil {
```
wiggin77, line 513: Not blocking but a comment...  with so many exit paths and the chance to forget to shut down the DB in the future, it would be nice to shutdown the DB anytime an err is returned from this method.  Either put all this logic in a setup method called by `OnActivate`, or use a defer and err return var.

wiggin77 (approved):

lieut-data (changes requested):
Thanks, @jespino -- a few follow ups!

lieut-data:
@agnivade, could you take a look here as well?

server/store/sqlstore/store.go
```diff
@@ -50,6 +50,10 @@ func New(db *sql.DB, api plugin.API, enabledTeams func() []string, encryptionKey
 	}
 }
 
+func (s *SQLStore) Shutdown() error {
+	return s.db.Close()
+}
```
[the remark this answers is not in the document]
lieut-data: cc @agnivade to add some perspective here -- I was hoping that the server would essentially "do this for free" when we shut down.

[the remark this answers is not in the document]
agnivade: Thanks, this is a good point. On the server side, we don't do the tracking of RPC connections at a plugin level. It will require some work, but not impossible. Then we can easily connect the plugin shutdown event to effectively closing those connections.

agnivade (commented):
Agree with others, let's use defer here for simplicity and clarity.

server/plugin.go
jespino: :+1:

server/plugin.go
```diff
@@ -494,16 +500,26 @@ func (p *Plugin) OnActivate() error {
 			func() []byte { return []byte(p.configuration.EncryptionKey) },
 		)
 		p.store = timerlayer.New(store, p.GetMetrics())
+
+		defer func() {
+			if err != nil {
+				if err = p.store.Shutdown(); err != nil {
+					p.API.LogWarn("failed to close db connection", "error", err)
```
lieut-data, line 507: Should this be an error like when it fails above? We /might/ want to take action in this case and rotate the pods.

server/plugin.go
```diff
@@ -557,13 +574,14 @@ func (p *Plugin) OnActivate() error {
...
-		if err := p.store.PrefillWhitelist(); err != nil {
-			p.API.LogWarn("Error in populating the whitelist with already connected users", "error", err.Error())
+		if err2 := p.store.PrefillWhitelist(); err2 != nil {
+			p.API.LogWarn("Error in populating the whitelist with already connected users", "error", err2.Error())
 		}
 	}()
 
 	go p.start(false)
-	return nil
+	err = nil
```
lieut-data, line 583: Why do we need to assign this to `nil`?

server/plugin.go
```diff
@@ -557,13 +574,14 @@ func (p *Plugin) OnActivate() error {
...
+	return err
```
lieut-data, line 584: Isn't this equivalent to `return nil` as before?

server/plugin.go
```diff
@@ -430,11 +438,11 @@ func (p *Plugin) generatePluginSecrets() error {
 	return nil
 }
 
-func (p *Plugin) OnActivate() error {
+func (p *Plugin) OnActivate() (err error) {
```
lieut-data, line 441: nit: I don't use named return values often, but wondering if it makes sense to do something like `retErr` and /not/ restructure all the other error handling below. Thea idea would be to only use `retErr` inside `defer` callbacks and otherwise /avoid/ named return values. Thoughts?

lieut-data (approved):
Thanks, @jespino! Two additional comments, but not-blocking.

server/plugin.go
jespino: I think I going to move it to an internal `onActivate` function and do the error handling outside, I think is going to be cleaner.

lieut-data (approved):
Thanks, @jespino! This looks great :)

lieut-data:
@lindy65, propose we test this during bug bash by having @jespino disable / enable the plugin on ms teams sync test a few times and make sure it doesn't get into a bad state again :)

lindy65:
Thanks @lieut-data - noted

PR #533, opened 2024-03-05
Add initial E2E tests on msteams change events

jespino:
This PR adds a couple of tests that simulate MSTeams events and checks if the change events are handled correctly.

This is testing the happy path, and this should work as an example to add more tests for other events and other cases.


enahum (approved):

server/ce2e/msteams_events_test.go
```diff
@@ -0,0 +1,267 @@
...
+			},
+			expectPostCreation: true,
+		},
+	}
+
+	for _, tc := range ttCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.mock != nil {
+				tc.mock()
+				defer mockClient.Reset()
```
wiggin77, line 239:
```suggestion
				defer require.NoError(t, mockClient.Reset())
```

wiggin77 (approved):
LGTM :+1:

PR #534, opened 2024-03-05
Mm 56726 reconnect

sbishel:

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -66,6 +66,7 @@ func NewAPI(p *Plugin, store store.Store) *API {
 	router.HandleFunc("/autocomplete/teams", api.autocompleteTeams).Methods("GET")
 	router.HandleFunc("/autocomplete/channels", api.autocompleteChannels).Methods("GET")
 	router.HandleFunc("/connect", api.connect).Methods("GET", "OPTIONS")
+	router.HandleFunc("/reconnect", api.reconnect).Methods("GET", "OPTIONS")
 	router.HandleFunc("/oauth-redirect", api.oauthRedirectHandler).Methods("GET", "OPTIONS")
 	router.HandleFunc("/connected-users", api.getConnectedUsers).Methods(http.MethodGet)
 	router.HandleFunc("/connected-users/download", api.getConnectedUsersFile).Methods(http.MethodGet)
@@ -316,8 +317,23 @@ func (a *API) autocompleteChannels(w http.ResponseWriter, r *http.Request) {
 }
 
 func (a *API) connect(w http.ResponseWriter, r *http.Request) {
+	connectURL := a.getConnectURL(w, r)
+	if connectURL != "" {
+		data, _ := json.Marshal(map[string]string{"connectUrl": connectURL})
+		_, _ = w.Write(data)
+	}
+}
+
+func (a *API) reconnect(w http.ResponseWriter, r *http.Request) {
+	connectURL := a.getConnectURL(w, r)
+	if connectURL != "" {
+		http.Redirect(w, r, connectURL, http.StatusSeeOther)
+	}
+}
+
+func (a *API) getConnectURL(w http.ResponseWriter, r *http.Request) string {
 	if r.Method == http.MethodOptions {
-		return
+		return ""
 	}
 	userID := r.Header.Get("Mattermost-User-ID")
 
@@ -325,20 +341,17 @@ func (a *API) connect(w http.ResponseWriter, r *http.Request) {
 	if err := a.store.StoreOAuth2State(state); err != nil {
 		a.p.API.LogWarn("Error in storing the OAuth state", "error", err.Error())
 		http.Error(w, "Error in trying to connect the account, please try again.", http.StatusInternalServerError)
-		return
+		return ""
 	}
 
 	codeVerifier := model.NewId()
 	if appErr := a.p.API.KVSet("_code_verifier_"+userID, []byte(codeVerifier)); appErr != nil {
 		a.p.API.LogWarn("Error in storing the code verifier", "error", appErr.Message)
 		http.Error(w, "Error in trying to connect the account, please try again.", http.StatusInternalServerError)
-		return
+		return ""
 	}
 
-	connectURL := msteams.GetAuthURL(a.p.GetURL()+"/oauth-redirect", a.p.configuration.TenantID, a.p.configuration.ClientID, a.p.configuration.ClientSecret, state, codeVerifier)
-
-	data, _ := json.Marshal(map[string]string{"connectUrl": connectURL})
-	_, _ = w.Write(data)
+	return msteams.GetAuthURL(a.p.GetURL()+"/oauth-redirect", a.p.configuration.TenantID, a.p.configuration.ClientID, a.p.configuration.ClientSecret, state, codeVerifier)
 }
 
 func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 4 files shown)


server/api.go
```diff
@@ -314,29 +315,41 @@ func (a *API) autocompleteChannels(w http.ResponseWriter, r *http.Request) {
...
+		_, _ = w.Write(data)
+	}
+}
+
+func (a *API) reconnect(w http.ResponseWriter, r *http.Request) {
+	connectURL := a.getConnectURL(w, r)
+	if connectURL != "" {
+		http.Redirect(w, r, connectURL, http.StatusSeeOther)
+	}
+}
```
calebroseland, line 330: nit, 2/5

I wonder if `/connect?format=json` & `/connect` would be more appropriate than `/connect` & `/reconnect`?

Seeing `reconnect` in the invite URL `You're invited ... [Click here to connect your acccount](../reconnect)` may seem a bit off to some users.


On the other hand, do we even need to return the raw connect URL? Perhaps a single `/connect` that does the redirect would be enough? https://github.com/mattermost/mattermost-plugin-msteams/blob/21b2e41752dbd663981ea7015c8dee3d46778d63/webapp/src/index.tsx#L24-L26 
^ could turn into:
```tsx
onClick: () => window.open(Client.getConnectRedirectUrl(), '_blank');
```

server/message_hooks.go
```diff
@@ -554,9 +554,10 @@ func (p *Plugin) handlePromptForConnection(userID, channelID string) {
 	}
 
 	if time.Until(timestamp) < -time.Hour*time.Duration(promptInterval) {
-		p.sendBotEphemeralPost(userID, channelID, "Your Mattermost account is not connected to MS Teams so your activity will not be relayed to users on MS Teams. You can connect your account using the `/msteams connect` slash command.")
-
-		if err = p.store.StoreDMAndGMChannelPromptTime(channelID, userID, time.Now()); err != nil {
+		connectURL := p.GetURL() + "/reconnect"
+		message := fmt.Sprintf("[Click here to re-connect your account](%s)", connectURL)
```
hmhealey, line 558: I had to double check, but reconnect doesn't need a hyphen
```suggestion
		message := fmt.Sprintf("[Click here to reconnect your account](%s).", connectURL)
```

sbishel:
Based on Caleb's comments, which I didn't completely understand, :) I started looking at the difference between `connect` and `reconnect`. Essentially, it came down to essentially where the redirect occurs.  

- One does a server-side redirect and the other does a client-side "redirect" via window.open. 

I have reworked everything so the initial `connect` command gets redirected from the server. Which therefore means, we don't need `reconnect`, since they now both work the same. Let me know if anyone knows why this won't work or are issues with this solution.

@lieut-data - adding you as reviewer for your input.

webapp/src/index.tsx
```diff
@@ -21,9 +21,7 @@ function getSettings(serverRoute: string, disabled: boolean) {
                 title: 'Connect your Microsoft Teams Account',
                 text: 'Connect your Mattermost and Microsoft Teams accounts to get the ability to link and synchronise channel-based collaboration with Microsoft Teams.',
                 buttonText: 'Connect account',
-                onClick: () => Client.connect().then((result) => {
-                    window.open(result?.connectUrl, '_blank');
-                }),
+                onClick: () => Client.connect(),
```
calebroseland, line 24: I think this still needs to be in a new window

```suggestion
                onClick: () => window.open(`${Client.url}/connect`),
```

(...or with a modified client method)
```suggestion
            onClick: () => window.open(Client.getConnectRoute()),
```

calebroseland (changes requested):
Sorry Scott, I was coming from the perspective of [connection invites](https://mattermost.atlassian.net/browse/MM-56858) where the redirect will be used in  an invite message like: `You're invited to be an early adopter for the MS Teams connected experience. [Click here to connect your account](../plugins/com.mattermost.msteams-sync/connect)`


That's where I thought `../plugins/com.mattermost.msteams-sync/reconnect` would be a bit odd.

sbishel:
I spoke with Caleb yesterday and am working for those fixes...should have it ready this afternoon.

sbishel:
@calebroseland @hmhealey - Should be ready to review.

server/command.go
```diff
@@ -477,19 +476,7 @@ func (p *Plugin) executeConnectCommand(args *model.CommandArgs) (*model.CommandR
...
-	}
-
-	codeVerifier := model.NewId()
-	if appErr := p.API.KVSet("_code_verifier_"+args.UserId, []byte(codeVerifier)); appErr != nil {
-		p.API.LogWarn("Error in storing the code verifier", "error", appErr.Error())
-		return p.cmdError(args.UserId, args.ChannelId, genericErrorMessage)
-	}
-
-	connectURL := msteams.GetAuthURL(p.GetURL()+"/oauth-redirect", p.configuration.TenantID, p.configuration.ClientID, p.configuration.ClientSecret, state, codeVerifier)
+	connectURL := p.GetURL() + "/connect"
```
esarafianou, line 479: Why is the PKCE part of the OAuth2 flow removed?

esarafianou (changes requested):

server/command.go
sbishel: It's not removed, just centralized. It happens in the connect api, [here](https://github.com/mattermost/mattermost-plugin-msteams/blob/a853fe81c8eef9e1ffb29143f8a17d30fef049ff/server/api.go#L316)

Rather than creating the redirect link in multiple places, we now create a `/connect` link which itself creates and handles the redirect.

webapp/src/index.tsx
```diff
@@ -62,7 +60,7 @@ export default class Plugin {
         registry.registerAdminConsoleCustomSetting('ConnectedUsersReportDownload', ListConnectedUsers);
 
         // let settingsEnabled = (state as any)[`plugins-${manifest.id}`]?.connectedStateSlice?.connected || false; //TODO use connected selector from https://github.com/mattermost/mattermost-plugin-msteams/pull/438
-        let settingsEnabled = true;
+        let settingsEnabled = false;
```
calebroseland, line 63: Not 100% sure whether this needs to be toggled or not

calebroseland (approved):

webapp/src/index.tsx
sbishel: Thanks. Nice catch...that was from testing...

server/plugin.go
```diff
@@ -182,10 +182,12 @@ func (p *Plugin) OnDisconnectedTokenHandler(userID string) {
 		p.API.LogWarn("Unable to get direct channel for send message to user", "user_id", userID, "error", appErr.Error())
 		return
 	}
+
+	connectURL := p.GetURL() + "/connect"
 	_, appErr = p.API.CreatePost(&model.Post{
 		UserId:    p.GetBotUserID(),
 		ChannelId: channel.Id,
-		Message:   "Your connection to Microsoft Teams has been lost. Please reconnect using `/msteams connect` slash command in any Mattermost channel.",
+		Message:   "Your connection to Microsoft Teams has been lost. " + fmt.Sprintf("[Click here to  reconnect your account](%s).", connectURL),
```
hmhealey, line 190:
```suggestion
		Message:   "Your connection to Microsoft Teams has been lost. " + fmt.Sprintf("[Click here to reconnect your account](%s).", connectURL),
```

hmhealey (approved):
I can't speak for anything about the semantics of OAuth and connect vs reconnect, but the rest looks good to me minus one last typo

esarafianou (approved):

sbishel:
In order to get in for v1.10 we will defer QA and bug bash tomorrow.

PR #536, opened 2024-03-06
Fix replica conn leak

jespino:
This uses the plugin-api to close the db connection the plugin is done with it. It should fix the connection leak for master db and for replica db.


agnivade (approved):
Probably we should spawn a spinwick and do some testing to confirm that the fix works.

jespino:
We can build it a deploy it in our test server after a restart.

lieut-data (approved):

PR #537, opened 2024-03-06
Adding e2e tests for selective syncing from mattermost to msteams

jespino:
This PR adds a table test for selective syncing from mattermost to msteams. The table test includes the following test cases:

 - from not connected to not connected
 - from not connected to mmprimary
 - from not connected to msteamsprimary
 - from not connected to synthetic
 - from mmprimary to not connected
 - from mmprimary to mmprimary
 - from mmprimary to msteamsprimary
 - from mmprimary to synthetic
 - from msteamsprimary to not connected
 - from msteamsprimary to mmprimary
 - from msteamsprimary to msteamsprimary
 - from msteamsprimary to synthetic

I also change some logic about what a teams primary user is, using the preference around teams primary instead of the "automute" one, that in theory is a "side effect", not the primary definition.


server/selective_sync.go
```diff
@@ -31,7 +31,7 @@ func (p *Plugin) ChatMembersSpanPlatforms(members model.ChannelMembers) (bool, *
 		if isRemoteUser(user) {
 			// Synthetic users are always remote.
 			atLeastOneRemoteUser = true
-		} else if p.getAutomuteIsEnabledForUser(user.Id) {
+		} else if p.getPrimaryPlatform(user.Id) == PreferenceValuePlatformMSTeams {
```
jespino, line 34: I change this to use the real preferences about primary platform, instead of auto-mute preference

server/preferences.go
```diff
@@ -8,3 +8,14 @@ const (
 	PreferenceValuePlatformMM      = "mattermost"
 	PreferenceValuePlatformMSTeams = "msteams"
 )
+
+// getPrimaryPlatform returns the user's primary platform preference.
+func (p *Plugin) getPrimaryPlatform(userID string) string {
+	pref, appErr := p.API.GetPreferenceForUser(userID, PreferenceCategoryPlugin, PreferenceNamePlatform)
+	if appErr != nil {
+		// Default to false if no preference is found
```
lieut-data, line 16: nit:
```suggestion
		// Default to Mattermost if not found or an error occurred
```

lieut-data (approved):
Very cool! Thank you, @jespino :)

server/ce2e/plugin_test.go
```diff
@@ -314,3 +331,269 @@ func TestMessageHasBeenPostedNewDirectMessageE2E(t *testing.T) {
...
+	defer conn.Close()
+
+	ttCases := []struct {
+		name                         string
+		fromUser                     *model.User
+		toUser                       *model.User
+		expectedWithSelectiveSync    bool
+		expectedWithoutSelectiveSync bool
+	}{
+		{
```
jespino, line 424: @asaadmahmood it would be great of you go through this table testing and you see something that we should test here, and if there is anything there that doesn't fit with your expectations.

Also, take into account that this is only "mattermost to msteams", the other way needs to go in another set of tests.

wiggin77 (approved):
LGTM :+1:

PR #538, opened 2024-03-07
Avoid crash if configuration is nil

enahum:
#### Summary
If the configuration is invalid, the value of `configuration` is nil, this is causing a crash when calling `generateSecrets`


jespino (approved):
LGTM

enahum:
@jespino any idea what might be failing here?

server/configuration.go
```diff
@@ -165,6 +165,9 @@ func (p *Plugin) OnConfigurationChange() error {
 	}
 
 	if err := p.validateConfiguration(configuration); err != nil {
+		if p.configuration == nil {
+			p.setConfiguration(configuration)
+		}
```
lieut-data, line 170: The `getConfiguration` handles this -- within the mutex -- by returning `&configuration{}`. Is there a reason we need to deal with `configuration == nil` again?

lieut-data (changes requested):
Thanks, @enahum! One question below, given `getConfiguration` handler logic already in place. Generally speaking, we don't want to use `configuration` directly because of the race conditions, but also because we would have to wait for the first `OnConfigurationChange` for it to become safe.

enahum:
@lieut-data not really sure.. I was looking at this with @jespino cause of the plugin crashing when deployed to the load test environment and we thought this might be the solution for the crash

lieut-data:
@enahum, it does seem like we have some incorrect direct usages of `p.configuration` in the plugin: I'd propose fixing those to either use `getConfiguration` or `setConfiguration` as appropriate, especially if the panic was related.

enahum:
> @enahum, it does seem like we have some incorrect direct usages of `p.configuration` in the plugin: I'd propose fixing those to either use `getConfiguration` or `setConfiguration` as appropriate, especially if the panic was related.

Ok I'll make that change

server/plugin.go
```diff
@@ -410,26 +410,27 @@ func (p *Plugin) restart() {
 
 func (p *Plugin) generatePluginSecrets() error {
 	needSaveConfig := false
-	if p.configuration.WebhookSecret == "" {
+	cfg := p.getConfiguration()
```
lieut-data, line 413: Let's `Clone` here, as the `configuration` struct should be considered immutable:
```suggestion
	cfg := p.getConfiguration().Clone()
```

server/plugin.go
```diff
@@ -410,26 +410,27 @@ func (p *Plugin) restart() {
...
 			return err
 		}
 
-		p.configuration.EncryptionKey = secret
+		cfg.EncryptionKey = secret
 		needSaveConfig = true
 	}
 	if needSaveConfig {
-		configMap, err := p.configuration.ToMap()
+		configMap, err := cfg.ToMap()
```
lieut-data, line 433: With the `Clone` above, we also have to call `p.setConfiguration(cfg)` if we intend to rely on the configuration change immediately. But then that causes a race in that we might be clobbering another update to the config that happened synchronously.

I wonder if we could replicate what Playbooks had to do here? See https://github.com/mattermost/mattermost-plugin-playbooks/blob/master/server/config/service.go, but in short configuration updates are done under a lock with a callback to effectively handle the direct patching and server update.

lieut-data (changes requested):
Thanks, @enahum! Sharing some context below as it jogged my memory of issues we encountered with Playbooks.

Also, are there other direct usages of `p.configuration` that we need to update?

server/plugin.go
jespino, line 413: My concern with the `getConfiguration` here is, if you have a invalid configuration (for example, manually change in the config.json), you will get an empty configuration here (because the OnConfigurationChange, check that is valid), this would wipe out all the other settings, valid or not.

lieut-data: @jespino, can you elaborate on how this makes things worse? If the configuration is indeed invalid and we throw it away `OnConfigurationChange`, would it be correct to treat it as empty as a result?

server/plugin.go
enahum: this seems to be already the case. where it differs is that we call this function `generatePluginSecrets` when activating the plugin, and if needed it will call the API SavePluginConfig method and rely on the `onConfigurationChange` hook.

So what I'm doing now is cloning the config and also calling `setConfiguration` so that is set locally before calling `SavePluginConfiguration` on the API

I think this should take care of the problem

server/plugin.go
enahum: Perhaps we should validate the config during `onActivate` ?

sbishel (approved):

PR #539, opened 2024-03-07
Adding the remoteID migration, and making the synthetic user identification more robust

jespino:
This should migrate the remoteID from the old users into the plugin remoteID, also ensure that there is always a remote ID, by default is msteams, but as soon as you register the remote on the shared channels it gets updated to a valid one. The reason to have a default remoteID is because not always we call the SharedChannels remote register. Probably we should do it always, and we will do eventually.

The migration is executed in every plugin restart, but in theory it is idenpotent, so it shouldn't be a problem.


server/plugin.go
```diff
@@ -484,6 +482,22 @@ func (p *Plugin) onActivate() error {
 		return err
 	}
 
+	p.remoteID = "msteams"
+	if !p.getConfiguration().DisableSyncMsg {
```
jespino, line 486: Can you confirm if this check is needed? I would like to always register the SharedChannels for the plugin.

jespino: @wiggin77 :point_up:

wiggin77: I think we want an easy way to to turn this off in case of a problem in production.  However we don't want to use a fake remoteID either since that may be hard to clean up later.   I would suggest calling `RegisterPluginForSharedChannels` always, but then calling `UnregisterPluginForSharedChannels` if `p.getConfiguration().DisableSyncMsg == true`.  That way you will have the real remoteID.

wiggin77 (changes requested):
Looks great.  I think we need the real remoteID which we can get by always registering then selectively unregistering based on config.

server/plugin.go
jespino: My concern is the remoteID changing over disabling/enabling. Is the remote id going to be stable or it is going to change every time that I unregister the plugin?

jespino: Anyway, I applied the changes that you proposed.

wiggin77: @jespino  You're right.  I should soft-delete the record on Unregister and clear the DeleteAt on register... or derive the remoteID from the pluginID.  Either way we want the same remoteID for every unique plugin ID.  Any preference?  I think I lend toward hashing the pluginID to make the remoteID.

jespino: Sounds good enough, but also slightly more error prone. I don't think is a big deal.

lieut-data: > Either way we want the same remoteID for every unique plugin ID.

This is probably fine for our current purposes, but thinking (very) long term, I could also see us supporting multiple Microsoft tenants concurrently which might necessitate a distinct `remoteID` for each (if we wanted to, for example, distinguish each in the UI?)

Stepping back, I'm also just wanting to confirm that /we/ (the plugin) don't really care about the value here at all, and that it's just an artifact of the way the shared channels subsystem works with users that happen to have a `RemoteID` set. In other words, we want all DMs and all GMs for all users, regardless of whether anyone has a `RemoteID` set. Does that remain accurate @wiggin77?

wiggin77: If you set the appropriate flag when registering then yes, the MS Teams plugin will get sync for all DM/GMs regardless of remoteID value.  Other plugins using the API will likely not set that flag and the remoteID on posts/users/etc will matter.  

Should we modify the register API to accept a remoteID, thus allowing a plugin to call register for each tenant?  That would also leave it up to the plugin to ensure idempotency.   cc @jespino @lieut-data

lieut-data: > Should we modify the register API to accept a remoteID, thus allowing a plugin to call register for each tenant? That would also leave it up to the plugin to ensure idempotency. cc @jespino @lieut-data

On the one hand, I don't want to preemptively build for the future. But I do like the idea of giving the plugin control over the "primary key" backing the cursor based model. At the moment, it feels just a bit too decoupled. So I'm in favour of the idea. Is the `RemoteId` constrained by some length we should remain aware of? (i.e. is it long enough for a UUID?)

wiggin77: It's the same length as most IDs in MM.

server/ce2e/plugin_test.go
```diff
@@ -379,11 +379,11 @@ func TestSelectiveSync(t *testing.T) {
 
 	conn, err := mattermost.PostgresConnection(context.Background())
 	require.NoError(t, err)
+	defer conn.Close()
 
 	// Mark user as synthetic
-	_, err = conn.Exec("UPDATE Users SET RemoteId = 'syntetic-user' WHERE Id = $1", synthetic.Id)
+	_, err = conn.Exec("UPDATE Users SET RemoteId = (SELECT remoteId FROM remoteclusters) WHERE Username = 'msteams_synthetic'")
```
lieut-data, line 385: Let's constrain `(SELECT remoteId FROM remoteclusters)` to only returning the `remoteId` for this plugin, otherwise this could return multiple results and fail.

server/plugin.go
```diff
@@ -484,6 +482,24 @@ func (p *Plugin) onActivate() error {
...
+		AutoShareDMs: true,
+		AutoInvited:  true,
+	})
+	if err != nil {
+		return err
+	}
+	p.API.LogInfo("Registered plugin for shared channels", "remote_id", p.remoteID)
+
+	if p.getConfiguration().DisableSyncMsg {
+		if err = p.API.UnregisterPluginForSharedChannels(pluginID); err != nil {
```
lieut-data, line 498: Propose a log here to clarify that we did indeed disable it:
```
p.API.LogInfo("Unregistering plugin for shared channels since sync msg disabled")
```

server/store/sqlstore/migrations.go
```diff
@@ -0,0 +1,15 @@
+package sqlstore
+
+import (
+	sq "github.com/Masterminds/squirrel"
+)
+
+func (s *SQLStore) runMigrationRemoteID(remoteID string) error {
+	_, err := s.getQueryBuilder().Update("Users").Set("RemoteID", remoteID).Where(sq.And{
+		sq.NotEq{"RemoteID": nil},
+		sq.NotEq{"RemoteID": ""},
```
lieut-data, line 10: To confirm, all users we've ever created have had a non-nil, non-empty `RemoteID`?

server/testutils/containere2e/containere2e.go
```diff
@@ -117,6 +117,7 @@ func NewE2ETestPlugin(t *testing.T, extraOptions ...mmcontainer.MattermostCustom
 	options := []mmcontainer.MattermostCustomizeRequestOption{
 		mmcontainer.WithPlugin(filename, "com.mattermost.msteams-sync", pluginConfig),
 		mmcontainer.WithLogConsumers(&tLogConsumer{t: t}),
+		mmcontainer.WithEnv("MM_EXPERIMENTALSETTINGS_ENABLESHAREDCHANNELS", "true"),
```
lieut-data, line 120: @wiggin77, can you remind me where things stand in terms of rolling this out to production? Will we require customers to enable this experimental setting (and the enduser functionality) before we can rely on shared channels, or is there a ticket to track decoupling these?

lieut-data (approved):
Thanks, @jespino! There's one blocker below, but it's in the unit tests, otherwise just a handful of non-blocking questions so I'm approving to unblock.

server/ce2e/plugin_test.go
jespino: As it is part of the plugin test, my expectation is there is nothing else there, I can add it, but it shouldn't be a problem.

server/plugin.go
jespino: Good call

server/store/sqlstore/migrations.go
jespino: All the synthetic users, by definition, are the ones that have `RemoteID` different that nil/"", so. Not "all the users" that we have created, the bot user, and any promoted user has no `RemoteID`, but that is the expected behavior.

server/testutils/containere2e/containere2e.go
jespino: Let me remove that, I think I added there while I was trying to figure out the how to fix the test while I was working on the PR, but if I can remove that line and keep going, works for me.

server/store/sqlstore/migrations.go
lieut-data: Awesome, thanks for confirming! (Mostly was just thinking back to early deployments with any customers and making sure there wasn't a time when `RemoteID` was left `nil`.)

server/plugin.go
```diff
@@ -498,6 +498,7 @@ func (p *Plugin) onActivate() error {
 		if err = p.API.UnregisterPluginForSharedChannels(pluginID); err != nil {
 			p.API.LogWarn("Unable to unregister plugin for shared channels", "error", err)
 		}
+		p.API.LogInfo("Unregistering plugin for shared channels since sync msg disabled")
```
lieut-data, line 501: nit: can we put this before we make the call? (If it fails and we log a warning, the logs won't quite make sense.)

lieut-data (approved):

server/plugin.go
jespino: :+1:

wiggin77 (approved):
LGTM :+1:

PR #543, opened 2024-03-08
more verbose logging around subscriptions

lieut-data:
#### Summary
Add some extra info logs around subscriptions, giving visibility into changes that are occurring. Add some godoc while we're in here.

#### Ticket Link
None

### DIFF
--- a/server/monitor/subscriptions.go
+++ b/server/monitor/subscriptions.go
@@ -11,6 +11,18 @@ import (
 	"github.com/mattermost/mattermost-plugin-msteams/server/store/storemodels"
 )
 
+func isExpired(expiresOn time.Time) bool {
+	return expiresOn.Before(time.Now())
+}
+
+func shouldRefresh(expiresOn time.Time) bool {
+	return time.Until(expiresOn) < (5 * time.Minute)
+}
+
+// checkChannelsSubscriptions maintains the per-channel subscriptions, creating one if it doesn't
+// already exist or refreshing the expiry time as needed.
+//
+// Subscriptions are refreshed in parallel, processing at most 20 concurrently.
 func (m *Monitor) checkChannelsSubscriptions(msteamsSubscriptionsMap map[string]*clientmodels.Subscription) {
 	defer func() {
 		if r := recover(); r != nil {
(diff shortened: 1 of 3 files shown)


server/monitor/subscriptions.go
```diff
@@ -59,21 +71,35 @@ func (m *Monitor) checkChannelsSubscriptions(msteamsSubscriptionsMap map[string]
...
 					return
 
 				case mmSubscription.Certificate != m.certificate:
+					m.api.LogInfo("Recreating subscription for channel on certificate mismatch", "channel_id", link.MattermostChannelID, "team_id", link.MattermostTeamID, "teams_channel_id", link.MSTeamsChannel, "teams_team_id", link.MSTeamsTeam)
+
 					m.recreateChannelSubscription(mmSubscription.SubscriptionID, mmSubscription.TeamID, mmSubscription.ChannelID, mmSubscription.Secret, true)
 
-				case time.Until(mmSubscription.ExpiresOn) < (5 * time.Minute):
+				case shouldRefresh(mmSubscription.ExpiresOn):
+					if isExpired(mmSubscription.ExpiresOn) {
```
lieut-data, line 87: Making it really explicit if we have to refresh an expired subscription.

server/testutils/mock.go
```diff
@@ -14,8 +14,8 @@ func GetMockArgumentsWithType(typeString string, num int) []interface{} {
...
-	mockAPI.On("LogWarn", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil).Maybe()
-	mockAPI.On("LogError", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil).Maybe()
+	mockAPI.On("LogDebug", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil).Maybe()
+	mockAPI.On("LogInfo", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil).Maybe()
+	mockAPI.On("LogWarn", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil).Maybe()
+	mockAPI.On("LogError", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(nil).Maybe()
 }
```
lieut-data, line 21: This will all be gone soon. For now, mock an extra pair of parameters :(

jespino (approved):
LGTM

lieut-data:
@wiggin77, just a small review here, hoping to get this into v1.10, in particular because I'm continuing to see really slow behaviour around subscription renewal, recently as high as multiple minutes :(. The logging here will just give some visibility into our own actions being taken.

server/monitor/subscriptions.go
```diff
@@ -155,28 +184,42 @@ func (m *Monitor) checkGlobalChatsSubscription(msteamsSubscriptionsMap map[strin
...
 			m.api.LogError("Unable to recreate all chats subscription", "error", err.Error())
 		}
 	}
 
-	if time.Until(mmSubscription.ExpiresOn) < (5 * time.Minute) {
+	if shouldRefresh(mmSubscription.ExpiresOn) {
+		if isExpired(mmSubscription.ExpiresOn) {
+			// In the future, this won't need to be an error if we can resync, but for
+			// now notify a human.
+			m.api.LogError("Global chats subscription expired")
```
wiggin77, line 206: Can we provide some context or next steps for that human?

wiggin77 (approved):
LGTM :+1:   Thanks for adding the comments.  I notice we're hoisting some logs that ordinarily would be Debug to Info.  I assume this is to avoid having to turn on debug logging and get a fire-hose.  I'll have a better way to handle this soon in the Logr lib.

lieut-data:
Thanks, @wiggin77!

> I notice we're hoisting some logs that ordinarily would be Debug to Info.

At least for this PR, no change to any Debug logs -- in fact, I'm proposing these should stay at "info", so enabled for all users by default. (My thinking is "low frequency actions" are good candidates for keeping at info.)

server/monitor/subscriptions.go
lieut-data: I like this idea, though would propose we host that externally, e.g. "what to do on ERROR log <...>". The number of actual error logs is now quite manageable, thankfully.

PR #544, opened 2024-03-08
MM-56858: Automated Connection Invites

calebroseland:
#### Summary
When `connectedUsersInvitePoolSize` is specified, connection invite DMs will be sent out as users login if there are open spaces available. 

ConnectionInvites drawio

- [x] Connection Invites:
  - On weekdays,
  - if there is an open spot in the invite pool,
  - when a user
    - ~~logs in, or~~
    - ~~refreshes, or~~
    - is active (click-check at most once per hour, after at least 0-10 minutes from initial login/refresh),
  - then a connection invite will be sent via DM.
- [x] - The number of invites is capped at the number remaining seats available according to `Max Connected Users`.
- [x] - Connection invites may not be retracted—invited users may connect as if they are on the whitelist. 
- ~~[x] - Config: `connectedUsersAllowed < whitelistedUsers + invitedUsers` is considered invalid.~~

#### Screenshots

`<1000` connected users
CleanShot 2024-03-12 at 16 10 06@2x

`>1000` connected users
CleanShot 2024-03-12 at 16 18 50@2x

CleanShot 2024-03-12 at 16 22 57@2x

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-56858

#### Related Pull Requests
- #534


plugin.json
```diff
@@ -129,11 +129,17 @@
...
+          "key": "connectedUsersAllowed",
+          "display_name": "Max Connected Users",
+          "type": "number",
+          "help_text": "The maximum number of users that may connect their MS Teams account. Once connected, the user is added to a white-list and may reconnect at any time.",
+          "default": 1000
+        },{
+          "key": "connectedUsersInvitePoolSize",
+          "display_name": "Connection Invites: Max Pending Invitations",
+          "type": "number",
+          "help_text": "Invite pool size: the maximum number of connection invites that may be pending at a given time. When enabled, connection invite DMs will be sent to users as they login, up to the maximum specified here. As invited users connect, spaces in the invite pool will open up and more invites will be sent out. (Leave empty to disable connection invites.)",
```
lieut-data, line 141: By "enabled" or "empty", do we mean "non-zero" or "0"? 0/5 on the language, just wanting to clarify.

server/api.go
```diff
@@ -334,9 +335,15 @@ func (a *API) connect(w http.ResponseWriter, r *http.Request) {
...
+	http.Redirect(w, r, connectURL, http.StatusSeeOther)
+}
 
-	data, _ := json.Marshal(map[string]string{"connectUrl": connectURL})
-	_, _ = w.Write(data)
+func (a *API) notifyConnect(w http.ResponseWriter, r *http.Request) {
+	userID := r.Header.Get("Mattermost-User-ID")
+
+	if _, err := a.p.MaybeSendInviteMessage(userID); err != nil {
+		a.p.API.LogWarn("Error in connection invite flow", "error", err.Error())
```
lieut-data, line 345: Can we log the `user_id` here and in the similar logs below?

server/api.go
```diff
@@ -462,6 +476,13 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	err = a.p.store.DeleteUserInvite(mmUserID)
```
lieut-data, line 479: This got me thinking: what happens if the user disconnects? Should we suppress inviting them again?

server/api.go
```diff
@@ -462,6 +476,13 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		a.p.API.LogWarn("Unable to clear user invite", "error", err.Error())
+		http.Error(w, "Something went wrong", http.StatusInternalServerError)
+		return
```
lieut-data, line 483: To confirm, do we definitely want to interrupt the rest of the flow should this fail? (And unrelated to the PR, but we have waaay too much logic in the API layer :p.)

server/configuration.go
```diff
@@ -85,8 +86,13 @@ func (p *Plugin) validateConfiguration(configuration *configuration) error {
...
-		if configuration.ConnectedUsersAllowed < whitelistSize {
-			return errors.New("failed to save configuration, no. of connected users allowed should be greater than or equal to the current size of the whitelist")
+		invitedSize, err := p.store.GetSizeOfInvitedUsers()
+		if err != nil {
+			return errors.New("failed to get the size of invited users from the DB")
+		}
+
+		if configuration.ConnectedUsersAllowed < (whitelistSize + invitedSize) {
+			return errors.New("failed to save configuration, no. of connected users allowed should be greater than or equal to the current size of the whitelist and invited users")
 		}
```
lieut-data, line 96: I'm wondering if we should just rip out this check altogether. It might be hard for the enduser to see what the current size of each is, and presumably if the `ConnectedUsersAllowed` is too small, it just means no one new can join? (Might simplify this code path!)

server/store/sqlstore/store.go
```diff
@@ -981,6 +995,81 @@ func (s *SQLStore) IsUserPresentInWhitelist(userID string) (bool, error) {
...
+		Where(sq.Eq{"mmUserID": mmUserID})
+
+	rows, err := query.Query()
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	var result *storemodels.InvitedUser
+	if rows.Next() {
```
lieut-data, line 1028: Since we only expect one, should we just query directly for a [single row](https://go.dev/doc/database/querying#single_row)? Note that this would need explicit handling for the not found case so we can continue to return `nil, nil` in that scenario.

server/store/sqlstore/store.go
```diff
@@ -981,6 +995,81 @@ func (s *SQLStore) IsUserPresentInWhitelist(userID string) (bool, error) {
...
+	return result, nil
+}
+
+func (s *SQLStore) DeleteUserInvite(mmUserID string) error {
+	if _, err := s.getQueryBuilder().Delete(invitedUsersTableName).Where(sq.Eq{"mmUserID": mmUserID}).Exec(); err != nil {
+		return err
+	}
+
+	return nil
+}
```
lieut-data, line 1053: I wonder if should instead mark this as `DeleteAt = NOW()`, with the idea being that we preserve this metadata and possibly also use it to decide if we resend to a user that was previously invited?

server/store/sqlstore/store.go
```diff
@@ -981,6 +995,81 @@ func (s *SQLStore) IsUserPresentInWhitelist(userID string) (bool, error) {
...
+
+func (s *SQLStore) GetSizeOfInvitedUsers() (int, error) {
```
lieut-data, line 1055: nit: `size` vs. `count`, might be easier to understand as `GetPendingUserInvitesCount()` or some such.

webapp/src/index.tsx
```diff
@@ -79,11 +82,35 @@ export default class Plugin {
...
     uninitialize() {
         this.removeStoreSubscription?.();
     }
 }
 
+const RootEffects = () => {
+    useEffect(() => {
+        Client.notifyConnect();
+    }, []);
+
```
lieut-data, line 110: I wonder: given we do this on the first click, do we need this at all? I'm thinking of a scenario where a user's desktop app updates or some such and reloads without their being involved. Should we wait until the user is really around?

server/connect.go
```diff
@@ -0,0 +1,149 @@
...
+
+	if err := p.store.StoreInvitedUser(invitedUser); err != nil {
+		p.API.LogWarn("Error storing user in invite list", "error", err.Error())
+		return err
+	}
+
+	connectURL := p.GetURL() + "/connect"
+
+	var message string
+	if nWhitelisted < earlyAdopterThreshold {
```
lieut-data, line 105: Propose we simplify to a single message vs distinguishing (and hard-coding) a threshold for early adopters.

server/connect.go
```diff
@@ -0,0 +1,149 @@
...
+		return false, err
+	}
+
+	if invitedUser != nil {
+		pendingSince = invitedUser.InvitePendingSince
+	} else {
+		moreInvitesAllowed, n, err := p.moreInvitesAllowed()
+		if err != nil {
+			p.API.LogWarn("Error checking invite pool size", "error", err.Error())
+			return false, errors.Wrapf(err, "Error checking invite pool size")
```
lieut-data, line 68: nit: propose we consistently either leave the caller to log this and keep returning the error, or we choose to log and don't return an error. (Probably the former, I'm guessing?)

server/connect.go
```diff
@@ -0,0 +1,149 @@
...
+	})
+}
+
+func (p *Plugin) MaybeSendInviteMessage(userID string) (bool, error) {
+	if p.getConfiguration().ConnectedUsersInvitePoolSize == 0 {
+		// connection invites disabled
+		return false, nil
+	}
+
+	var nWhitelisted int
```
lieut-data, line 35: nit: can we declare these closer to where they are initialized?

server/connect.go
```diff
@@ -0,0 +1,149 @@
...
+	return true, nil
+}
+
+func (p *Plugin) SendInviteMessage(user *model.User, pendingSince time.Time, currentTime time.Time, nWhitelisted int) error {
+	invitedUser := &storemodels.InvitedUser{ID: user.Id, InvitePendingSince: pendingSince, InviteLastSentAt: currentTime}
+	if invitedUser.InvitePendingSince.IsZero() {
+		invitedUser.InvitePendingSince = currentTime
+	}
+
+	if err := p.store.StoreInvitedUser(invitedUser); err != nil {
```
lieut-data, line 97: I'm assuming this guarantees we store the user regardless of the current whitelist limit?

Also, is there a scenario where a user happens to trigger this for themselves twice because they have two tabs open (e.g. Channels and Playbooks)? I'm wondering if avoiding this is as simple as distinguishing the `INSERT` from the `UPSERT` case in the store? (The database would serialize for us.)

server/connect.go
```diff
@@ -0,0 +1,149 @@
...
+		return err
+	}
+
+	connectURL := p.GetURL() + "/connect"
+
+	var message string
+	if nWhitelisted < earlyAdopterThreshold {
+		message = fmt.Sprintf("@%s, you're invited to be an early adopter for the MS Teams connected experience. [Click here to connect your account](%s).", user.Username, connectURL)
+	} else {
+		message = fmt.Sprintf("@%s, you're invited to use the MS Teams connected experience. [Click here to connect your account](%s).", user.Username, connectURL)
```
lieut-data, line 108: Would it be easy to send back a button to click vs. a link, for just a slightly more polished look? (Using interactive actions only, so it's mobile compatible.)

server/connect.go
```diff
@@ -0,0 +1,149 @@
...
+			// user not connected, but invite threshold is presently met
+			return false, nil
+		}
+
+		nWhitelisted = n
+	}
+
+	if !p.shouldSendInviteMessage(pendingSince, now, user.GetTimezoneLocation()) {
+		return false, nil
+	}
```
lieut-data, line 81: In a scenario where people arrive at 9am and all login within a minute of each other, is there a thundering herd problem here where we substantially over-invite? Should we force the `notify-connect` to be called stochastically over a (1,10) minute window to prevent this?

server/connect.go
```diff
@@ -0,0 +1,149 @@
...
+
+		if !moreInvitesAllowed {
+			// user not connected, but invite threshold is presently met
+			return false, nil
+		}
+
+		nWhitelisted = n
+	}
+
+	if !p.shouldSendInviteMessage(pendingSince, now, user.GetTimezoneLocation()) {
```
lieut-data, line 79: I seem to recall some weirdness around `GetTimezoneLocation` possibly being `nil` or invalid -- do we handle that well?

server/connect.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	return p.botSendDirectMessage(user.Id, message)
+}
+
+func (p *Plugin) shouldSendInviteMessage(
+	pendingSince time.Time,
+	currentTime time.Time,
+	timezone *time.Location,
+) bool {
+	now := currentTime.In(timezone)
```
lieut-data, line 118: If `timezone` was invalid (hopefully not `nil`!), should we default to something sensible like Pacific? (Or maybe even the server configured timezone?)

lieut-data (changes requested):
This is fantastic work, @calebroseland! You've dug deep to fit in with the current model and extend it elegantly. A few thoughts below to discuss.

plugin.json
calebroseland: Updated to "When specified, ..." and "Set to 0 or leave empty to..."

server/api.go
calebroseland: Once they've connected for the first time, they're added to the whitelist and not invited again.

server/api.go
calebroseland: That's a fair point—if it's gotten this far and only errors here, it would probably be fine to continue on. Will change.

server/configuration.go
calebroseland: I'm not opposed to that. I had added similar check to validate that the invite pool size didn't go over seats available, but that was a bit over the top and might have caused problems towards the top end of max connected users.

Similarly, I guess this check in question could cause the same issue... 

Let's say max 200, with 100 seats open, and 100 invites then get sent out... if non-invited users connect (as allowed because `whitelistSize >= a.p.getConfiguration().ConnectedUsersAllowed` is still false), that would cause this check to fail.

server/api.go
```diff
@@ -444,7 +451,14 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	if whitelistSize >= a.p.getConfiguration().ConnectedUsersAllowed {
+	invitedUser, err := a.p.store.GetInvitedUser(mmUserID)
+	if err != nil {
+		a.p.API.LogWarn("Error in getting invited user", "error", err.Error())
+		http.Error(w, "Something went wrong", http.StatusInternalServerError)
+		return
+	}
+
+	if whitelistSize >= a.p.getConfiguration().ConnectedUsersAllowed && invitedUser == nil {
```
calebroseland, line 461: @lieut-data I think we need this change to ensure non-invited new connections don't "jump the line" during the last batch of invites.

e.g.: max 100, 90 connected, 10 pending invites—then 5 non-invited users connect, then the 10 invites connect—now we're at 105 connected


```suggestion
	if (whitelistSize + invitedSize) >= a.p.getConfiguration().ConnectedUsersAllowed && invitedUser == nil {
```

server/connect.go
calebroseland: Will go with logging at the API layer, and leave these to simply return errors

server/connect.go
calebroseland: 0/5, happy to remove, just following https://docs.google.com/document/d/14Wu18J5Ss-rnxlwmFRiMg2Qt4qih3Qo_Q-0bO4rkBP8/edit?usp=sharing. @johndavidlugtu good with you to remove?

server/store/sqlstore/store.go
calebroseland: Happy to do that if better, but this was mostly copied from whitelist store methods for expediency

server/store/sqlstore/store.go
calebroseland: That is indeed better, but I'm partially opposed to it as it would introduce `include_deleted` complexity, and it may cause mis-matched semantics vs the whitelist store methods.

I also don't see us needing to keep the exact timestamp the invite was originally sent because the invite is preserved until first connect, after which we have the whitelist.

Reading between the lines, I think `OnDisconnectedTokenHandler` handles that sort of flow better than resending an initial connection invite.

server/store/sqlstore/store.go
calebroseland: Agreed, but copied from `GetSizeOfWhitelist` 1/5? 4/5?

server/connect.go
calebroseland: Yes, I suppose `INSERT ` and insertion error on conflict would save us, but,

I think we may need a mutex for this the last-invite scenario—to prevent a race condition there? (To avoid sending N+1 invites?)

webapp/src/index.tsx
calebroseland: That's a fair point; I don't see any issue with it—though I'd have to change the click handler logic to be permitted during the first hour.

server/connect.go
calebroseland: Will look into, not 100% how tight this is currently

server/connect.go
calebroseland: That's a great idea; will do 👍

server/store/sqlstore/store.go
lieut-data: Got it: no need to fix now, but there's definitely some technical debt in the store to resolve long term.

server/store/sqlstore/store.go
lieut-data: Sounds good, especially given the whitelist comment you shared earlier preventing re-invites.

server/store/sqlstore/store.go
lieut-data: -1/5 for now, something to go through and fix en masse later :)

server/api.go
lieut-data: That makes sense to me!

server/api.go
sbishel: Probably beyond scope, but shouldn't this portion of the function be transactional...completely succeed or fail. I didn't completely study it, but with all the error checks and database calls...both storing and retrieving a database error may leave the user in an unknown state.

sbishel (commented):
Looks good.

Not part of this PR, but the use of the term "whitelist" is confusing to me and I think unnecessary.

Does "whiteList" really mean "connected users"? I see we pre-fill the white list table with all the connected users and add to the white list table whenever a user connects. 

So isn't the whitelist table the same as 
`Select msteamssync_users with token != ""`
This is what PrefilWhiltelist does. So do we even need the `msteamsync_whitelisted_users` table?

Isn't the new table "msteamssync_invited_users" really a WhiteList?
cc: @lieut-data

server/connect.go
calebroseland: Done

server/connect.go
```diff
@@ -32,20 +28,17 @@ func (p *Plugin) MaybeSendInviteMessage(userID string) (bool, error) {
...
 
-	var nWhitelisted int
-	var pendingSince time.Time
-	now := time.Now()
-
 	user, err := p.apiClient.User.Get(userID)
 	if err != nil {
-		p.API.LogWarn("Error getting user", "user_id", userID, "error", err.Error())
-		return false, err
+		return false, errors.Wrapf(err, "Error getting user")
```
lieut-data, line 33: nit: lower case errors
```suggestion
		return false, errors.Wrapf(err, "error getting user")
```

webapp/src/index.tsx
```diff
@@ -85,15 +86,15 @@ export default class Plugin {
     userActivityWatch(): void {
         // Listen for new activity to trigger a call to the server
         // Hat tip to the Github and Playbooks plugin
-        let lastActivityTime = Number.MAX_SAFE_INTEGER;
-        const activityTimeout = 60 * 60 * 1000; // 1 hour
+        let nextCheckAfter = Date.now() + Math.max(MINUTE, randomInt(10) * MINUTE);
```
lieut-data, line 89: I think this creates just ~9 "buckets" across which requests are randomized (1 minute, 2 minutes, ..., 9 minutes). Could we tweak this to randomize across seconds?

```suggestion
        let nextCheckAfter = Date.now() + Math.max(MINUTE, randomInt(10 * MINUTE));
```

lieut-data (approved):
Thanks, @calebroseland! A handful of non-blocking comments below.

plugin.json
```diff
@@ -129,11 +129,17 @@
...
+          "type": "number",
+          "help_text": "The maximum number of users that may connect their MS Teams account. Once connected, the user is added to a white-list and may reconnect at any time.",
+          "default": 1000
+        },{
+          "key": "connectedUsersInvitePoolSize",
+          "display_name": "Connection Invites: Max Pending Invitations",
+          "type": "number",
+          "help_text": "Invite pool size: the maximum number of connection invites that may be pending at a given time. When specified, connection invite DMs will be sent to users as they login, up to the maximum specified here. As invited users connect, spaces in the invite pool will open up and more invites will be sent out. (Set to 0 or leave empty to disable connection invites.)",
+          "default": 0
         },{
```
calebroseland, line 143: @cwarnermm

sbishel (approved):
Thanks Caleb. Yeah. For all of my questions, I don't want you changing this PR, but wanted to ensure I understood.

plugin.json
cwarnermm: @calebroseland - This reads well; my only nit would be ``DMs`` -- consider spelling out in full as ``direct messages`` to communicate with as wide an audience as possible.

PR #546, opened 2024-03-11
Allow self chats to sync when selective sync is true

sbishel:
#### Summary
Currently, with Selective Sync turned on, self chats will never sync. This is actually a valuable test tool and may people may "test" the system using there self channels. 

This PR checks the number of members in the channel and allows it to sync if there is only 1.


lieut-data (approved):
Nice!

lindy65 (approved):
Thanks @sbishel - will test post merge

PR #548, opened 2024-03-12
[MM-57209] Set EmailVerified to true when promoting synthetic users

JulienTant:
#### Summary
Set EmailVerified to true when promoting a synthetic user. 	

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57209

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -632,6 +632,7 @@ func (p *Plugin) executePromoteUserCommand(args *model.CommandArgs, parameters [
 
 	user.RemoteId = nil
 	user.Username = newUsername
+	user.EmailVerified = true
 	_, appErr = p.API.UpdateUser(user)
 	if appErr != nil {
 		p.sendBotEphemeralPost(args.UserId, args.ChannelId, "Error: Unable to promote account "+username)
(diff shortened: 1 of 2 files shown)


BenCookie95 (approved):

lieut-data (changes requested):
Thank you, @JulienTant! Can you also make these changes in the `syncJob` that's in `plugin.go`? This is where we first create synthetic users, and where we'll need to support updating these in bulk as well, as some users are promoted automatically without the explicit slash command.

JulienTant:
Thanks, @lieut-data! I applied the changes to the sync users method.

Applying those changes, and adding tests, made me realize that the tests for this method were not working because the MSUser was created with a `Mail` field, which would cause the main loop to `continue` without doing anything (first if of the loop). I added the `Mail` field and fixed the mock calls as needed + added 2 tests for user creation and update.

server/testutils/data.go
```diff
@@ -102,6 +102,16 @@ func GetUser(role, email string) *model.User {
 	}
 }
 
+func GetRemoteUser(role, email, remoteID string) *model.User {
+	return &model.User{
+		Id:       GetID(),
+		Username: "test-user",
+		Roles:    role,
+		Email:    email,
+		RemoteId: &remoteID,
```
lieut-data, line 111: For completeness, should we set `EmailVerified` to true for this test user?

lieut-data (approved):
Thanks, @JulienTant!

server/testutils/data.go
JulienTant: Hum, I purposely left it to `false` to make sure my code set it to `true` 😅

lieut-data: No, that's fine too! Just wanted to clarify -- your call either way :)

lieut-data (approved):

lindy65 (approved):
Thanks @JulienTant - this will be tested post-merge, QA-deferring.

PR #550, opened 2024-03-12
[MM-56908] Make promote command accepts @ for both arguments

JulienTant:
#### Summary
When running the `promote` command, the first argument is accepted with or without a `@`, while the second does not support a `@` at all. This PR allows both arguments to be prefixed by a `@`, that will immediately be trimmed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56908


jespino (approved):
LGTM

JulienTant:
@lindy65 deferring QA on this to get it in!

PR #551, opened 2024-03-13
Adding more tests to the selective sync

jespino:


server/ce2e/plugin_test.go
```diff
@@ -555,25 +551,233 @@ func TestSelectiveSync(t *testing.T) {
...
+									},
+								},
+								"body": map[string]any{
+									"contentType": "text",
+									"content":     "Hello World",
+								},
+								"channelIdentity": map[string]any{
+									"channelId": newDMID,
+								},
+							},
```
enahum, line 586: I see no difference between the three content created for the body, define as a variable and reduce code?

server/ce2e/plugin_test.go
```diff
@@ -555,25 +551,233 @@ func TestSelectiveSync(t *testing.T) {
...
+						Method: http.MethodGet,
+						URL:    "/chats/" + newDMID + "/messages/" + newPostID,
+					}},
+					[]containere2e.BatchResponse{
+						{
+							StatusCode: http.StatusNoContent,
+						},
+						{
+							StatusCode: http.StatusOK,
+							Body:       fakePost,
```
enahum, line 656: Exactly as it is done here

server/testutils/containere2e/apimock.go
```diff
@@ -126,6 +145,57 @@ func (m *MockClient) Delete(id string, url string, body map[string]any) error {
...
+	req, err := http.NewRequest("PUT", m.api+"/mockserver/expectation", bytes.NewReader(testMock))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
```
enahum, line 195: Any reason to close the body here? This is just for my own understanding as I'm doing something similar with the load test

enahum (approved):
Lgtm

server/testutils/containere2e/apimock.go
jespino: That is expected in go to avoid some leaks that can happen. Whenever you do a request, you should close the body of the response, it is recommended even if you don't read it.

PR #545, opened 2024-03-11
Improve Users sync

enahum:
#### Summary
Refactor syncUsers so it reads better and  fixes the issue where users being activated and deactivated constantly.

We now do not create synthetic users if they are not present in MM but disabled on MS Teams

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56942
https://mattermost.atlassian.net/browse/MM-56886

PR #553, opened 2024-03-15
Fix username dedup for synthetic users

JulienTant:
#### Summary
When creating a new synthetic user, the synthetic user is being given a username of `msteams_[username]`. If this is not available, we want to suffix this with `-{idx}`. The current code would add `-{idx}` at the end of each iteration, leading to the username looking like `msteams_[username]-1-2-3-4` instead of `msteams_[username]-4`.

This PR fixes this issue.

#### Ticket Link
N/A


lieut-data (approved):
Nice!

enahum (approved):
This may conflict with https://github.com/mattermost/mattermost-plugin-msteams/pull/545

JulienTant:
@enahum  I don't mind dealing with the conflict if you wanna merge your branch first! I don't think this is super urgent (We would need at least three users with the exact same username for this to happen)

PR #555, opened 2024-03-18
MM-56788 notify need to connect

lieut-data:
#### Summary
Streamline the notification sent when a user tries to communicate in a DM/GM with a user relying on Teams:
* Remove the `promptIntervalForDMsAndGMs` configuration: we now send this for every message, but only ephemerally.
* We no longer send these for reactions, updates or deletes: we only send for new messages (where the ephemeral reply is likely to be noticed)
* We only send these if the chat members span platforms: there's no reason to warn users for messages that would be delivered normally in Mattermost.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-56788


server/message_hooks.go
```diff
@@ -65,22 +64,22 @@ func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
...
 
-		if p.getConfiguration().SelectiveSync {
-			shouldSync, appErr := p.ChatMembersSpanPlatforms(members)
-			if appErr != nil {
-				p.API.LogWarn("Failed to check if chat should be synced", "error", appErr.Error(), "post_id", post.Id, "channel_id", post.ChannelId)
-				return
-			} else if !shouldSync {
-				return
-			}
+		chatMembersSpanPlatforms, appErr := p.ChatMembersSpanPlatforms(members)
```
lieut-data, line 67: We look up `chatMembersSpanPlatforms` for use in deciding if we need to notify the user of the need to connect.

server/message_hooks.go
```diff
@@ -255,13 +254,11 @@ func (p *Plugin) MessageHasBeenUpdated(c *plugin.Context, newPost, _ /*oldPost*/
...
 	srcUserID, err := p.store.MattermostToTeamsUserID(srcUser)
 	if err != nil {
-		p.handlePromptForConnection(srcUser, channelID)
 		return err
 	}
 
 	client, err := p.GetClientForUser(srcUser)
 	if err != nil {
-		p.handlePromptForConnection(srcUser, channelID)
 		return err
```
lieut-data, line 262: We don't bother notifying in these cases: the ephemeral post appears at the bottom in all cases, and not necessarily in tandem with the reaction (or edit or delete).

server/message_hooks.go
```diff
@@ -454,13 +449,17 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post) (
 
 	_, err := p.store.MattermostToTeamsUserID(srcUser)
 	if err != nil {
-		p.handlePromptForConnection(srcUser, post.ChannelId)
+		if chatMembersSpanPlatforms {
+			p.handlePromptForConnection(srcUser, post.ChannelId)
+		}
```
lieut-data, line 454: We use `chatMembersSpanPlatforms` to decide whether to notify: if it doesn't span, the user is expected to receive it in Mattermost and there's no "need" to connect. (We still try to `SendChat` when `selectiveSync` is disabled if bidirectional sync is expected.)

server/message_hooks.go
```diff
@@ -551,23 +550,7 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post) (
...
-	}
-
-	if time.Until(timestamp) < -time.Hour*time.Duration(promptInterval) {
-		p.sendBotEphemeralPost(userID, channelID, "Your Mattermost account is not connected to MS Teams so your activity will not be relayed to users on MS Teams. You can connect your account using the `/msteams connect` slash command.")
-
-		if err = p.store.StoreDMAndGMChannelPromptTime(channelID, userID, time.Now()); err != nil {
-			p.API.LogWarn("Unable to store the last prompt timestamp for the channel", "channel_id", channelID, "error", err.Error())
-		}
-	}
+	p.sendBotEphemeralPost(userID, channelID, "Some users in this conversation rely on Microsoft Teams to receive your messages, but your account is not connected. Connect your account using the `/msteams connect` slash command.")
```
lieut-data, line 553: @cwarnermm, might I again get your feedback on the proposed language here?

cwarnermm: @lieut-data - For this one, I propose: "Some users in this conversation rely on Microsoft Teams to receive your messages, but your account isn't connected. Connect your account using the `/msteams connect` slash command."

johndavidlugtu:
> We no longer send these for reactions, updates or deletes: we only send for new messages (where the ephemeral reply is likely to be noticed)

@lieut-data is the idea here that if a user reacts to a post and isn't connected we don't send an ephemeral message? Or is the idea we don't send an ephemeral message to the user notifying them that syncing of emoji reactions are not supported?

Or both?

lieut-data:
> if a user reacts to a post ... we don't send an ephemeral message?

@johndavidlugtu, the idea is we never send a message. Ephemeral messages can only appear at the bottom of a thread or channel, and aren't generally suitable for warnings as such.

I'm optimistic that we can restore support for reactions in a timely fashion, minimizing this concern.

johndavidlugtu:
> I'm optimistic that we can restore support for reactions in a timely fashion, minimizing this concern.

This definitely reduces the concern.

@asaadmahmood should we support reactions for June, one thing for consideration is if and how we communicate the fact that MS Teams only supports 1 emoji reactions per message

asaadmahmood:
@johndavidlugtu I think we should inform the user about that beforehand, informing them everytime they react to a file can be overwhelming.
Maybe the first time they react to someone, we send them a bot message, let me think about it.

JulienTant:
@lieut-data any chance main can be merged in? The `notify-attachments-not-synced` branch is showing on the diff and makes it hard to focus on the topic!

lieut-data:
@JulienTant, updated from `main` -- however the `notify-attachments-not-synced` branch should have already disappeared after the base changed automatically. I suspect this clean up commit is causing most of the noise: https://github.com/mattermost/mattermost-plugin-msteams/pull/555/commits/25cbf72654ab1b3ea76e9c797d7565965da61435

JulienTant (approved):
LGTM @lieut-data! I also find the refactoring in the `MessageHasBeenPosted` to be much clearer this way 🎉

lieut-data:
Moving this back to v1.10, but will ship in a patch after the bug bash today.

calebroseland (approved):

PR #558, opened 2024-03-21
Disable temporoarily auto share DMs

jespino:


lieut-data (approved):

wiggin77 (approved):

server/plugin.go
```diff
@@ -488,7 +488,7 @@ func (p *Plugin) onActivate() error {
 		Displayname:  pluginID,
 		PluginID:     pluginID,
 		CreatorID:    p.userID,
-		AutoShareDMs: true,
+		AutoShareDMs: false,
 		AutoInvited:  true,
```
wiggin77, line 492: @jespino @lieut-data is we what to ensure no ShareChannel records get created we need to disable AutoInvited as well.

PR #559, opened 2024-03-21
MM-57367: log when invite sent

calebroseland:
#### Summary
- Log when connection invite successfully sent. 
- Also adds `clean up activity handler` during `uninitialize` that was missing in #544.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57367

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -341,10 +341,10 @@ func (a *API) connect(w http.ResponseWriter, r *http.Request) {
 func (a *API) notifyConnect(w http.ResponseWriter, r *http.Request) {
 	userID := r.Header.Get("Mattermost-User-ID")
 
-	if inviteWasSent, err := a.p.MaybeSendInviteMessage(userID); inviteWasSent {
-		a.p.API.LogInfo("Successfully sent connection invite", "user_id", userID)
-	} else if err != nil {
+	if inviteWasSent, err := a.p.MaybeSendInviteMessage(userID); err != nil {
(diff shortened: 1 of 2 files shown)


server/api.go
```diff
@@ -341,7 +341,9 @@ func (a *API) connect(w http.ResponseWriter, r *http.Request) {
 func (a *API) notifyConnect(w http.ResponseWriter, r *http.Request) {
 	userID := r.Header.Get("Mattermost-User-ID")
 
-	if _, err := a.p.MaybeSendInviteMessage(userID); err != nil {
+	if inviteWasSent, err := a.p.MaybeSendInviteMessage(userID); inviteWasSent {
+		a.p.API.LogInfo("Successfully sent connection invite", "user_id", userID)
+	} else if err != nil {
 		a.p.API.LogWarn("Error in connection invite flow", "user_id", userID, "error", err.Error())
 	}
```
lieut-data, line 348: nit: should we test `err` first? It makes no semantic difference in this case, but in practice guards against weird issues like `return true, err` downstream.

```suggestion
	if inviteWasSent, err := a.p.MaybeSendInviteMessage(userID); err != nil {
		a.p.API.LogWarn("Error in connection invite flow", "user_id", userID, "error", err.Error())
	} else if inviteWasSent {
		a.p.API.LogInfo("Successfully sent connection invite", "user_id", userID)	
	}
```

lieut-data (approved):

sbishel (approved):

server/api.go
calebroseland: Absolutely, thanks for spotting

PR #560, opened 2024-03-22
Enabling shared channels behind a config flag

jespino:
This PR does multiple things. One of them is adding some "hacks" to make shared channels work without server related changes. That hacks are going to be removed as soon as we have the changes in the Mattermost server.

Other thing is to ensure the DMs/GMs that are not associated with MSTeams at start (and they should be associated), gets linked at start up.

Also it adds a configuration setting that allow us to Enable/Disable shared channels, falling back to plugin hooks.

Also, on start we enforce the sharing of already existing linked channels.


server/plugin.go
```diff
@@ -860,7 +905,7 @@ func (p *Plugin) updateMetrics() {
 	p.GetMetrics().ObserveLinkedChannels(stats.LinkedChannels)
 }
 
-func (p *Plugin) OnSharedChannelsPing(_ *model.RemoteCluster) bool {
+func (p *Plugin) OnSharedChannelsPing(rc *model.RemoteCluster) bool {
```
wiggin77, line 908: This will create linter warnings.

wiggin77 (approved):
LGTM :+1:

server/message_hooks.go
```diff
@@ -97,22 +170,26 @@ func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
...
 func (p *Plugin) ReactionHasBeenAdded(c *plugin.Context, reaction *model.Reaction) {
-	if !p.getConfiguration().SyncReactions {
+	if p.getConfiguration().UseSharedChannels {
 		return
 	}
+	p.reactionAddedHandler(reaction)
+}
 
-	updateRequired := true
-	if c.RequestId == "" {
```
lieut-data, line 111: What was the purpose of the `c.RequestId` check before?

server/message_hooks.go
```diff
@@ -670,8 +758,14 @@ func (p *Plugin) Delete(teamID, channelID string, user *model.User, post *model.
 	return nil
 }
 
-func (p *Plugin) DeleteChat(chatID string, user *model.User, post *model.Post) error {
-	client, err := p.GetClientForUser(user.Id)
+func (p *Plugin) DeleteChat(userID string, post *model.Post) error {
+	client, err := p.GetClientForUser(userID)
+	if err != nil {
+		p.handlePromptForConnection(userID, post.ChannelId)
```
lieut-data, line 764: Let's not prompt for connection on delete -- we're only doing it when sending for now, and will circle back with messaging in these other cases differently.
```suggestion
```

server/plugin.go
```diff
@@ -268,6 +270,27 @@ func (p *Plugin) start(isRestart bool) {
 		}
 	}
 
+	if p.getConfiguration().UseSharedChannels {
+		var ids []string
+		ids, err = p.store.ListDMsGMsToConnect()
```
lieut-data, line 275: This is an `N^2` query on first run in terms of number of users. Should we do this asynchronously? Should we paginate? Also, should we put this behind a mutex to avoid each plugin instance doing the same? (Maybe it's a job?)

server/plugin.go
```diff
@@ -513,6 +535,25 @@ func (p *Plugin) onActivate() error {
 			return dbErr
 		}
 
+		// simulate a ping to the plugin from server so it appears "online" immediately.
+		// !!!! remove this when fixed in server
+		_, err = db.Exec("update remoteclusters set lastpingat=$1 where remoteid=$2;", model.GetMillis(), p.remoteID)
+		if err != nil {
+			return fmt.Errorf("cannot simulate ping: %w", err)
```
lieut-data, line 542: Should this fail the plugin startup, or can we continue safely?

server/plugin.go
```diff
@@ -513,6 +535,25 @@ func (p *Plugin) onActivate() error {
...
+		}
+
+		// writing the invite directly in the DB
+		// !!!! remove this when fixed in server
+		p.inviteRemoteToChannel = func(channelID, remoteID, userID string) error {
```
lieut-data, line 547: Should this just be a method on the `plugin` instead of a possibly `nil` function pointer (depending on when we initialize)?

server/plugin.go
```diff
@@ -513,6 +535,25 @@ func (p *Plugin) onActivate() error {
...
+			now := model.GetMillis()
+			_, err2 := db.Exec("INSERT into sharedchannelremotes (id, channelid, creatorid, createat, updateat, isinviteaccepted, isinviteconfirmed, remoteid, lastpostupdateat, lastpostid, lastpostcreateat) values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11) ON CONFLICT (channelid, remoteid) DO NOTHING", model.NewId(), channelID, userID, now, now, true, true, remoteID, now, "", now)
+			if err2 != nil {
+				p.API.LogError("cannot simulate invite", "error", err2)
```
lieut-data, line 551: Should we log this as a warning, or will a human be asked to intervene when this log is emitted?

server/plugin.go
```diff
@@ -882,20 +927,22 @@ func (p *Plugin) OnSharedChannelsAttachmentSyncMsg(fi *model.FileInfo, _ *model.
 }
 
 func (p *Plugin) OnSharedChannelsSyncMsg(msg *model.SyncMsg, _ *model.RemoteCluster) (model.SyncResponse, error) {
-	now := model.GetMillis()
-
 	var resp model.SyncResponse
+	if !p.getConfiguration().UseSharedChannels {
+		return resp, nil
```
lieut-data, line 932: If we don't reply with `resp.PostsLastUpdateAt` set to the incoming value, does that mean all the shared channels are effectively "queued up" until we turn on shared channels? And at that point we'll replay potentially weeks of messages through the plugin and over to MS Teams, possibly risking duplicate messages?

server/plugin.go
```diff
@@ -923,3 +971,21 @@ func (p *Plugin) OnSharedChannelsSyncMsg(msg *model.SyncMsg, _ *model.RemoteClus
 
 	return resp, nil
 }
+
+func (p *Plugin) ChannelHasBeenCreated(c *plugin.Context, channel *model.Channel) {
+	_ = p.updateAutomutingOnChannelCreated(channel)
+
+	if p.getConfiguration().UseSharedChannels && (channel.Type == model.ChannelTypeDirect || channel.Type == model.ChannelTypeGroup) {
```
lieut-data, line 978: Let's use `channel.IsGroupOrDirect()` for clarity here.

server/plugin.go
```diff
@@ -923,3 +971,21 @@ func (p *Plugin) OnSharedChannelsSyncMsg(msg *model.SyncMsg, _ *model.RemoteClus
 
 	return resp, nil
 }
+
+func (p *Plugin) ChannelHasBeenCreated(c *plugin.Context, channel *model.Channel) {
```
lieut-data, line 975: Given we're still relying on hooks to make shared channels work (for now), I'm guessing that in the event of a downtime incident, we'll want to plan to rotate the plugin to trigger the bulk share channel calls on startup?

server/testutils/containere2e/containere2e.go
```diff
@@ -165,3 +171,28 @@ func NewE2ETestPlugin(t *testing.T, extraOptions ...mmcontainer.MattermostCustom
...
+					{"ID": 5, "Name": "debug", "Stacktrace": false},
+					{"ID": 4, "Name": "info", "Stacktrace": false},
+					{"ID": 3, "Name": "warn", "Stacktrace": false},
+					{"ID": 2, "Name": "error", "Stacktrace": false},
+					{"ID": 1, "Name": "fatal", "Stacktrace": true},
+					{"ID": 0, "Name": "panic", "Stacktrace": true},
+					{"ID": 130, "Name": "RemoteClusterServiceDebug", "Stacktrace": false},
+					{"ID": 131, "Name": "RemoteClusterServiceError", "Stacktrace": true},
+					{"ID": 200, "Name": "SharedChannelServiceDebug", "Stacktrace": false},
+					{"ID": 201, "Name": "SharedChannelServiceError", "Stacktrace": true}
```
lieut-data, line 190: Should we prepare guidance for SRE to enable these logs in test/production?

lieut-data (changes requested):
Thanks, @jespino! A handful of minor suggestions, but also a blocking concern around the `O(N^2)` startup time required in terms of number of users and plugin contention.

server/message_hooks.go
```diff
@@ -36,7 +36,80 @@ func (p *Plugin) UserWillLogIn(_ *plugin.Context, user *model.User) string {
...
+	p.messageDeletedHandler(post)
+}
+
+func (p *Plugin) messageDeletedHandler(post *model.Post) {
+	channel, appErr := p.API.GetChannel(post.ChannelId)
+	if appErr != nil {
+		return
+	}
+
+	isDirectOrGroupMessage := channel.Type == model.ChannelTypeDirect || channel.Type == model.ChannelTypeGroup
```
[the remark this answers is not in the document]
enahum: Also move to line 63, no need to compute if we return early

server/plugin.go
enahum: How does it work when a new dm/gm is created ?

enahum (commented):
Some comments but more importantly.. what if shared channels is disabled at the server level but useSharedChannrls is set to true?

server/command.go
```diff
@@ -227,20 +227,22 @@ func (p *Plugin) executeLinkCommand(args *model.CommandArgs, parameters []string
 		return p.cmdError(args.UserId, args.ChannelId, "Error occurred while saving the subscription")
 	}
 
-	if !p.getConfiguration().DisableSyncMsg {
+	if !p.getConfiguration().DisableSyncMsg && p.getConfiguration().UseSharedChannels {
```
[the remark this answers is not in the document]
jespino: They are two different things, AFAIK

server/message_hooks.go
jespino: It was used to differenciate requests that comes from a user interaction through the api or an internal call. It is not longer differenciable so I'm checking always the IgnorePluginHooksMap. Anyway, this is possible that it is not needed for shared channels, we can verify that later and check if we can remove it atfer removing the hooks.

server/message_hooks.go
jespino: Ok, I still think can be missleading that you delete a message here and you don't know that is not really deleted from the other side.

server/plugin.go
jespino: I guess we could continue safely, but if this is failing, almost 100% sure is because something is really wrong the the DB connection, so I don't see the point on keep going.

server/plugin.go
jespino: Good call, it should be a warning.

server/plugin.go
jespino: OK

server/plugin.go
jespino: That is exactly what we are doing in the migration that we created.

server/testutils/containere2e/containere2e.go
jespino: This was introduced by @wiggin77 in the ce2e tests and I found them very useful for us in that use case, I don't know what do you think about enable that logs in the cloud @wiggin77?

server/message_hooks.go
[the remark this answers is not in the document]
jespino: Done

server/message_hooks.go
jespino: Done

server/plugin.go

server/plugin.go
lieut-data: Right, I'm referring to steps we would need to manually take if an incident occurred, specifically that we'd want to re-run that migration by ensuring we rotate a pod and get the plugin to start from scratch?

server/store/sqlstore/store.go
```diff
@@ -936,8 +936,8 @@ func (s *SQLStore) IsUserPresentInWhitelist(userID string) (bool, error) {
 	return result != "", nil
 }
 
-func (s *SQLStore) ListDMsGMsToConnect() ([]string, error) {
-	query := s.getQueryBuilder().Select("c.Id").From("Channels AS c").LeftJoin("sharedchannelremotes AS scr ON scr.channelid = c.id").Where(sq.Eq{"scr.remoteid": nil}).Where(sq.Or{sq.Eq{"c.Type": "G"}, sq.Eq{"c.Type": "D"}})
+func (s *SQLStore) ListDMsGMsToConnectBatch() ([]string, error) {
+	query := s.getQueryBuilder().Select("c.Id").From("Channels AS c").LeftJoin("sharedchannelremotes AS scr ON scr.channelid = c.id").Where(sq.Eq{"scr.remoteid": nil}).Where(sq.Or{sq.Eq{"c.Type": "G"}, sq.Eq{"c.Type": "D"}}).Limit(500)
```
lieut-data, line 940: To confirm, this query doesn't use our own `p.remoteId` at all: is that because a shared channel is shared "once" for all remotes? What does the `scr.remoteid` mean in that case?

server/plugin.go
```diff
@@ -357,6 +337,43 @@ func (p *Plugin) start(isRestart bool) {
...
+				ShareName: id,
+			}); err != nil {
+				p.API.LogWarn("Unable to share channel", "channel_id", id, "error", err.Error())
+			}
+			if err = p.inviteRemoteToChannel(id, p.remoteID, p.userID); err != nil {
+				p.API.LogWarn("Unable simulate the invite remote channel", "channel_id", id, "error", err.Error())
+			}
+		}
+		// Give some time to other work to happen in the server
+		time.Sleep(1 * time.Second)
```
lieut-data, line 375: On a server with 10,000 users, there's theoretically up to 100,000,000 direct message channels alone. Even if we assume only a 5% connected graph, that's still `5,000,000` channels. With a batch size of `500`, this will still take 10,000 seconds or or almost 3 hours.

This is fine for our bug bash, but let's not merge this.

lieut-data (changes requested):
I think this is fine for our bug bash, but just highlighting a few things to discuss.

server/plugin.go
wiggin77: Ideally if we go from shared channels off to on, we should set the cursors to current.  That will ensure we don't replay history.  Alternatively we can delete all the SharedChannelRemotes records which contain the cursors when shared channels is turned off, via UninviteRemoteToChannel.

plugin.json
```diff
@@ -146,6 +146,12 @@
             "type": "bool",
             "help_text": "When true, the plugin will not enable any sync msg infrastructure.",
             "default": false
+        },{
+            "key": "useSharedChannels",
```
wiggin77, line 150: @jespino @lieut-data we should consider renaming this to something like "Reliable Message Delivery".  "Shared Channels" doesn't have much meaning in the MS Teams context, and is potentially confusing.

jespino: This is a temporary setting, but yes, we can change it.

server/store/sqlstore/store.go
jespino: Good catch, this works because we don't have DMs connected to any other shared channel thing, so I was checking if there is not remoteId, what means, the channel is not connected, I think the question is more "remoteId != p.remoteId", let me change that.

server/message_hooks.go
```diff
@@ -40,111 +40,110 @@ func (p *Plugin) MessageHasBeenDeleted(_ *plugin.Context, post *model.Post) {
 	if p.getConfiguration().UseSharedChannels {
 		return
 	}
-	p.messageDeletedHandler(post)
+	_ = p.messageDeletedHandler(post)
```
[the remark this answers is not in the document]
jespino: For me was ideal to introduce it here because I did want to give also that support for shared channels, we can reintroduce it later if the PR is reverted, not a big deal, IMO.

server/plugin.go
jespino: Every restart of the plugin, including configuration changes, are going to try to run the migration again.

jespino:
> Some comments but more importantly.. what if shared channels is disabled at the server level but useSharedChannrls is set to true?

In that case, the plugin doesn't work. The idea is that in the future `useSharedChannels` is going to be the only option (it is not going to be in the config file), so you will need to have shared channels enabled to make this plugin work.

server/plugin.go
jespino: Should this something done by the plugin? Looks like a Shared Channels responability. If I disable shared channels during 1 month, and enable it again after that, should it sync the whole month, or since the re-enabling? For me, shared channels, whenever is disabled/enabled should handle that cases.

server/plugin.go
jespino: The DM/GM creations are handle by a hook.

server/plugin.go
lieut-data: Since this particular configuration is in the plugin, I think this particular decision has to be owned by the plugin. It's perhaps unclear what should happen if the overall shared channels functionality is disabled at the server level.

server/plugin.go
lieut-data: > can be 3 hours, can be 10 hours, it doesn't matter that much

This is fine for testing, but we absolutely can't ship something with this much variability to production. I'd propose we wait for v9.7 and shared channels doing this natively (and efficiently) instead before considering this for production.

lieut-data:
@jespino, heads up I've extracted the portion of this PR that adds support for deleting messages from Mattermost over at https://github.com/mattermost/mattermost-plugin-msteams/pull/569. This will allow us to ship that feature while we continue to develop confidence in this PR and shared channels overall.

server/plugin.go
jespino: As far as I understand, this can be completely unneeded in the future with some shared channels code changes.

plugin.json
jespino: Works, for me, any name suggestion?

jespino:
@lieut-data sounds good to me moving that to another PR.

server/testutils/containere2e/containere2e.go
wiggin77: They can be helpful.  I can see them not wanting *ServiceDebug but *ServiceError is a good idea.  FWIW, these are enabled for community.

server/plugin.go
wiggin77: https://github.com/mattermost/mattermost/pull/26753 is merged.   It will share any DM/GM not previously shared on next post.

server/plugin.go
wiggin77: This is no longer needed.  Covered by https://github.com/mattermost/mattermost/pull/26753

server/plugin.go
wiggin77: This is no longer needed.  Handled by https://github.com/mattermost/mattermost/pull/26753

lieut-data:
Closing this PR for now as we're slimming down the plugin to reach code complete for security review. Let's keep the branch around though!

PR #563, opened 2024-03-28
MM-5701: reenable reattached plugin tests

lieut-data:
#### Summary
At a high level, this PR begins our move move away from mocks for the pluginapi -- that means no more fiddling with making tests pass only to realize the mock assertions were broken to begin with. All the rewritten tests now use a real pluginapi and a real store backed by a real Mattermost server and a real Postgres instance. We use @jespino's newly added `testcontainers` to host the Postgres instance, but run the server in memory meaning we're not dependent on a particular Docker container build. 

(@jespino, I wonder if we should migrate our `ce2e/` tests in this direction as this provides greater flexibility when testing plugin + server changes. As these have been very flaky lately, I've temporarily disabled them as part of this PR.)

This PR depends the plugin reattach model introduced with https://github.com/mattermost/mattermost/pull/26421, but we don't actually need it to be merged right away, as we can point directly at the commit via `go.mod`. (Naturally, we'll clean this up once merged!)

This is "part 1" of what I hope to be a series of PRs that evict almost all mocking from our unit tests. Of course, we still mock for the MS Teams Graph API of course, but even that's improved because we no longer `panic` when the test fails: the errors now get emitted through the `testing.T` normal interface.

It may be helpful to review the individual commits. I briefly explored putting a bunch of the `helper_*` code directly into our `public` package in the monorepo for immediate reuse, but there's some dependencies on `server/v8` that make this impossible, and I also wanted some time to soak these changes in practice before promoting them more widely.

#### Ticket Link
Relates-to: https://mattermost.atlassian.net/browse/MM-56022

#### Important Notes
This change requires go 1.21 as a result of the updated server dependency. I'll sort out any downstream impacts of this change as we discover them (e.g. if the plugin tooling needs to be updated or not).


lieut-data:
@cpoile, @fmartingr, @crspeller, @hanzei & @mickmister, same as https://github.com/mattermost/mattermost/pull/26421, I welcome all feedback, but don't actually expect everyone to review. Feel free to remove yourself or review as you see fit -- mostly wanting to maximize the success of this effort.

server/ce2e/iframe_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build exclude
```
lieut-data, line 1: Temporarily disabled due to flakiness. Will discuss with @jespino, but maybe there's a way we can consolidate our testing approaches.

server/command.go
```diff
@@ -33,13 +33,18 @@ func (p *Plugin) createCommand(syncLinkedChannels bool) *model.Command {
...
-	})
-	return &model.CommandResponse{}, nil
+func (p *Plugin) cmdSuccess(text string) (*model.CommandResponse, *model.AppError) {
+	return &model.CommandResponse{
+		ResponseType: model.CommandResponseTypeEphemeral,
+		Text:         text,
+	}, nil
+}
+
+func (p *Plugin) cmdError(detailedError string) (*model.CommandResponse, *model.AppError) {
```
lieut-data, line 43: In the happy path, we don't need to send our own ephemeral messages but can have the server do that -- plus, the unit tests are simpler when we can simply test the returned `model.CommandResponse` in most cases. (In others, where we still need ephemeral messages, I do go to the effort of monitoring the websocket directly.)

server/command_test.go
```diff
@@ -1,1088 +1,916 @@
...
+
+func assertCommandResponse(t *testing.T, text string, actual *model.CommandResponse) {
+	require.NotNil(t, actual)
+	require.Equal(t, &model.CommandResponse{
+		ResponseType: model.CommandResponseTypeEphemeral,
+		Text:         text,
+	}, actual)
+}
+
+func assertEphemeralMessage(t *testing.T, websocketClient *model.WebSocketClient, channelID, message string) {
```
lieut-data, line 33: There's opportunities to share this code in the `public` submodule and make these patterns more reusable. I dream of a day plugins can import all the helper functions they need and focus on just the business semantics of the test.

server/command_test.go
```diff
@@ -1,1088 +1,916 @@
...
+	team := th.SetupTeam(t)
+	user1 := th.SetupUser(t, team)
+	user2 := th.SetupUser(t, team)
+	user3 := th.SetupUser(t, team)
+
+	t.Run("invalid channel", func(t *testing.T) {
+		th.p.configuration.DisableSyncMsg = true
+		commandResponse, appErr := th.p.executeUnlinkCommand(&model.CommandArgs{
+			UserId:    user1.Id,
+			ChannelId: model.NewId(),
```
lieut-data, line 71: I hope this feels familiar: it's basically what we've been doing in the server forever. Setup a server instance (in this case, a plugin instance: the server + database is a singleton for the package), and just test things directly. No mocking required 🎉.

server/main_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	// Export the site url for later use in tests.
+	os.Setenv("MM_SERVICESETTINGS_SITEURL", fmt.Sprintf("http://localhost:%v", server.ListenAddr.Port))
+
+	return nil
+}
+
+// TestMain is run before any tests within this package and helps setup a singleton Postgres and
+// Mattermost intance for use with tests.
+func TestMain(m *testing.M) {
```
lieut-data, line 195: I opted for a singleton Mattermost + Postgres instance instead of spinning these up and down for each test. It's more interesting to start and stop the plugin as we need to (which is lightning fast).

server/plugin.go
```diff
@@ -312,20 +313,22 @@ func (p *Plugin) start(isRestart bool) {
...
+			p.API.LogError("error in scheduling the check credentials job", "error", jobErr)
+			return
+		}
+		p.checkCredentialsJob = checkCredentialsJob
 
-	// Run the job above right away so we immediately populate metrics.
-	p.checkCredentials()
+		// Run the job above right away so we immediately populate metrics.
+		p.checkCredentials()
+	}
```
lieut-data, line 331: Allow the tests not to run this job and trigger unexpected mock calls to the MS Teams Graph API. There's an opportunity to improve things here in the future.

server/plugin.go
```diff
@@ -479,7 +489,7 @@ func (p *Plugin) onActivate() error {
 		Username:    botUsername,
 		DisplayName: botDisplayName,
 		Description: "Created by the MS Teams Sync plugin.",
-	}, pluginapi.ProfileImagePath("assets/icon.png"))
+	}, pluginapi.ProfileImageBytes(assets.Icon))
```
lieut-data, line 492: `ProfileImagePath` had assumptions about the filesystem that were complicated by the reattach -- moving to embed and passing the bytes solves this.

server/plugin_test.go
```diff
@@ -68,6 +69,9 @@ func newTestPlugin(t *testing.T) *Plugin {
 	plugin.API.(*plugintest.API).On("Conn", true).Return("connection-id", nil)
 	plugin.API.(*plugintest.API).On("GetUnsanitizedConfig").Return(&config)
 	plugin.API.(*plugintest.API).On("EnsureBotUser", bot).Return("bot-user-id", nil).Times(1)
+	plugin.API.(*plugintest.API).On("SetProfileImage", mock.AnythingOfType("string"), mock.AnythingOfType("[]uint8")).Return(nil).Times(1)
+	plugin.API.(*plugintest.API).On("RegisterPluginForSharedChannels", mock.Anything).Return("remote-id", nil).Times(1)
+	plugin.API.(*plugintest.API).On("UnregisterPluginForSharedChannels", mock.Anything).Return(nil).Times(1)
```
lieut-data, line 74: Make some of the existing tests "happy" until we rewrite those too.

server/selective_sync_test.go
```diff
@@ -1,287 +1,201 @@
...
-	"time"
 
-	goPlugin "github.com/hashicorp/go-plugin"
 	"github.com/mattermost/mattermost/server/public/model"
-	"github.com/mattermost/mattermost/server/public/plugin"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func setupPlugin(t *testing.T) *Plugin {
```
lieut-data, line 19: Refactored to be reusable in more than one test.

lieut-data:
Note that a full suite execution without caching is still tolerably fast -- the overhead of reattaching plugins is negligible. It is slower, but mostly because there's a real `pluginapi` and database doing work now, so the tradeoff to make the tests "real" is worth it.

CleanShot 2024-03-28 at 09 53 21@2x

server/command.go
lieut-data: I ended up reverting this part to continue to send ephemeral messages from the bot. I realized we can't reply with a non-system user, so the experience changed. I also discovered that we needed some extra helper functionality to monitor for websocket events to ensure correct behaviour.

server/command_test.go
```diff
@@ -1,1088 +1,951 @@
...
+	require.NotNil(t, actual)
+	require.Equal(t, &model.CommandResponse{}, actual)
+}
+
+func assertEphemeralResponse(th *testHelper, t *testing.T, args *model.CommandArgs, message string) {
+	t.Helper()
+
+	th.assertEphemeralMessage(t, args.UserId, args.ChannelId, message)
+}
+
```
hanzei, line 31: I would love to see these helper methods in a shared repo, potentially https://github.com/mattermost/mattermost/tree/master/server/public/pluginapi, so other plugins can reuse them. Same goes for the `testHelper`.

hanzei (commented):
I don't know the code base well, so my contribution here is limited. One suggestion below regarding reusing some of the test code.

server/command_test.go
hanzei: This is awesome :tada:

server/main_test.go
```diff
@@ -0,0 +1,218 @@
...
+// Done calls all cleanup functions with defer-like semantics (last function added called first).
+func (mt *mainTest) Done() {
+	for i := range mt.cleanupFunctions {
+		f := mt.cleanupFunctions[len(mt.cleanupFunctions)-i-1]
+		f()
+	}
+}
+
+// setupServerPath defines MM_SERVER_PATH correctly for server initialization. One day this won't
+// be necessary at all.
```
hanzei, line 45: One can only dream...

server/main_test.go
```diff
@@ -0,0 +1,218 @@
...
+	mt.cleanupFunctions = append(mt.cleanupFunctions, f)
+}
+
+// Done calls all cleanup functions with defer-like semantics (last function added called first).
+func (mt *mainTest) Done() {
+	for i := range mt.cleanupFunctions {
+		f := mt.cleanupFunctions[len(mt.cleanupFunctions)-i-1]
+		f()
+	}
+}
```
hanzei, line 42: Neat :+1:

server/main_test.go
hanzei: Let's see how many tests break due to missing cleanup in other tests.

hanzei: One downside of using `testing.M` is that it's run for every test, even if the tests don't need it, which can slow down development work.

How much does running `TestMain` in each test add to the test duration? I.e. is it worth adding the code from `TestMain` to `setupTestHelper`?

lieut-data: I like this! We could certainly make this a `sync.Once`-style implementation instead of `TestMain`, then any tests that don't use this won't incur the cost, while we still get memoization. Let me explore this, even though I expect 99% of tests to transition to this model at least in this repository, as it's also tidier.

server/command_test.go
lieut-data: Yes, 100%. I wanted to let these soak a bit, but also there's the problem that the `public/` module can't reimport the `server/v8` module that's currently required to spin up the server. Need a larger solution (which I intend to eventually tackle!)

server/main_test.go
lieut-data: I **will** get back to https://github.com/mattermost/mattermost/pull/24968 one day :)

server/main_test.go
```diff
@@ -0,0 +1,218 @@
...
+	_, appErr := ap.CreateUser(request.EmptyContext(testLogger), user)
+	if appErr != nil {
+		return appErr
+	}
+
+	// Export the site url for later use in tests.
+	os.Setenv("MM_SERVICESETTINGS_SITEURL", fmt.Sprintf("http://localhost:%v", server.ListenAddr.Port))
+
+	return nil
+}
```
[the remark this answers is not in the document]
lieut-data: In the long run, I think we'll want both options, as I agree running against a particular server version is super powerful. But for now I think the defaults of the `go.mod` version is way more powerful, since with a simple `go mod replace`, I can make server-side changes and test them immediately with the plugin (e.g. think shared channels iteration). It's way more painful to force a Docker image.

```sh
go mod edit -replace=github.com/mattermost/mattermost/server/public=../../platform/server/public
go mod edit -replace=github.com/mattermost/mattermost/server/v8=../../platform/server
```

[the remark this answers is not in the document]
hanzei: One downside of using `go.mod` is that any breaking changes in `public` (and there have been ones) will hinder using an older version for tests.

This is surely not a blocker for this PR, but I wonder if we would benefit from decoupling the MM version the plugin is compiled against from the MM version the plugin runs against. (This is also how a "real" manual test would work!)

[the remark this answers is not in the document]
lieut-data: Agreed, and this is where the Docker approach could work well. It should be an easy extension to what we have now.

server/main_test.go
lieut-data: >  and I'm not sure it is worth it to have the tests run faster.

After mocked tests, slow tests are my next biggest impediment. Mocked tests are lightning fast, but often meaningless. E2E tests are molasses slow, but arguably most correct. The sweet spot is something that approximates E2E with the speed of mocked tests which is where we land with some of these tradeoffs. In this case, clearing the database is an order of magnitude faster than a fresh database.

The test helper paradigm here (and in the server) automates this for the most part, with each new `Test*` getting a fresh set of tables related to the plugin. We spin up new teams, channels and users all over the place without clearing them between tests, but I haven't yet found this to be problematic (unless one starts hard-coding usernames or something).

I do need to fix one thing in the `testHelper` setup to ensure a fresh instance also clears the database, but I'm happy to own this going forward if we find this more painful than I predict.

wiggin77: If the tables are cleared automatically I have no issue.

hanzei (approved):
I love the speed of the test due to re-use of the DB. Let's see if this approach proves to be robust.

server/ce2e/iframe_test.go
wiggin77: We should probably move the embedded feature into its own plugin.

server/ce2e/plugin_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build exclude
+
```
wiggin77, line 2: Are we disabling all the e2e tests that Jesus built?  If they are flaky I would rather take a crack at fixing them versus disabling then completely.

wiggin77: The failing unit tests aren't flaky, but were broken by two PRs last week.  I have fixed one, and the other will take some more work as an index was added that now causes ID collisions in the tests.  I will modify the tests to use unique IDs.

wiggin77: It raises the question how the PRs were merged if the tests were failing.

server/main_test.go
```diff
@@ -0,0 +1,230 @@
...
+	os.Setenv("MM_SERVER_PATH", strings.Trim(strings.TrimSpace(string(serverPath)), "'"))
+
+	return nil
+}
+
+// setupDatabase initializes a singleton Postgres testcontainer and mattermost_test database for
+// use with tests.
+func setupDatabase(mt *mainT) error {
+	// Setup a Postgres testcontainer for all tests.
+	pgContainer, err := postgres.RunContainer(context.TODO(),
```
jespino, line 63: I love this :)

jespino (changes requested):
My only concern with this is the fact that we are disabling the e2e testing when I don't think is needed.

lieut-data:
Thanks, @jespino! Disabling those tests definitely itself isn't needed as part of this PR: could I ask you to push a patch that removes those build flags?

And could you help own the flakiness of those tests in CI? These are ones I've had to re-run repeatedly over the last few weeks, slowing down development. Disabling was a coarse solution, but if we can fix them that would be better :)

jespino:
@lieut-data no problem, I'll remove that, about making the tests pass, I can take care of it too.

jespino:
@lieut-data tests are working and main branch is merged :)

jespino (approved):
LGTM

lieut-data:
Thank you, @jespino! I really appreciate that :)

PR #564, opened 2024-03-28
[MM-56982] UNIQUE index on msteamsuserid 

JulienTant:
#### Summary

- Add UNIQUE index on our users table to ensure msteamsuserid is unique.
- Created a migration to ensure we only have one msteamsuserid before apply the index
- Move migrations related methods to the migration file.

This PR does not include any check on the constraint on purpose: this will help us find where and why this happens in the first place.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56982

### DIFF
--- a/server/store/sqlstore/helper_test.go
+++ b/server/store/sqlstore/helper_test.go
@@ -83,13 +83,13 @@ func setupTestStore(t *testing.T) (*SQLStore, *plugintest.API) {
 	store.api = api
 	store.db = db
 
-	err := store.Init("")
-	require.NoError(t, err)
-	err = store.createTable("Teams", "Id VARCHAR(255), DisplayName VARCHAR(255)")
+	err := store.createTable("Teams", "Id VARCHAR(255), DisplayName VARCHAR(255)")
 	require.NoError(t, err)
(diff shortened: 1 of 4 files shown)


sbishel (commented):
Looks good. One question. 
When we delete synthetic users from `msteamssync_users` table should we also delete those users from the `users` table?

JulienTant:
@sbishel Unless I'm missing something, there's no case where we delete anything from `msteamssync_users` or `users`. The promote operation updates the synthetic user fields to turn them into normal user (=remove RemoteID); I've seen a few tickets talking about "restoring" the synthetic user, but it we were to do that, it would mean un-promoting a normal user (restoring RemoteID)

JulienTant:
@sbishel ohhh I thought we were on another ticket 😓 

I don't have a strong opinion on that; I would need to refactor my PR to get a list of the synthetic user IDs to remove,  when right now I do the reverse: I keep a list of IDs to keep.

sbishel (approved):

mgdelacroix (approved):

mgdelacroix (approved):

PR #567, opened 2024-03-30
MM-57534: monitor message creation delay

lieut-data:
#### Summary
Monitor the delta between when a message is created on one platform and when it's recorded as such on the other. We acknowledge that without synchronized clocks between Mattermost and Teams, the absolute value here isn't necessarily accurate. But the idea is to measure the variability in the delay between messages flowing in each direction.

As part of this change, drop the unused "confirmed message" metric.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-57534

### DIFF
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -276,7 +276,6 @@ func (ah *ActivityHandler) handleCreatedActivity(msg *clientmodels.Message, subs
 	// Avoid possible duplication
 	postInfo, _ := ah.plugin.GetStore().GetPostInfoByMSTeamsID(msg.ChatID+msg.ChannelID, msg.ID)
 	if postInfo != nil {
-		ah.plugin.GetMetrics().ObserveConfirmedMessage(metrics.ActionSourceMattermost, isDirectMessage)
 		return metrics.DiscardedReasonDuplicatedPost
 	}
 
--- a/server/handlers/handlers.go
+++ b/server/handlers/handlers.go
@@ -371,6 +371,7 @@ func (ah *ActivityHandler) handleCreatedActivity(msg *clientmodels.Message, subs
 	}
 
 	ah.plugin.GetMetrics().ObserveMessage(metrics.ActionCreated, metrics.ActionSourceMSTeams, isDirectMessage)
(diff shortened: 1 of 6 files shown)


server/handlers/handlers.go
```diff
@@ -372,6 +371,7 @@ func (ah *ActivityHandler) handleCreatedActivity(msg *clientmodels.Message, subs
 	}
 
 	ah.plugin.GetMetrics().ObserveMessage(metrics.ActionCreated, metrics.ActionSourceMSTeams, isDirectMessage)
+	ah.plugin.GetMetrics().ObserveMessageDelay(metrics.ActionCreated, metrics.ActionSourceMSTeams, isDirectMessage, time.Since(msg.CreateAt))
```
lieut-data, line 374: We're measuring the time between when we received the message (`time.Now()` implicit via `time.Since()`) and the reported time of the message creation from Teams.

server/message_hooks.go
```diff
@@ -541,6 +542,8 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post, c
 	}
 
 	p.GetMetrics().ObserveMessage(metrics.ActionCreated, metrics.ActionSourceMattermost, true)
+	p.GetMetrics().ObserveMessageDelay(metrics.ActionCreated, metrics.ActionSourceMattermost, true, newMessage.CreateAt.Sub(time.UnixMilli(post.CreateAt)))
```
lieut-data, line 545: Here, we're measuring the time between when we Teams received the message (`newMessage.CreateAt`, returned via the API) and the time we created the post in Mattermost.

server/message_hooks.go
```diff
@@ -625,6 +628,7 @@ func (p *Plugin) Send(teamID, channelID string, user *model.User, post *model.Po
 	}
 
 	p.GetMetrics().ObserveMessage(metrics.ActionCreated, metrics.ActionSourceMattermost, false)
+	p.GetMetrics().ObserveMessageDelay(metrics.ActionCreated, metrics.ActionSourceMattermost, false, newMessage.CreateAt.Sub(time.UnixMilli(post.CreateAt)))
```
lieut-data, line 631: Same as above, except for non-chats. (We really need to consolidate these two methods!)

JulienTant (approved):

sbishel (approved):

PR #569, opened 2024-03-30
MM-56809: support deleting messages from Mattermost

lieut-data:
#### Summary
Extract the portion of https://github.com/mattermost/mattermost-plugin-msteams/pull/560 that adds support for deleting messages from Mattermost and soft deleting those in Teams.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-56809

### DIFF
--- a/server/message_hooks.go
+++ b/server/message_hooks.go
@@ -68,7 +68,7 @@ func (p *Plugin) MessageHasBeenDeleted(_ *plugin.Context, post *model.Post) {
 			}
 		}
 
-		if err := p.DeleteChat(post.UserId, post); err != nil {
+		if err := p.DeleteChat(post); err != nil {
 			p.API.LogWarn("Unable to delete chat", "error", err.Error())
 			return
 		}
@@ -724,8 +724,8 @@ func (p *Plugin) Delete(teamID, channelID string, user *model.User, post *model.
 	return nil
 }
 
-func (p *Plugin) DeleteChat(userID string, post *model.Post) error {
-	client, err := p.GetClientForUser(userID)
+func (p *Plugin) DeleteChat(post *model.Post) error {
+	client, err := p.GetClientForUser(post.UserId)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


server/message_hooks.go
```diff
@@ -670,8 +724,13 @@ func (p *Plugin) Delete(teamID, channelID string, user *model.User, post *model.
 	return nil
 }
 
-func (p *Plugin) DeleteChat(chatID string, user *model.User, post *model.Post) error {
-	client, err := p.GetClientForUser(user.Id)
+func (p *Plugin) DeleteChat(userID string, post *model.Post) error {
```
enahum, line 727: Are we expecting that the user ID will be other than the post author? If not, I think with only with the post as argument should be enough

enahum (approved):

server/message_hooks.go
lieut-data: Good question: @jespino, do you know if there are cases where we `DeleteChat` for a different user?

jespino: A system admin can delete other user post, but not only that, in theory, the permissions system allows that anybody that has permission to delete others posts in a channel, can deleted that posts, and that can be configured with the advanced permissions system.

wiggin77 (approved):
LGTM :+1:

server/message_hooks.go
lieut-data: In those cases, would the `userID` passed as a parameter be different from `post.UserId`? It seems redundant at present, so I'm guessing even for the system admin cases we could just pass `post`.

jespino: Yes, maybe it is possible that we can use the post.UserId, because we are using it to decide in behave of who we delete the post. I think it is safe enough to use the post.UserId directly.

lieut-data:
@lindy65, should we review this one pre-merge or at the bug bash?

lindy65 (approved):
Let's test this at the bug bash @lieut-data. I've QA-deferred this PR and noted to test it during the bug bash.

PR #582, opened 2024-04-02
[MM-57568] Synthetic user created from DM/GM have wrong RemoteID

JulienTant:
#### Summary
When creating a synthetic user from a DM/GM, we set a remoteID based on the user UUID instead of using the plugin remoteID.

Example of how it's done for the syncUsers job: https://github.com/mattermost/mattermost-plugin-msteams/blob/main/server/plugin.go#L727

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57568

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,6 @@ require (
 	github.com/microsoft/kiota-http-go v1.3.1
 	github.com/microsoftgraph/msgraph-sdk-go v1.36.0
 	github.com/microsoftgraph/msgraph-sdk-go-core v1.1.0
-	github.com/pborman/uuid v1.2.1
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.19.0
 	github.com/sirupsen/logrus v1.9.3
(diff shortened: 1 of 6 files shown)


lieut-data (approved):
Nice catch, @JulienTant! @jespino, will we need to re-migrate anything here?

JulienTant:
@lieut-data we do have this migration that runs everytime:

```
func (s *SQLStore) runMigrationRemoteID(remoteID string) error {
	_, err := s.getQueryBuilder().Update("Users").Set("RemoteID", remoteID).Where(sq.And{
		sq.NotEq{"RemoteID": nil},
		sq.NotEq{"RemoteID": ""},
		sq.Expr("RemoteID NOT IN (SELECT remoteid FROM remoteclusters)"),
		sq.Like{"Username": "msteams_%"},
	}).Exec()
	return err
}
```

So anyone with a `msteams_` name, that has a RemoteID AND this RemoteID is NOT part of `remoteclusters` will be assigned the plugin RemoteID. So in theory, every restart of the plugin """fixes""" the existing users.

sbishel (approved):
LGTM

PR #583, opened 2024-04-02
MM-57563 - update method to allow self posts to sync, only if connected

sbishel:
#### Summary
Changes how we make sure self DMs sync, rather then setting the variable `chatMembersSpanPlatforms`. 
Skip the "Selective Sync" check altogether.

Most of these changes are due to the change of function name from `ChatSpansPlatforms` to `ChatShouldSync`. The functions that called it actually set variable names of `shouldSync` and with the check for single user, makes more sense.

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-57563


server/selective_sync.go
```diff
@@ -6,25 +6,24 @@ import (
...
 }
 
 // ChatMembersSpanPlatforms determines if the given channel members span both Mattermost and
 // MS Teams. Chats between users on the same platform are skipped if selective sync is enabled.
-// Chats with only a single member are self chats and always sync.
 func (p *Plugin) ChatMembersSpanPlatforms(members model.ChannelMembers) (bool, *model.AppError) {
-	if len(members) == 1 {
-		return true, nil
-	}
-
```
sbishel, line 27: I considered throwing an error here, if only one member is passed. As this function really shouldn't be called for a single user...based on name. Thoughts?

server/message_hooks.go
```diff
@@ -119,14 +119,18 @@ func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
...
 		}
 
-		chatMembersSpanPlatforms, appErr := p.ChatMembersSpanPlatforms(members)
-		if appErr != nil {
-			p.API.LogWarn("Failed to check if chat members span platforms", "error", appErr.Error(), "post_id", post.Id, "channel_id", post.ChannelId)
-			return
-		}
+		isSelfPost := len(members) == 1
+		chatMembersSpanPlatforms := false
+		if !isSelfPost {
```
sbishel, line 124: If selfPost...allow to sync, setting `chatMembersSpanPlatforms := false`. This variable is passed to `SendChat` below. `SendChat` uses this variable to determine whether or not to display the "connect" message if the user is not connected.

sbishel:
I need to add a couple unit tests...but ready to take a look if we want to try for v1.11.0.

server/selective_sync.go
enahum: I thunk it makes sense to return an error in that case

enahum (approved):

server/selective_sync.go
sbishel: Added error.

server/selective_sync_test.go
```diff
@@ -8,19 +8,33 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestChatSpansPlatforms(t *testing.T) {
+func TestChatShouldSync(t *testing.T) {
+	t.Skip("pending upstream changes: https://mattermost.atlassian.net/browse/MM-57018")
```
wiggin77, line 12: The upstream changes have been merged.

wiggin77: I tested the unit test after removing "skip" and it looks good.

wiggin77 (approved):
LGTM :+1:    One unit test should be re-enabled.

mgdelacroix (approved):
LGTM! :+1:

sbishel:
will qa in bug bash.

PR #594, opened 2024-04-11
Set EmailVerified to true for all remote users during migration

JulienTant:
#### Summary
`UpdateUser` ignores the `EmailVerified`, so our `UpdateUser` calls have no effect and we would try to do it at every `syncUser`. In addition, doing the Uodate one by one was quite hurtful.

This fixes the issue by migrating all necessary users in one query during a migration.

#### Ticket Link
N/A

### DIFF
--- a/server/message_hooks_test.go
+++ b/server/message_hooks_test.go
@@ -2825,7 +2825,7 @@ func TestUserWillLogin(t *testing.T) {
 			AutomaticallyPromoteSyntheticUsers: true,
 			SetupAPI: func(api *plugintest.API) {
 				api.On("UpdateUser", mock.MatchedBy(func(user *model.User) bool {
-					return *user.RemoteId == ""
+					return !user.IsRemote()
 				})).Once().Return(nil, nil)
 				api.On("LogInfo", "Promoted synthetic user", "user_id", testutils.GetID()).Once()
 			},
@@ -2844,7 +2844,7 @@ func TestUserWillLogin(t *testing.T) {
 			AutomaticallyPromoteSyntheticUsers: true,
 			SetupAPI: func(api *plugintest.API) {
 				api.On("UpdateUser", mock.MatchedBy(func(user *model.User) bool {
-					return *user.RemoteId == ""
+					return !user.IsRemote()
 				})).Once().Return(nil, model.NewAppError("UpdateUser", "err from test", nil, "", 500))
 				api.On("LogWarn", "Failed to promote synthetic user", "user_id", testutils.GetID(), "err", "err from test").Once()
 			},
--- a/server/message_hooks_test.go
+++ b/server/message_hooks_test.go
@@ -2819,22 +2819,13 @@ func TestUserWillLogin(t *testing.T) {
 		{
 			Name: "Autopromotion works",
 			User: &model.User{
-				Id:            testutils.GetID(),
-				EmailVerified: false,
+				Id: testutils.GetID(),
 			},
 			UserIsRemote:                       true,
 			AutomaticallyPromoteSyntheticUsers: true,
 			SetupAPI: func(api *plugintest.API) {
 				api.On("UpdateUser", mock.MatchedBy(func(user *model.User) bool {
-					if user.EmailVerified == false {
-						return false
-					}
-
-					if *user.RemoteId != "" {
-						return false
-					}
-
-					return true
+					return *user.RemoteId == ""
 				})).Once().Return(nil, nil)
 				api.On("LogInfo", "Promoted synthetic user", "user_id", testutils.GetID()).Once()
 			},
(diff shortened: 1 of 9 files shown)


server/store/sqlstore/migrations.go
```diff
@@ -17,6 +17,18 @@ func (s *SQLStore) runMigrationRemoteID(remoteID string) error {
 	return err
 }
 
+func (s *SQLStore) runSetEmailVerifiedToTrueForRemoteUsers(remoteID string) error {
+	_, err := s.getQueryBuilder().
```
JulienTant, line 21: Changing the code locally  to get the SQL query if you are interested:

```sql
UPDATE Users SET EmailVerified = $1 WHERE (RemoteID = $2 AND EmailVerified = $3)
```

```
ARGS=[true pp9fktok8bfr8bmztwkyerifoh false]
```

note:remoteID will be different for each installation so yours will differ

sbishel: Is there any way we can test the performance of this query against a large `Users` table.

sbishel (commented):
I'm ok with the changes, but would like to test the query for performance.

server/message_hooks_test.go
```diff
@@ -2847,22 +2838,13 @@ func TestUserWillLogin(t *testing.T) {
...
-					if user.EmailVerified == false {
-						return false
-					}
-
-					if *user.RemoteId != "" {
-						return false
-					}
-
-					return true
+					return *user.RemoteId == ""
```
wiggin77, line 2847: This will panic when RemoteId is nil.  I suggest using `user.IsRemote`

wiggin77 (changes requested):

server/message_hooks_test.go
JulienTant: Thanks @wiggin77 I pushed a fix for this!

server/store/sqlstore/migrations.go
JulienTant: I'm adding 30K users to my local env right now, not sure if there's better way! If perfs are good, we'll ship my mac to production :D

wiggin77 (approved):
LGTM :+1:

server/store/sqlstore/migrations.go
JulienTant: Status update:

With 9628 users, it takes ~730ms locally.

```jsonl
{"timestamp":"2024-04-11 15:23:53.975 -07:00","level":"debug","msg":"Starting runSetEmailVerifiedToTrueForRemoteUsers","caller":"app/plugin_api.go:1001","plugin_id":"com.mattermost.msteams-sync"}
{"timestamp":"2024-04-11 15:23:54.705 -07:00","level":"debug","msg":"Finished runSetEmailVerifiedToTrueForRemoteUsers","caller":"app/plugin_api.go:1001","plugin_id":"com.mattermost.msteams-sync","duration":"729.708167ms","updated_users":"9628"}
```

Still waiting to get my 30K users ready!

sbishel (approved):
I just tested on my local machine, by setting several (9) of the user "EmailVerified" fields to false. Running v1.11.0 and seeing that number (9) of `Updating user profile` messages. I then built and ran this PR. Everything got set to true and no `Updating user profile` messages were seen.

server/store/sqlstore/migrations.go
sbishel: Thanks for doing this...make me feel better.

enahum (approved):

server/store/sqlstore/migrations.go
```diff
@@ -17,6 +17,18 @@ func (s *SQLStore) runMigrationRemoteID(remoteID string) error {
 	return err
 }
 
+func (s *SQLStore) runSetEmailVerifiedToTrueForRemoteUsers(remoteID string) error {
```
calebroseland, line 20: Not an issue per se, but this will run on every startup since there's no short-circuit or negation switch. 

(Looking forward to the day we have sequenced/conditional migrations!)

calebroseland (approved):
LGTM

server/store/sqlstore/migrations.go
JulienTant: Yeah, that's why the condition has the `RemoteID=xx AND Emailverified == false`, in theory, the resultset with those conditions should always be 0... except for the first run.

server/store/sqlstore/migrations.go
JulienTant: 15K users => 1.134357s

```
{"timestamp":"2024-04-11 17:21:16.516 -07:00","level":"debug","msg":"Starting runSetEmailVerifiedToTrueForRemoteUsers","caller":"app/plugin_api.go:1001","plugin_id":"com.mattermost.msteams-sync"}
{"timestamp":"2024-04-11 17:21:17.650 -07:00","level":"debug","msg":"Finished runSetEmailVerifiedToTrueForRemoteUsers","caller":"app/plugin_api.go:1001","plugin_id":"com.mattermost.msteams-sync","duration":"1.134357s","updated_users":"15032"}
```

sbishel: So double or triple that and we are at 2.5 - 3 seconds. Which I think is ok for this, at least its not 30. But as Caleb mentioned below, we have to be careful because right now, these will run every time the plugin starts. We ran into some long migration queries in Boards.

I'm ok merging and moving forward.

JulienTant: Final report here, with 27K users it took 2.1sec so it does not increase exponentially. 

Restarting the plugin a few times in a row with the 27K users gives me 2.1sec the first time (because we update all the users the first time), and all following restarts of the plugin take ~50ms with 0 users affected. So I don't see how the migration can grow much slower, unless we somehow insert a ton of synthetic user with EmailVerified=false... which should not happen anymore I reckon.

sbishel: Awesome. I say we move forward, merge and cherry-pick. Then I can start the build. Thanks.

PR #598, opened 2024-04-12
POC of queue to store events from MSTeams

jespino:
This is a POC in draft of storing things in a queue inside mattermost.

The idea is, each event get stored in the databases, if the even gets processed right away, we delete it from the queue.

In the future I will add a pooling mechanism that is going to try again up to 5 times each element in the queue that has more than 0 tries, and we start trying to deliver everything that is in the queue.

This would allow us to provide a more resilient event handling. My main concern here is the performance impact, I think we have space to reduce the database load, but I'm not sure how much.

Probably would be cool to have a load test with this code to figure out the degree of impact on the database.


enahum:
I have run a load test with this PR.

1. There are no crashes with high load which is good
2. There are no metrics on the three new store functions as there are not present in the timelayer (will try again after generating)
3. a previous run was able to support 13953 users and with this PR it went down to 13776 users
4. Database latency seems to have increased as well (see graphs)

| Previous | This PR |
| --- | --- |
| image | image | 

I'll run it again after executing `make generate` and report back

lieut-data:
@jespino, propose we pause on these changes while we define next steps for the project.

lieut-data:
We're slimming down the plugin to reach code complete for security review, so won't need this update.

PR #600, opened 2024-04-16
MM-57565 - Set root id if it exists on attachment sync warning

sbishel:
#### Summary
Set the rootID for the new post to the existing posts root id. Result is if null, new post displays in center section and if a reply it will appear in the thread list.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-57565

Screenshot 2024-04-15 at 6 24 46 PM


calebroseland (approved):
Is is necessary to remove the CreateAt?

JulienTant (approved):
Looks good @sbishel

sbishel:
Will QA during bug bash

PR #601, opened 2024-04-16
Track status_code as part of the MSGraphClient method duration

mgdelacroix:
#### Summary
Status code will be set at 2XX for all successful requests as we don't have a returning code coming from the SDK in those cases. For errors, the status code will come from the SDK error and will be set to zero if not present.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56678


lieut-data (approved):
This is awesome, thank you @mgdelacroix!

sbishel (approved):

PR #602, opened 2024-04-17
Add DB replica support

enahum:
#### Summary
Adds Database read replica support when available

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57139


jespino (approved):
I like this one :)

enahum:
@jespino @mgdelacroix @wiggin77 are we ok merging this one ?

server/store/sqlstore/migrations.go
```diff
@@ -37,7 +37,7 @@ const (
 
 func (s *SQLStore) runMSTeamUserIDDedup() error {
 	// get all users with duplicate msteamsuserid
-	rows, err := s.getQueryBuilder().Select(
+	rows, err := s.getMasterQueryBuilder().Select(
```
wiggin77, line 40: Why is master needed here for a SELECT?

wiggin77 (approved):
This looks great!  :+1:   One item for consideration re using master for a select where replication lag would not be a factor.

server/store/sqlstore/migrations.go
enahum: @wiggin77 I assumed is needed as this is a migration to avoid any sort of lag. I can change it to use the replica if you think is best

enahum: I'll change it to use the replica and then merge.. if this causes issues we can always switch to use master again

PR #605, opened 2024-04-17
Adds loglines and elapsed times to database migrations

mgdelacroix:


calebroseland (approved):

sbishel (approved):

calebroseland (approved):

mgdelacroix:
Using mocks for the API in the store turns out to be a problem, as there is no easy way to mock the variadic arguments of the logger. At this point, I think it's worth to close this PR in favor of working in the morph integration directly

PR #612, opened 2024-04-18
Fix whitelist upload batching

calebroseland:
#### Summary
Fix bug where whitelist uploads would fail if size >100.

CleanShot 2024-04-18 at 12 20 35

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -807,6 +807,8 @@ func (a *API) updateWhitelist(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	a.p.API.LogWarn("Whitelist updated", "size", len(ids))
+
 	w.WriteHeader(http.StatusOK)
 	w.Header().Set("Content-Type", "application/json")
 	if err := json.NewEncoder(w).Encode(&UpdateWhitelistResult{
--- a/server/api.go
+++ b/server/api.go
@@ -807,7 +807,7 @@ func (a *API) updateWhitelist(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


server/api.go
```diff
@@ -807,6 +807,8 @@ func (a *API) updateWhitelist(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	a.p.API.LogWarn("Whitelist updated", "size", len(ids))
```
JulienTant, line 810: shouldn't this be an info?

server/store/sqlstore/store.go
```diff
@@ -1006,11 +1006,12 @@ func (s *SQLStore) SetWhitelist(userIDs []string, batchSize int) error {
 	for i, id := range userIDs {
 		currentBatch = append(currentBatch, id)
 		if len(currentBatch) >= batchSize || i == len(userIDs)-1 {
+			// batch threshold met, or end of list
 			if err = s.storeUsersInWhitelist(currentBatch, tx); err != nil {
 				s.api.LogDebug("Error adding batched users to whitelist", "error", err.Error(), "userIds", currentBatch)
 				return err
 			}
-			clear(currentBatch)
+			currentBatch = nil
```
JulienTant, line 1014: TIL clear keeps the size!

JulienTant (approved):
LGTM, just a small remark on the log level

server/api.go
calebroseland: Yes, bad copy-paste. Thanks for catching!

sbishel (approved):

PR #619, opened 2024-04-20
Integrate morph into the plugin

mgdelacroix:
#### Summary
This PR integrates `morph` into the plugin store code, moving all migrations to SQL files, splitting them for improved clarity and structure. It creates a new `msteamssync_system_settings` table that stores one row per data migration, allowing them to indicate when they've run and removing the need for checking the database status every time the plugin starts to see if each data migration should run.

The `Migrate` method implements a sequence of migrations that ensures that at every step all the requirements of the next migration are met or checked (either by morph or the data migration through the previously mentioned table), and every step logs when the migration starts, stops and the elapsed time.


server/store/sqlstore/migrate.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (s *SQLStore) ensureMigrationsAppliedUpToVersion(engine *morph.Morph, driver drivers.Driver, version int) error {
+	applied, err := driver.AppliedMigrations()
+	if err != nil {
+		return err
+	}
+	currentVersion := len(applied)
+
+	s.api.LogDebug("== Ensuring migrations applied up to version ====================", "version", version, "current_version", currentVersion)
```
lieut-data, line 117: This is unrelated to this PR, but I've always found it odd that the morph logs are formatted (i.e. with `====...` as it they were being consumed by a user at a console, when instead we mostly see them as messages inside server logs. 0/5

server/store/sqlstore/migrate.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// runMigrationSequence executes all the migrations in order, both
+// plain SQL and data migrations.
+func (s *SQLStore) runMigrationSequence(engine *morph.Morph, driver drivers.Driver, remoteID string) error {
+	if mErr := s.ensureMigrationsAppliedUpToVersion(engine, driver, migrationRemoteIDRequiredVersion); mErr != nil {
+		return mErr
+	}
+
+	if remoteID != "" {
```
lieut-data, line 76: As I recall, these are data migrations that require running code vs. just SQL. Is the idea to continue to append to this method as such migrations are required, or in the long run would we split apart the two concepts (schema vs data)? I realize that would require some finagling given the need to run some data migrations "between" others.

@jespino, is this some shared channels complexity would could eliminate if we required users adopting two way sync to "reset" somehow?

server/store/sqlstore/migrations/000001_system_settings_table.down.sql
```diff
@@ -0,0 +1 @@
+SELECT 1;
```
lieut-data, line 1: nit: could morph just treat an empty file as a no-op?

lieut-data (approved):
Thanks, @mgdelacroix! Besides the few comments below, wondering if we "need" 15 separate migrations or if we could just have a single monolithic migration to get us "up to date", and start adding new ones discretely thereafter?

server/store/sqlstore/migrate.go
mgdelacroix: Agree, the idea was to make all migration logs (inside and outside of `morph`) similar, but we can change the ones that come from the store to avoid the padding and the `=`s if we want, 0/5 really on my side

mgdelacroix:
@lieut-data we could do one big "init" migration only, but there are two downsides that I can think of:
- There are some data migrations that can only run in an intermediate state, I'm thinking specifically on the whitelisted_users data migration, that we run just before migration 14, which deletes the table. We could have two steps then, or ask current plugin users to follow an update plan (so they update to the last version before `morph`, run the plugin and then update to the first version including it, for example), but that's added complexity.
- The migration will run as a single big transaction for instances that already have the plugin installed. The queries are thought to be no-ops for the current last version of the plugin's database, but if something unexpected happens, I think we'd be in a better situation having small steps to rollback and rerun.

server/store/sqlstore/migrations/000001_system_settings_table.down.sql
mgdelacroix: Heh, actually having an empty file is only a problem for MySQL, Postgres will process those perfectly, so I'm making that change

mgdelacroix: Done in 3225998f7452cd5132b3e3d0c556eb8c84d64ed1

sbishel (approved):
Nice to get this in.

PR #621, opened 2024-04-23
Hide connect message, hide attachment message on update

sbishel:
#### Summary

- Hide messages that ask a user to "Click here to connect".
- The message will still display if a connected user loses their connection `OnDisconnectedTokenHandler` and if the command `/msteams connect` is run. 

- The warning message regarding messages not being delivered will not be displayed.

Code was only commented out, as we expect to re-enable in the future.

#### Ticket Link

 Fixes https://mattermost.atlassian.net/browse/MM-57920


lieut-data (approved):
Thanks, @sbishel! I assume we'll cherry pick this to a release-v1.12 branch?

calebroseland (approved):

PR #586, opened 2024-04-04
MM-57494: welcome message after connecting

lieut-data:
#### Summary
This PR welcomes users when the connect with a message from the bot:

CleanShot 2024-04-04 at 17 54 18@2x

When they choose Mattermost or Microsoft Teams during onboarding (or anytime afterwards from the user preferences pane), they are notified as such:

**Choosing Mattermost**
CleanShot 2024-04-04 at 17 54 39@2x

**Choosing Teams**
CleanShot 2024-04-04 at 17 54 56@2x

##### Other Changes

This pull request depends on https://github.com/mattermost/mattermost-plugin-msteams/pull/563 for updated unit tests. (Do not merge until that is merged to `main`.) To test the above, I took @hmhealey's excellent mocked unit tests and modified them to fit into the new "mock-free" testing paradigm. (That's the bulk of this PR -- sorry for the noise! This will all get easier once we've expunged the mock tests.)

I also took the opportunity to centralize most bot messages into a `bot_messages.go`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57494

PR #623, opened 2024-04-23
rewrite automute tests

lieut-data:
#### Summary
This extracts the portion of https://github.com/mattermost/mattermost-plugin-msteams/pull/586 that finished rewriting the automute tests using the new (mock free) testing framework.

#### Ticket Link
None.

### DIFF
--- a/server/automute_channel_test.go
+++ b/server/automute_channel_test.go
@@ -5,32 +5,32 @@ import (
 	"time"
 
 	"github.com/mattermost/mattermost/server/public/model"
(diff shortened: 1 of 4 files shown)


sbishel (approved):

hmhealey (approved):
Looks great! I'm a bit sad to lose the mocked server since I'm still proud of it, but being able to test against a real server without having to constantly confirm that it works the same way as the mock makes me a lot more confident in these tests

PR #624, opened 2024-04-24
only display message if user is connected

sbishel:
#### Summary
When the user changes their "primary" platform, we display a bot message. We also implemented a fix that resets the users primary platform back to Mattermost upon disconnect. This resulted in the bot message whenever `/msteams disconnect` was run.

This PRs as a check for the user being active and only showing the message if the user is active.

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-57975

### DIFF
--- a/server/automute_preferences.go
+++ b/server/automute_preferences.go
@@ -35,7 +35,16 @@ func (p *Plugin) updateAutomutingOnPreferencesChanged(_ *plugin.Context, prefere
 	}
 
 	for _, userID := range userIDsToDisable {
-		p.notifyUserMattermostPrimary(userID)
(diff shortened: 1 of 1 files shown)


lieut-data (approved):
Thanks, @sbishel! Would you consider rebasing this on https://github.com/mattermost/mattermost-plugin-msteams/pull/623 and then updating those tests to reflect this new requirement?

calebroseland (approved):

sbishel:
https://github.com/mattermost/mattermost-plugin-msteams/pull/623 is required for the new tests to pass.

PR #626, opened 2024-04-25
[MM-57197] Weekly active users

JulienTant:
#### Summary
Add weekly active users metrics, letting us know how many users sent/received DM/GMs (aka chat in the MSTeams language) through the plugin in the last 7 days.

- For each message sent through the plugin, we update the sender user line to set a new LastChatSendAt.
- For each message received through the plugin, we update each person who will receive the message, excluding the original sender.
- Created the 2 associated gauges.

Maybe some changes deserved to be in their own PRs but I wanted to "fix" some things I've seen:
- `store.GetStats()` has been removed in favor of having a list of individual queries
- `api.siteStats()` now uses one of those new methods from above.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57197


server/api.go
```diff
@@ -927,7 +927,9 @@ func (a *API) siteStats(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	stats, err := a.p.store.GetStats(a.p.GetRemoteID(), a.p.GetPreferenceCategoryName())
+	stats, err := a.p.store.GetStats(storemodels.GetStatsOptions{
+		Stats: []storemodels.StatType{storemodels.StatsConnectedUsers},
```
JulienTant, line 931: I realized that we are adding more stats to parse from `GetStats()` but this API call just needs one of them, so I added a way to only parse certain stats when needed

.tool-versions
```diff
@@ -0,0 +1,2 @@
+golang 1.21.8
```
JulienTant, line 1: Can remove if needed: I'm using `asdf` to manager my nodejs/go versions.

server/handlers/handlers.go
```diff
@@ -388,6 +390,24 @@ func (ah *ActivityHandler) handleCreatedActivity(msg *clientmodels.Message, subs
...
+		// exclude the sender from the list of participants
+		participants := []string{}
+		for _, userID := range userIDs {
+			if userID != post.UserId {
+				participants = append(participants, userID)
+			}
+		}
+
+		if len(participants) > 0 {
+			err := ah.plugin.GetStore().SetUsersLastChatReceivedAt(participants, post.CreateAt*1000)
```
JulienTant, line 404: times are `*1000` because Mattermost uses milliseconds, but we use microseconds in the plugin.

server/message_hooks.go
```diff
@@ -609,6 +609,11 @@ func (p *Plugin) SendChat(srcUser string, usersIDs []string, post *model.Post, c
 			p.API.LogWarn("Error updating the msteams/mattermost post link metadata", "error", err)
 		}
 	}
+
+	if err := p.store.SetUserLastChatSentAt(post.UserId, post.CreateAt*1000); err != nil {
```
JulienTant, line 613: times are `*1000` because Mattermost uses milliseconds, but we use microseconds in the plugin.

JulienTant:
@lindy65 I marked this PR as deferred because I feel like metrics are easier to debug with grafana!

lindy65 (approved):
Thanks @JulienTant - this can be tested in Grafana post merge

.tool-versions
lieut-data: One concern is that this "duplicates" the source of truth that is now go.mod, at least as of Go v1.21. I'm only 0/5, though if this is useful!

server/store/sqlstore/store.go
```diff
@@ -871,69 +879,110 @@ func (s *SQLStore) StoreOAuth2State(state string) error {
 	return nil
 }
 
-func (s *SQLStore) GetStats(remoteID, preferenceCategory string) (*storemodels.Stats, error) {
-	query := s.getQueryBuilder().Select("count(mmChannelID)").From(linksTableName)
-	row := query.QueryRow()
+func (s *SQLStore) GetStats(opts storemodels.GetStatsOptions) (*storemodels.Stats, error) {
```
lieut-data, line 882: Given these are mostly distinct queries, would it be clearer if we just exposed `GetLinkedChannelsCount`, `GetConnectedUsersCount`, etc.?

server/store/sqlstore/store_test.go
```diff
@@ -1017,6 +1018,121 @@ func TestWhitelistIO(t *testing.T) {
...
+	err := store.SetUserInfo(mmUserID, "ms-"+mmUserID, nil)
+	assert.Nil(err)
+
+	getLastChatSentAtForUser := func(mmUserID string) int64 {
+		t.Helper()
+		var lastChatSentAt int64
+		err = store.getQueryBuilder().
+			Select("LastChatSentAt").
+			From(usersTableName).
+			Where(sq.Eq{"mmuseriD": mmUserID}).
```
lieut-data, line 1035:
```suggestion
			Where(sq.Eq{"mmuserID": mmUserID}).
```

server/store/sqlstore/store_test.go
```diff
@@ -1017,6 +1018,121 @@ func TestWhitelistIO(t *testing.T) {
...
+			QueryRow().
+			Scan(&lastChatSentAt)
+		assert.Nil(err)
+		return lastChatSentAt
+	}
+
+	{
```
lieut-data, line 1042: I haven't seen this pattern of brackets -- can you elaborate on the intent? Should these be `t.Run` with a description instead of the comment?

server/plugin.go
```diff
@@ -878,19 +880,29 @@ func (p *Plugin) GetRemoteID() string {
...
 	p.GetMetrics().ObserveConnectedUsers(stats.ConnectedUsers)
 	p.GetMetrics().ObserveSyntheticUsers(stats.SyntheticUsers)
 	p.GetMetrics().ObserveLinkedChannels(stats.LinkedChannels)
 	p.GetMetrics().ObserveMattermostPrimary(stats.MattermostPrimary)
 	p.GetMetrics().ObserveMSTeamsPrimary(stats.MSTeamsPrimary)
-	p.API.LogDebug("Updating metrics done", "duration", time.Since(start))
+	p.GetMetrics().ObserveActiveUsersSending(stats.ActiveUsersSending)
+	p.GetMetrics().ObserveActiveUsersReceiving(stats.ActiveUsersReceiving)
+
+	p.API.LogInfo("Updating metrics done", "duration_ms", time.Since(now).Milliseconds())
```
lieut-data, line 905: nit: I think we already get this from metrics for each SQL query. Is there value in logging separately?

server/plugin.go
```diff
@@ -878,19 +880,29 @@ func (p *Plugin) GetRemoteID() string {
...
-	p.API.LogDebug("Updating metrics")
-	stats, err := p.store.GetStats(p.remoteID, PreferenceCategoryPlugin)
+	now := time.Now()
+	p.API.LogInfo("Updating metrics")
+
+	stats, err := p.store.GetStats(storemodels.GetStatsOptions{
+		RemoteID:           p.remoteID,
+		PreferenceCategory: PreferenceCategoryPlugin,
+		ActiveUsersFrom:    now.Add(-metricsActiveUsersRange),
+		ActiveUsersTo:      now,
```
lieut-data, line 890: This parameter feels incorrect: is it actually meaningful to query to anything other than `ActiveUsersTo == now`?

server/metrics/metrics.go
```diff
@@ -381,6 +385,24 @@ func NewMetrics(info InstanceInfo) Metrics {
 	})
 	m.registry.MustRegister(m.upstreamUsers)
 
+	m.activeUsersSending = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemApp,
+		Name:        "active_users_sending",
+		Help:        "The total number of active users sending messages.",
```
lieut-data, line 392:
```suggestion
		Help:        "The number of users who have sent messages in the last week.",
```

server/metrics/metrics.go
```diff
@@ -381,6 +385,24 @@ func NewMetrics(info InstanceInfo) Metrics {
...
+		ConstLabels: additionalLabels,
+	})
+	m.registry.MustRegister(m.activeUsersSending)
+
+	m.activeUsersReceiving = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemApp,
+		Name:        "active_users_receiving",
+		Help:        "The total number of active users receiving messages.",
```
lieut-data, line 401:
```suggestion
		Help:        "The number of users who have received messages in the last week.",
```

server/api.go
lieut-data: Commented below, but wondering if it would be simpler just to expose discrete store methods instead?

server/handlers/handlers.go
```diff
@@ -388,6 +390,24 @@ func (ah *ActivityHandler) handleCreatedActivity(msg *clientmodels.Message, subs
 		}
 	}
 
+	// Update the last chat received at for the participants of DM/GM
+	if chat != nil && len(userIDs) > 0 {
+		// exclude the sender from the list of participants
+		participants := []string{}
```
lieut-data, line 396: Propose sizing this to match the expected output:
```suggestion
		participants := make([]string{}, len(userIDs) - 1)
```

server/handlers/handlers.go
lieut-data: Is there a way to make the `*1000` explicit? (i.e. using some `time` method?)

lieut-data (changes requested):
This is fantastic! Looking at the schema changes, I think it's backwards compatible (so we can easily roll back if needed), but can we be sure to test this? A few other comments below.

server/plugin.go
JulienTant: I don't run a grafana locally (maybe I should? :P) and I usually like to try those new queries against a bigger db, the log gives me a way to know how long it took. 0/5 to switch this to `LogDebug` if it makes more sense.

server/handlers/handlers.go
JulienTant: Using go's `time` package would be quite ugly:

```go
(time.Duration(post.CreateAt) * time.Millisecond).Microseconds()
```

But for the sake of making this more clear, I'm ok with creating a helper `MilliToMicroSeconds()` that does the `*1000`

lieut-data: All good, only 0/5 here -- whatever is clearest.

server/plugin.go
lieut-data: > I don't run a grafana locally

Yes, definitely do this -- it's awesome! It's semi-automated with the monorepo if you're using Docker, but also trivial to spin up manually.

I'm leaning against adding one-off metrics as debug logs /if/ we have good ways of measuring them otherwise. (Log hygiene and all that.) But only 1/5 here.

server/store/sqlstore/store_test.go
JulienTant: That was the sign of too long days and not enough will left to write `t.Run` tests 😅 i refactored it!

server/handlers/handlers.go
JulienTant: I'm switching this to a capacity set `make([]string, 0, len(userIDs) - 1)`

I think there's risk in setting a strict length here because the source of this is an external API, and history gave me many reasons to be scared of setting hard expectations when data is external 😅

lieut-data: Yes! Agree with `append` semantics, I should have suggested that :p

JulienTant:
Hey @sbishel when you have a minute can you take a look? 👀

sbishel (approved):
LGTM

PR #629, opened 2024-04-26
MM-57535 Don't automute guests nor bots

sbishel:
#### Summary
Don't automute Direct messages with Guests, nor Bots

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-57288
  Fixes https://mattermost.atlassian.net/browse/MM-57535

Unit test won't pass until we are using v9.8 of the server, as it contains a fix for creating channels with remote users.


server/helper_test.go
```diff
@@ -321,6 +321,36 @@ func (th *testHelper) SetupRemoteUser(t *testing.T, team *model.Team) *model.Use
...
+	bot := &model.Bot{
+		Username:    "bot" + id,
+		DisplayName: "a bot",
+		Description: "bot",
+		OwnerId:     "ownerID",
+	}
+
+	bot, err := th.p.API.CreateBot(bot)
+	if err != nil {
+		panic(err)
```
lieut-data, line 349: Can we pass `t *testing.T`, use `t.Helper` and `require.NoError` here instead of panicking?

server/automute_channel.go
```diff
@@ -45,6 +45,12 @@ func (p *Plugin) updateAutomutingOnChannelCreated(channel *model.Channel) error
 		return nil
 	}
 
+	canAutomute, err := p.canAutomuteChannel(channel)
```
lieut-data, line 48: Can we remove the `channel.IsGroupOrDirect` above now that it's part of the `canAutomuteChannel` check?

server/automute.go
```diff
@@ -159,8 +160,20 @@ func (p *Plugin) canAutomuteChannelID(channelID string) (bool, error) {
 // canAutomuteChannel returns true if the channel is either explicitly linked to a channel in MS Teams or if it's a
 // DM/GM channel that is implicitly linked to MS Teams.
 func (p *Plugin) canAutomuteChannel(channel *model.Channel) (bool, error) {
-	// Automute all DM/GM channels
-	if channel.IsGroupOrDirect() {
+	// Automute all GM channels
+	if channel.Type == model.ChannelTypeGroup {
+		return true, nil
+	} else if channel.Type == model.ChannelTypeDirect {
+		userIDs := strings.Split(channel.Name, "__")
```
lieut-data, line 167: I was surprised we didn't expose this from the `model` package! Maybe a future addition :)

lieut-data (approved):
This is awesome! Wondering how we want to approach "migrating" existing such channels? Maybe we can defer to do this holistically with https://mattermost.atlassian.net/browse/MM-58054 where we need to migrate again.

sbishel:
@lieut-data - I don't think this PR is even necessary anymore.
https://mattermost.atlassian.net/browse/MM-58053 only mutes if Synthetic user, and Guest nor Bot would ever be synthetic. I think we can close.

lieut-data:
It's true that https://mattermost.atlassian.net/browse/MM-58053 supercedes this, but I'd love to keep the unit tests at least as they are valid cases that we'd want to guarantee going forward regardless.

server/helper_test.go
sbishel: Yes...thanks.

server/automute_channel.go
sbishel: We could, but it still servers a purpose to short-circuits channels and not even bother with canAutomuteChannel.

enahum (approved):
Lgtm

PR #630, opened 2024-04-27
[MM-57533] Send a warning when selective sync blocks a message

JulienTant:
#### Summary
Send a warning when selective sync blocks a message + Add tests on the `MessageWIllBePosted` hook.

#### QA Steps
- Make sure Selective Sync is on
- Have 2 MSTeams users connected to Mattermost
- When user 1 tries to message user 2, an ephemeral message let them know that the message won't be sent to MS Teams

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57533

### DIFF
--- a/server/helper_test.go
+++ b/server/helper_test.go
@@ -140,12 +140,14 @@ func (th *testHelper) clearDatabase(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func (th *testHelper) Reset(t *testing.T) *testHelper {
+func (th *testHelper) reset(t *testing.T, keepDB bool) *testHelper {
 	t.Helper()
 
-	// Wipe the tables for this plugin to ensure a clean slate. Note that we don't currently
-	// touch any Mattermost tables.
-	th.clearDatabase(t)
+	if !keepDB {
+		// Wipe the tables for this plugin to ensure a clean slate. Note that we don't currently
+		// touch any Mattermost tables.
+		th.clearDatabase(t)
+	}
 
 	appClientMock := &mocks.Client{}
 	clientMock := &mocks.Client{}
@@ -198,6 +200,17 @@ func (th *testHelper) Reset(t *testing.T) *testHelper {
 	})
 
 	return th
+}
+
+func (th *testHelper) Reset(t *testing.T) *testHelper {
+	t.Helper()
+	return th.reset(t, false)
+}
+
+func (th *testHelper) ResetPreserveDB(t *testing.T) *testHelper {
+	t.Helper()
+	return th.reset(t, true)
+
 }
 
 func (th *testHelper) SetupTeam(t *testing.T) *model.Team {
@@ -321,30 +334,10 @@ func (th *testHelper) SetupRemoteUser(t *testing.T, team *model.Team) *model.Use
 	return user
 }
 
-func (th *testHelper) SetupSyntheticUser(t *testing.T, team *model.Team) *model.User {
-	t.Helper()
-
-	user := th.SetupRemoteUser(t, team)
-	teamsID := "t" + user.Id
-	err := th.p.store.SetUserInfo(user.Id, teamsID, nil)
-	require.NoError(t, err)
-
-	return user
-}
-
 func (th *testHelper) ConnectUser(t *testing.T, userID string) {
 	teamID := "t" + userID
 	err := th.p.store.SetUserInfo(userID, teamID, &oauth2.Token{AccessToken: "token", Expiry: time.Now().Add(10 * time.Minute)})
 	require.NoError(t, err)
-}
-
-func (th *testHelper) SetupConnectedUser(t *testing.T, team *model.Team) *model.User {
-	t.Helper()
-
-	user := th.SetupUser(t, team)
-	th.ConnectUser(t, user.Id)
-
-	return user
 }
 
 func (th *testHelper) SetupSysadmin(t *testing.T, team *model.Team) *model.User {
--- a/server/helper_test.go
+++ b/server/helper_test.go
@@ -140,14 +140,12 @@ func (th *testHelper) clearDatabase(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func (th *testHelper) reset(t *testing.T, keepDB bool) *testHelper {
+func (th *testHelper) Reset(t *testing.T) *testHelper {
 	t.Helper()
 
-	if !keepDB {
-		// Wipe the tables for this plugin to ensure a clean slate. Note that we don't currently
-		// touch any Mattermost tables.
-		th.clearDatabase(t)
-	}
+	// Wipe the tables for this plugin to ensure a clean slate. Note that we don't currently
+	// touch any Mattermost tables.
+	th.clearDatabase(t)
 
 	appClientMock := &mocks.Client{}
 	clientMock := &mocks.Client{}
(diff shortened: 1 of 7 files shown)


lindy65 (approved):
I'm QA-deferring this PR @JulienTant and will test post merge

server/helper_test.go
```diff
@@ -449,6 +461,39 @@ func (th *testHelper) assertEphemeralMessage(t *testing.T, userID, channelID, me
 	}
 }
 
+func (th *testHelper) assertNoEphemeralMessage(t *testing.T, userID, channelID, message string, maxWaitTime time.Duration) {
```
lieut-data, line 464: Neat! I had half a mind to make this the default behaviour at the end of every test, but naturally it's wasteful to wait every time. Thank you!

server/helper_test.go
```diff
@@ -449,6 +461,39 @@ func (th *testHelper) assertEphemeralMessage(t *testing.T, userID, channelID, me
...
+				require.True(t, ok, "failed to find post in ephemeral message websocket event")
+
+				var post model.Post
+				err := json.Unmarshal([]byte(postJSON), &post)
+				require.NoError(t, err)
+
+				if post.ChannelId == channelID && post.Message == message {
+					t.Fatalf("received undesired ephemeral message in channel %s: %s", post.ChannelId, post.Message)
+				}
+			}
```
lieut-data, line 488: I wonder if asserting "no" message should be simpler and just accept the `channelID`? I could see case where we start emitting a different message but don't check for that because of this equality test.

server/message_hooks.go
```diff
@@ -97,6 +97,10 @@ func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
 		return
 	}
 
+	if post.UserId == p.GetBotUserID() {
+		return
+	}
+
```
lieut-data, line 103: Fascinating: is this an extra check, or was this literally something we didn't previously catch?

server/message_hooks_test.go
```diff
@@ -2880,3 +2881,136 @@ func TestUserWillLogin(t *testing.T) {
 		})
 	}
 }
+
+func TestMessageHasBeenPosted(t *testing.T) {
```
lieut-data, line 2885: ❤️

server/message_hooks_test.go
```diff
@@ -2880,3 +2881,136 @@ func TestUserWillLogin(t *testing.T) {
...
+	assert := require.New(t)
```
lieut-data, line 2886: I think this `assert` is tied to the wrong `testing.T` -- typically, you'd want it anchored to the nearest one, i.e. recreating it inside each `t.Run`.

server/message_hooks_test.go
```diff
@@ -2880,3 +2881,136 @@ func TestUserWillLogin(t *testing.T) {
...
+		if appErr != nil {
+			if appErr.Id != "app.sharedchannel.dm_channel_creation.internal_error" {
+				assert.Nil(appErr)
+			}
+			channel, _ = th.p.API.GetDirectChannel(senderID, receiverID)
+		}
+		return channel
+	}
+
+	setup := func(t *testing.T, selectiveSync bool) (msTeamsUser, MMUser, syntheticUser *model.User) {
```
lieut-data, line 2903: nit: `mmUser` to match `msTeamsUser`?

server/message_hooks_test.go
```diff
@@ -2880,3 +2881,136 @@ func TestUserWillLogin(t *testing.T) {
...
+		th.ResetPreserveDB(t)
+
+		team := th.SetupTeam(t)
+		msTeamsUser = th.SetupUser(t, team)
```
lieut-data, line 2907: nit: I thought `msTeamsUser` would be an instance of the remote Teams structure. Should this be named `teamsPrimaryUser` (and similarly `mmPrimaryUser`?)

lieut-data (changes requested):
This is awesome, thank you @JulienTant!

One potential change might be coming with https://docs.google.com/document/d/18A228q51idY2lAT89O2AnT5VZRqzqxOM9j-wv0A6tXE/edit -- namely the feedback from endusers that "blocking" messaging in these cases is unhelpful (and the bot messages distracting). I'd love to keep the majority of the changes you've added here (including the unit tests! YAY!), but we may end up pivoting on this to just recognize Teams Primary <> Teams Primary chat as an "unmuted" MM chat.

server/message_hooks_test.go
```diff
@@ -2880,3 +2881,136 @@ func TestUserWillLogin(t *testing.T) {
...
+			if appErr.Id != "app.sharedchannel.dm_channel_creation.internal_error" {
+				assert.Nil(appErr)
+			}
+			channel, _ = th.p.API.GetDirectChannel(senderID, receiverID)
+		}
+		return channel
+	}
+
+	setup := func(t *testing.T, selectiveSync bool) (msTeamsUser, MMUser, syntheticUser *model.User) {
+		th.ResetPreserveDB(t)
```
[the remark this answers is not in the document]
JulienTant: Oops it was just a development need, I should remove this... and/or make it better eventually.

context: I had a failing test, and at the end of the test my container was removed which was unfortunate to go see what hapenned. So I commented the lines that would remove the container + made a reset method that would preserve my DB so I can inspect it.

Ideally, it would be nice to have a way (maybe an env variable?) to instruct the test engine to not remove the database container and not do the `DELETE`s when you want to inspect what's happening but don't want to have to run tons of "debug" queries in the test 😄

server/message_hooks.go
JulienTant: I could not find a check in the code before, and I believe this is the cause of this issue: https://mattermost.atlassian.net/browse/MM-56022

server/message_hooks_test.go
JulienTant: Good call, changing the names for extra clarity.

server/message_hooks_test.go
JulienTant: changed to `${platform}UserPrimary`

server/helper_test.go
JulienTant: I intended to test specifically that this message (message as in content, not post) was not sent. I guess the name of the method is not clear enough, I wish we talked about ephemeralPost, then the word 'message' would be more clear :P 

Will rename to `assertNoEphemeralMessageWithContent`

JulienTant:
I want to make sure I understand your feedback correctly @lieut-data; do I leave my ephemeral message in place now until we pivot to unmuting Teams Primary<-->Teams Primary or do you think it's better not to implement it at all, and I clean this out of this PR but leave my other changes?

lieut-data:
> I want to make sure I understand your feedback correctly @lieut-data; do I leave my ephemeral message in place now until we pivot to unmuting Teams Primary<-->Teams Primary or do you think it's better not to implement it at all, and I clean this out of this PR but leave my other changes?

I think we ended up chatting 1:1 in Mattermost, but to close the loop: let's keep this for interactions between an MS Teams primary user with only synthetic users, as those won't sync, but otherwise remove.

lieut-data:
@JulienTant, propose we keep just the unit test coverage from this change but otherwise revert any net new messages as we work towards defining next steps for the project.

JulienTant:
WIll close this as the feature is not relevant and tests have been added since

PR #637, opened 2024-05-01
MM-58052 - Only sync if a user in the chat is remote

sbishel:
#### Summary
Updates `chatMembersSpanPlatforms" to only return true if a synthetic user (user.IsRemote()) is included. 

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-58052


server/selective_sync_test.go
```diff
@@ -67,7 +67,7 @@ func TestChatShouldSync(t *testing.T) {
 
 		chatShouldSync, appErr := th.p.ChatShouldSync(channel.Id)
 		require.Nil(t, appErr)
-		assert.False(t, chatShouldSync)
+		assert.True(t, chatShouldSync)
```
sbishel, line 70: Not sure this test even makes sense, since "remote users" are now defined as "synthetic" and synthetic users won't ever message each other from Mattermost. Leaving test, in case we change the definition back again.

server/selective_sync_test.go
```diff
@@ -144,7 +144,7 @@ func TestChatShouldSync(t *testing.T) {
...
```
sbishel, line 147: Again...all remote users.

server/selective_sync_test.go
```diff
@@ -251,7 +251,7 @@ func TestChatMembersSpanPlatforms(t *testing.T) {
 			model.ChannelMember{UserId: user2.Id},
 		})
 		require.Nil(t, appErr)
-		assert.False(t, chatShouldSync)
+		assert.True(t, chatShouldSync)
```
sbishel, line 254: Again...remote users.

server/selective_sync_test.go
```diff
@@ -261,10 +261,6 @@ func TestChatMembersSpanPlatforms(t *testing.T) {
 
 		var appErr *model.AppError
 
-		user1.RemoteId = model.NewString(th.p.remoteID)
-		user1, appErr = th.p.API.UpdateUser(user1)
-		require.Nil(t, appErr)
```
sbishel, line 266: `th.SetupRemoteUser` does this, this code isn't necessary.

server/selective_sync.go
```diff
@@ -23,13 +22,12 @@ func (p *Plugin) ChatShouldSync(channelID string) (bool, *model.AppError) {
 }
 
 // ChatMembersSpanPlatforms determines if the given channel members span both Mattermost and
-// MS Teams. Chats between users on the same platform are skipped if selective sync is enabled.
+// MS Teams. Only Synthetic users are considered MS Teams users.
+// Chats between users on the same platform are skipped if selective sync is enabled.
```
lieut-data, line 26: The last part of this comment is probably ripe for removal, given that this function is used for other purposes now. Propose:
```suggestion
// ChatMembersSpanPlatforms determines if the given channel members span both Mattermost and
// MS Teams. Synthetic users are considered to be on MS Teams, and all others on Mattermost.
```

server/selective_sync_test.go
```diff
@@ -194,7 +194,7 @@ func TestChatMembersSpanPlatforms(t *testing.T) {
 
 	t.Run("empty set of channel members", func(t *testing.T) {
 		chatMembersSpanPlatforms, appErr := th.p.ChatMembersSpanPlatforms(model.ChannelMembers{})
-		require.Nil(t, appErr)
+		require.Error(t, appErr)
```
lieut-data, line 197: I don't believe we can use `require.Error` with a function returning `*model.AppError`, because a `nil` value would still pass:
```go
package main_test

import (
	"testing"

)

func TestAppError(t *testing.T) {
	var nilAppErr *model.AppError
	require.Error(t, nilAppErr)
}
```

```
> go test .
ok      error   0.189s
```

Let's either use `require.NotNil` or better yet, change the code in `selective_sync.go` to use the `p.apiClient` instead of `p.API`, with the former managing this complexity and always returning a proper `error` interface that works with `require.Error`.

server/selective_sync_test.go
```diff
@@ -213,7 +213,7 @@ func TestChatMembersSpanPlatforms(t *testing.T) {
 	t.Run("user with empty id", func(t *testing.T) {
 		team := th.SetupTeam(t)
 		user1 := th.SetupRemoteUser(t, team)
-		chatMembersSpanPlatforms, appErr := th.p.ChatMembersSpanPlatforms(model.ChannelMembers{model.ChannelMember{UserId: user1.Id}, model.ChannelMember{UserId: ""}})
+		chatMembersSpanPlatforms, appErr := th.p.ChatMembersSpanPlatforms(model.ChannelMembers{model.ChannelMember{UserId: ""}, model.ChannelMember{UserId: user1.Id}})
```
lieut-data, line 216: Is this particular change intentional?

lieut-data (approved):
Thanks, @sbishel! Two comments below, but non-blocking from me :)

enahum (approved):

server/selective_sync_test.go
sbishel: Actually, it was. `ChatMembersSpanPlatforms` now returns as soon as a remote user is found. So in order to catch this error, it has to happen on the first user. This function is called internally, so other than a test, a member would always have an ID. I didn't think it it necessary to loop all members just to check ID. I probably should have removed this test. 

This whole function will be removed with the PR that handles disconnected users.

sbishel:
@lieut-data - Could you take a quick look at the changes for calling `apiclient` instead of `API`? More changes than I expected and I want to make sure I did it correctly, I wasn't aware of `apiclient`. 

This was more of an exercise for me, as the PR I'm currently working on handling disconnected users changes most of this code. ie. `ChannelMembersSpanPlatforms` gets replaced, that name no longer makes sense as that's not what we check anymore.

I do want to get this one merged, so I can get the other PR merged and ready. Thanks

lieut-data (approved):
Makes sense to me! Thanks, @sbishel :)

lindy65:
Thanks for QA-deferring @sbishel - we can test this at a bug bash :)

PR #639, opened 2024-05-03
Adds the transactional store generator

mgdelacroix:
#### Summary
This PR adds a generator to the Store that allows to make any method transactional in a quick and easy way, and makes store methods easily reusabe inside the store itself, allowing them to be used alone or as part of a transaction without having to adapt anything in the code. This is achieved through the following changes:
- The implementations of the methods that we want to expose through the Store interface are now private, and always receive a `db sq.BaseRunner` as the first parameter. This parameter can be either the database connection or a transaction, allowing the implementation to work in both contexts.
- Those methods that require to be run within a transaction, should be commented in the Store interface with a `@withTransaction` annotation.
- A new generator has been added, that goes through all methods in the Store interface and creates their public implementations in the `sqlstore/public_methods.go` file. The implementation will be a simple proxy if the method doesn't require to start a transaction, but if it does, the implementation will create, commit and rollback the transaction depending on the return values of the private method.


jespino (approved):
LGTM

server/store/generators/transactional_store.go.tmpl
```diff
@@ -0,0 +1,58 @@
...
+// To add a public method, create an entry in the Store interface,
+// prefix it with a @withTransaction comment if you need it to be
+// transactional and then add a private method in the store itself
+// with db sq.BaseRunner as the first parameter before running `make
+// generate`
+
+package sqlstore
+
+import (
+	"context"
```
lieut-data, line 16: nit: indent off?

server/store/sqlstore/public_methods.go
```diff
@@ -0,0 +1,342 @@
...
+	return s.mattermostToTeamsUserID(s.db, userID)
+
+}
+
+func (s *SQLStore) RecoverPost(postID string) error {
+	return s.recoverPost(s.db, postID)
+
+}
+
+func (s *SQLStore) SaveChannelSubscription(subscription storemodels.ChannelSubscription) error {
```
lieut-data, line 204: What makes these "different"?

server/store/store.go
```diff
@@ -24,6 +24,7 @@ type Store interface {
 	UserHasConnected(mmUserID string) (bool, error)
 	GetUserConnectStatus(mmUserID string) (*storemodels.UserConnectStatus, error)
 	GetHasConnectedCount() (int, error)
+	//@withTransaction
```
lieut-data, line 27: Instead of annotating the interface, is it possible to annotate the implementation instead? This is an implementation detail, doesn't apply to all implementations of the interface, and doesn't concern the caller anyway. Plus I'd hate to see a refactor accidentally move a comment away from the "line" after.

lieut-data (approved):
A few, non-blocking comments, including re: `@withTransaction`, but I'm curious why we feel the need to generate this code vs. just having explicit, inline `DoBlah()` and `doBlah()` implementations (using sq.BaseRunner: that's the key!)?

I'm (slightly) worried that we're trading clear, unambiguous code with metaprogramming and without a dramatically better outcome.

server/store/sqlstore/public_methods.go
mgdelacroix: Those are the ones that run multiple DB queries, so they get their transactions managed

server/store/store.go
mgdelacroix: Totally agree with the reasoning here. I'll give it a try as it's a sizeable change

mgdelacroix:
> A few, non-blocking comments, including re: `@withTransaction`, but I'm curious why we feel the need to generate this code vs. just having explicit, inline `DoBlah()` and `doBlah()` implementations (using `sq.BaseRunner`: that's the key!)?
> 
> I'm (slightly) worried that we're trading clear, unambiguous code with metaprogramming and without a dramatically better outcome.

The end result is still explicit, although all the generated public methods end in a specific `public_methods.go` file and not right above their private implementations. The key reason to use a generator is that all methods need to have their public/private pair for the store to be callable internally and fully transactional, and that's going to be a lot of repeated code if we do it inline, even if it's just 90% proxy methods that don't require a transaction.

I don't think generating those adds that much confusion really, and there is no way to bypass this without understanding it (a developer would need to add a method to the generator's skip list). Can you think of a specific concern or use case that you see as specially confusing?

server/store/sqlstore/public_methods.go
lieut-data: Ah, yes, somehow I misread this -- that's obvious from context now, thank you!

lieut-data:
> Can you think of a specific concern or use case that you see as specially confusing?

Nothing concrete yet! I think your explanation above makes sense, and I'm less concerned if we can (eventually) move the annotation to the implementation. Thank you!

mgdelacroix:
@jespino @lieut-data requesting re-review as I've moved the annotations to golang `db:withTransaction` directives in the private method implementation and added the `db:withReplica` to use database replicas instead of the master node when possible

lieut-data (approved):
This is great, thank you @mgdelacroix!

sbishel (approved):
Nice!

PR #643, opened 2024-05-07
[MM-58155] Separate DM/GM option

JulienTant:
#### Summary
This PR separates the "Enable DM/GM" option into "Enable DM" + "Enable GM", with all the supporting code.

##### Idea behind how it's done

I kept the same code path mostly because, at least for now, we are treating DMs and GMs the same way.

I renamed a lot of `isDirectMessage` to `isDirectOrGroupMessage` to make it clearer that it could be both now that there are two options.

##### Though/questions

We do have a tag `is_direct` in a lot of our prometheus metrics, not quite sure if we should rename those, OR only set to true if the message is an actual DM (exclude GM from this tag). ANd do we add a `is_gm` tag?

##### Demo

Video of the PR in action:

https://github.com/mattermost/mattermost-plugin-msteams/assets/785518/7bec4c3e-aedd-4d74-80d5-c437f4ae8c8d

#### Ticket Link
[https://mattermost.atlassian.net/browse/MM-58155](https://github.com/mattermost/mattermost-plugin-msteams/pull/641)


jespino (approved):
Nice, LGTM.

lindy65 (approved):
Thanks @JulienTant - I've QA deferred and we'll test this at a bug bash after merge

PR #646, opened 2024-05-09
Rewrite message hooks tests: reactions

lieut-data:
#### Summary
Continuing to adopt the new framework, this time focusing on coverage for `ReactionHasBeenAdded` and `ReactionHasBeenRemoved`.

As part of these changes, I abandoned the previous optimization to skip setting up the test framework from `main_test.go` until the first test that uses it. There is a significant downside to doing so for the first test: all logs from the setup are incorrectly associated with that test, making it harder to trace what happened. Since we're migrating everything anyway, it doesn't seem necessary to keep the optimization unto itself.

#### Ticket Link
None.

### DIFF
--- a/server/message_hooks.go
+++ b/server/message_hooks.go
@@ -195,6 +195,10 @@ func (p *Plugin) ReactionHasBeenAdded(c *plugin.Context, reaction *model.Reactio
 		return
 	}
 
+	if !p.getConfiguration().SyncLinkedChannels {
+		return
+	}
+
 	post, appErr := p.API.GetPost(reaction.PostId)
 	if appErr != nil {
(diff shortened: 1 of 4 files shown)


server/message_hooks.go
```diff
@@ -195,6 +195,10 @@ func (p *Plugin) ReactionHasBeenAdded(c *plugin.Context, reaction *model.Reactio
 		return
 	}
 
+	if !p.getConfiguration().SyncLinkedChannels {
+		return
+	}
```
lieut-data, line 200: We previously didn't honour this setting for reactions in a channel: now we do!

calebroseland (approved):

sbishel (approved):
Great. I needed this one. :)

PR #652, opened 2024-05-15
Sync remote config

sbishel:
#### Summary
 - Remove sync'ing for disconnected users.
 - Support both Remote User Sync and Primary Platform sync via configuration flag.
 - Updates automute to work with these settings as well

There are three different sync'ing options with DM/GMs
- Nothing Selected
  - Sync's all DM/GMs
- Selective Sync
  - Sync's DM/GM that "Span Platforms"
- Selective Sync Remote Only
  - Sync's DM/GM between remote and connected users

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-58152
  Fixes https://mattermost.atlassian.net/browse/MM-58053


sbishel (commented):
If the automute logic is correct, then more tests will need to be added for automute. One issue with automute is if you change selective sync settings and un-mute, we may not unmute the correct channels. As unmute and mute use the same logic.

Switching selective sync setting may also cause issues with regular messages, ie. attempt to update a message that wasn't sync'ed when created.

server/plugin.go
```diff
@@ -888,6 +893,15 @@ func (p *Plugin) IsRemoteUser(user *model.User) bool {
 	return user.RemoteId != nil && *user.RemoteId == p.remoteID
 }
 
+func (p *Plugin) IsUserConnected(userID string) (bool, error) {
+	token, err := p.store.GetTokenForMattermostUser(userID)
+	if err != nil && err != sql.ErrNoRows {
```
JulienTant, line 898: In the store, if the user has no encrypted token, we return an error:

https://github.com/mattermost/mattermost-plugin-msteams/blob/ae3640f39f667d0d03134cf46065da7124a307f8/server/store/sqlstore/store.go#L364-L366

This is a bit weird, I think we should either change the store to return `nil, nil` (no token but no error), or we should make it a proper `ErrNoTokenFound`, and this method could check if `errors.Is(err, ErrNoTokenFound)` to return false rather than an error.

JulienTant (changes requested):
Nothing really on the feature side of the PR, it looks good even tho I start to want to print a cheat sheet of all the combinations 🤣 

Mostly some dev logs left, and a remark about `IsRemoteUser` returning an error when the user has no token

plugin.json
```diff
@@ -106,6 +106,13 @@
                 "help_text": "Skip syncing messages between users on the same platform.",
                 "default": false
             },
+            {
+                "key": "syncRemoteOnly",
+                "display_name": "Selective sync - Remote Only",
+                "type": "bool",
+                "help_text": "Selective Syncing only syncs with connected and remote users.",
+                "default": false
+            },
```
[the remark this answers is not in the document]
JulienTant: I think our only option would be to develop plugin side option config in the webapp.. ugh 👎🏻

server/handlers/handlers.go
```diff
@@ -349,6 +350,14 @@ func (ah *ActivityHandler) handleCreatedActivity(msg *clientmodels.Message, subs
 		senderID = ah.plugin.GetBotUserID()
 	}
 
+	if ah.plugin.GetSyncRemoteOnly() {
```
lieut-data, line 353: Should this be "inside" the check for `plugin.SelectiveSync()` so that it only applies when the latter is enabled?

server/handlers/handlers.go
```diff
@@ -616,6 +625,27 @@ func (ah *ActivityHandler) handleReactions(postID, channelID string, isDirectOrG
 }
 
 func (ah *ActivityHandler) handleDeletedActivity(activityIds clientmodels.ActivityIds) string {
+	if ah.plugin.GetSelectiveSync() {
+		_, chat, err := ah.getMessageAndChatFromActivityIds(&clientmodels.Message{}, activityIds)
```
lieut-data, line 629: If we need to do this at all, can we just look up the channel information from the post info we already get below? Doesn't seem necessary to make a client call to look it up a different way.

server/handlers/handlers.go
```diff
@@ -616,6 +625,27 @@ func (ah *ActivityHandler) handleReactions(postID, channelID string, isDirectOrG
 }
 
 func (ah *ActivityHandler) handleDeletedActivity(activityIds clientmodels.ActivityIds) string {
+	if ah.plugin.GetSelectiveSync() {
```
lieut-data, line 628: It looks like this was a gap in selective sync before, in that with the feature off we would still have deleted messages that had previously been synced. I'm wondering if, for simplicity, we "don't solve this" and leave as-is?

server/message_hooks.go
```diff
@@ -119,26 +119,24 @@ func (p *Plugin) MessageHasBeenPosted(_ *plugin.Context, post *model.Post) {
...
-				p.API.LogWarn("Failed to check if chat members span platforms", "error", err.Error(), "post_id", post.Id, "channel_id", post.ChannelId)
-				return
-			}
-
-			if p.getConfiguration().SelectiveSync && !chatMembersSpanPlatforms {
-				return
+		chatShouldSync := false
+		if p.getConfiguration().SelectiveSync {
+			isSelfPost := len(members) == 1
+			if !isSelfPost {
```
lieut-data, line 125: I'm wondering if/why we need to distinguish self chats anymore? Can you elaborate?

server/selective_sync.go
```diff
@@ -3,41 +3,129 @@ package main
...
 	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/mattermost/mattermost/server/public/shared/mlog"
+	"github.com/pkg/errors"
 )
 
-// ChatShouldSync determines if the members of the given channel span both Mattermost and
-// MS Teams. Chats between users on the same platform are skipped if selective sync is enabled.
-// Chats with only a single member are self chats and always sync.
-func (p *Plugin) ChatShouldSync(channelID string) (bool, error) {
+func (p *Plugin) ChannelShouldSyncCreated(channelID, senderID string) (bool, error) {
```
lieut-data, line 11: Why the `Created` suffix? Are the semantics unique to creation vs other events?

server/plugin.go
sbishel: I updated to return 'nil, nil', as most places that call this function don't even check the error that is returned, and it didn't cause a bunch of unit test failures. If we want to return a proper, `ErrNoTokenFound`, we should probably handle that in a separate PR.

lieut-data (changes requested):
Thanks, @sbishel! A few comments below, but really great work teasing this all apart.

server/automute.go
```diff
@@ -158,23 +149,39 @@ func (p *Plugin) canAutomuteChannelID(channelID string) (bool, error) {
...
+				// do not automute any DM/GM channels,
+				// in this mode all uses considered MM primary
 				return false, nil
 			}
+			// if Selective Sync, automute if members
+			// span platforms
+			return p.ChannelShouldSync(channel.Id)
+		}
+
+		// if not selective sync
```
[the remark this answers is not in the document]
sbishel: Yep...that's why I asked...the original implementation was not wrapped in 
```
if p.configuration.SelectiveSync {
```

server/message_hooks.go
sbishel: removed...not sure when/if self chats should sync.

server/selective_sync.go
sbishel: With GetSyncRemoteOnly on, the created event needs to ensure the Sender is either "Connected" or "Remote". Other events don't need this check, so this was me trying to not duplicate code and  have a single place that requires slightly different functionality. I wanted to try to keep all the logic in `selective_sync.go`. I could check the connection status in the handler that makes this call.

Willing to rename, move or change login.

server/handlers/handlers.go
sbishel: Ok..I'll leave as is.

server/selective_sync.go
sbishel: `ChannelShouldSync` gets called from other places that doesn't have a `senderID`. This same code is used to determine automute. I originally wrote this as a single function and passed in "" for sender where we didn't have or need it. That to me was more confusing. I also tried doing the "connection" check outside "ChannelShouldSync", but that resulted in duplicating API calls, such as calling `ListMembers` twice. I also think this separates the code, so parts can be removed/changed, and keeps all the "selective sync" decisions in a single place.

sbishel:
Git handled the handlers relocation merge pretty well. A few straight-forward conflicts, but no issues.

JulienTant (approved):
Thanks @sbishel !

JulienTant (approved):
Thanks @sbishel !

server/automute.go
```diff
@@ -151,24 +141,24 @@ func (p *Plugin) canAutomuteChannelID(channelID string) (bool, error) {
...
-			if appErr != nil {
-				return false, errors.Wrap(appErr, fmt.Sprintf("Unable to get user for channel member %s ", userID))
-			}
-			if user.IsBot || user.IsGuest() {
+	if channel.IsGroupOrDirect() {
+		if p.configuration.SelectiveSync {
+			if p.configuration.SyncRemoteOnly {
+				// if only sync'ing with remote users,
+				// do not automute any DM/GM channels,
+				// in this mode all uses considered MM primary
```
lieut-data, line 152:
```suggestion
				// in this mode all users considered MM primary
```

server/automute.go
```diff
@@ -151,24 +141,24 @@ func (p *Plugin) canAutomuteChannelID(channelID string) (bool, error) {
...
+			// span platforms
+			return p.ChannelShouldSync(channel.Id)
 		}
-		return true, nil
+
+		// if not selective sync
+		return false, nil
 	}
 
 	link, err := p.store.GetLinkByChannelID(channel.Id)
```
lieut-data, line 164: Should we check `p.configuration.SelectiveSync` at the top of this method? Otherwise, it looks like we can returne `true` here for linked channels even when selective sync is false.

server/selective_sync.go
```diff
@@ -35,9 +51,76 @@ func (p *Plugin) ChatMembersSpanPlatforms(members []*model.ChannelMember) (bool,
...
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (p *Plugin) ChannelConnectedOrRemote(channelID, senderID string) (bool, error) {
+	senderConnected, err := p.IsUserConnected(senderID)
+	if err != nil {
+		// ah.plugin.GetAPI().LogWarn("Unable to determine if sender is connected", "error", err.Error())
```
lieut-data, line 73: Should we keep this log?

lieut-data (changes requested):
Thanks, @sbishel! Just a few extra comments below.

server/automute.go
sbishel: @lieut-data  Sorry, I don't understand this comment. `p.configuration.SelectiveSync` is checked above if a GM or DM. And previous comment was to not to automute unless SS was enabled. 

> Commented above, but I thought we didn't automute /unless/ selective sync was enabled. So maybe we can simplify?

I did not think "linked channels" had anything to do with Selective sync. The logic around linked channels has not changed.

server/selective_sync.go
sbishel: It gets logged in the caller `handleCreatedActivity`, so no not needed, but I did remove the commented out line.

server/automute.go
lieut-data: My mistake: I think I conflated the automute/Teams primary functionality with selective sync, but the latter came after. This is definitely ripe for consolidation -- discard my feedback for now, and we can revisit later.

lieut-data (approved):

lindy65:
Thanks for QA-deferring @sbishel - we can test this after merging

PR #654, opened 2024-05-17
Prototype chat notifications

lieut-data:
#### Summary
This is a POC for chat notifications, enabled with a feature flag / plugin configuration. When enabled, two-way sync is completely disabled (as the semantics are otherwise not well defined). I'll follow up on this change with some refactoring and unit tests as we solidify this.

(This doesn't completely resolve https://mattermost.atlassian.net/browse/MM-58297, as there are styling concerns there, but I wanted to get a prototype up for review on which we can flesh out next steps.)

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-58289
Relates-to: https://mattermost.atlassian.net/browse/MM-58297


JulienTant (changes requested):
test

server/handlers/handlers.go
```diff
@@ -423,6 +430,10 @@ func (ah *ActivityHandler) handleCreatedActivity(msg *clientmodels.Message, subs
 }
 
 func (ah *ActivityHandler) handleUpdatedActivity(msg *clientmodels.Message, subscriptionID string, activityIds clientmodels.ActivityIds) string {
+	if ah.plugin.GetSyncNotifications() {
```
JulienTant, line 433: nit: should we consider updating the quote in the bot message?

server/handlers/handlers.go
```diff
@@ -616,6 +627,10 @@ func (ah *ActivityHandler) handleReactions(postID, channelID string, isDirectOrG
 }
 
 func (ah *ActivityHandler) handleDeletedActivity(activityIds clientmodels.ActivityIds) string {
+	if ah.plugin.GetSyncNotifications() {
```
JulienTant, line 630: nit: should we update the notification to let the user know the message has been removed? The permalink we posted will lead nowhere.

server/handlers/notifications.go
```diff
@@ -0,0 +1,96 @@
...
+	post, _, _ := ah.msgToPost("", botUserID, msg, chat, false, []string{})
+
+	if chat == nil {
+		// We're only going to support notifications from chats for now.
+		return metrics.DiscardedReasonChannelNotificationsUnsupported
+	}
+
+	for _, member := range chat.Members {
+		// Don't notify senders about their own posts.
+		if member.UserID == msg.UserID {
```
JulienTant, line 22: nit: Should we have an exception for self messaging?

JulienTant (approved):
Left a few nits, that are in fact suggestions for new tickets now that I think about them 😅 But LGTM overall

server/handlers/notifications.go
lieut-data: I'd advocate for silencing self-messaging here -- maaaaybe useful for testing, but super annoying if Teams users use their own chats like we sometimes do on Mattermost as a scratch pad.

server/handlers/handlers.go
lieut-data: Yes, we do need to define the behaviour here -- will track as a task.

server/handlers/handlers.go
lieut-data: Yes, we do need to define the behaviour here -- will track as a task.

sbishel (approved):

PR #655, opened 2024-05-17
Collapse handlers into root

lieut-data:
#### Summary
This change collapses the `handlers` package into the root server package. If you take a look at the [interface](https://github.com/mattermost/mattermost-plugin-msteams/blob/ae3640f39f667d0d03134cf46065da7124a307f8/server/handlers/handlers.go#L36-L59) we require to make this package work, it's a close approximation of the `Plugin` itself, suggesting we have tight coupling anyway without the benefits of at least being able to move quickly.

Moving the code into the root package allows us to iterate faster and eventually carve out a better package organization strategy. In addition to moving the code, this change consolidates our mock testing strategy to be consistent with the soon-to-be-deprecated `newTestPlugin` approach. While this is "throw away" in a sense, it actually helps accelerate that transition.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-58320


JulienTant (approved):
Yay! I restarted the CI as it fails on testcontainers failing to start a container

sbishel (approved):

PR #656, opened 2024-05-17
[MM-58322] Add notification preference slash command

JulienTant:
#### Summary
This PR introduces the preference for the new notification feature. the UI will come in a future PR (probably right after this is merged), we started with the slash command to get something quickly.

- adds a slash command to allow the user to configure it.
- add methods to set/get the preference for the user

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-58322

### DIFF
--- a/server/automute.go
+++ b/server/automute.go
@@ -113,14 +113,7 @@ func (p *Plugin) getAutomuteIsEnabledForUser(userID string) bool {
 
 // setAutomuteIsEnabledForUser sets a preference to track if we've muted all of the user's linked channels.
 func (p *Plugin) setAutomuteIsEnabledForUser(userID string, channelsAutomuted bool) error {
-	appErr := p.API.UpdatePreferencesForUser(userID, []model.Preference{
-		{
-			UserId:   userID,
-			Category: PreferenceCategoryPlugin,
-			Name:     PreferenceNameAutomuteEnabled,
-			Value:    strconv.FormatBool(channelsAutomuted),
-		},
-	})
+	appErr := p.updatePreferenceForUser(userID, PreferenceNameAutomuteEnabled, strconv.FormatBool(channelsAutomuted))
 	if appErr != nil {
 		return errors.Wrap(appErr, fmt.Sprintf("Unable to set preference to track that channels are automuted for user %s", userID))
 	}
--- a/server/command.go
+++ b/server/command.go
@@ -98,6 +98,14 @@ func getAutocompleteData(syncLinkedChannels bool) *model.AutocompleteData {
 	promoteUser.RoleID = model.SystemAdminRoleId
 	cmd.AddCommand(promoteUser)
 
+	notification := model.NewAutocompleteData("notifications", "", "Enable or disable notifications from MSTeams. You must be connected to perform this action.")
+	notification.AddStaticListArgument("status", true, []model.AutocompleteListItem{
+		{Item: "status", HelpText: "Show current notification status"},
+		{Item: "on", HelpText: "Enable notifications"},
+		{Item: "off", HelpText: "Disable notifications"},
+	})
+	cmd.AddCommand(notification)
+
 	return cmd
 }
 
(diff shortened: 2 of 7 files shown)


server/automute.go
```diff
@@ -113,14 +113,7 @@ func (p *Plugin) getAutomuteIsEnabledForUser(userID string) bool {
...
 func (p *Plugin) setAutomuteIsEnabledForUser(userID string, channelsAutomuted bool) error {
-	appErr := p.API.UpdatePreferencesForUser(userID, []model.Preference{
-		{
-			UserId:   userID,
-			Category: PreferenceCategoryPlugin,
-			Name:     PreferenceNameAutomuteEnabled,
-			Value:    strconv.FormatBool(channelsAutomuted),
-		},
-	})
+	appErr := p.updatePreferenceForUser(userID, PreferenceNameAutomuteEnabled, strconv.FormatBool(channelsAutomuted))
```
JulienTant, line 116: offtopic: I really dislike the syntax of `API.UpdatePreferencesForUser(...)` so I made a helper for the plugin

server/command.go
```diff
@@ -21,14 +21,19 @@ func (p *Plugin) createCommand(syncLinkedChannels bool) *model.Command {
 		p.API.LogWarn("Unable to get the MS Teams icon for the slash command")
 	}
 
+	autoCompleteData := getAutocompleteData(syncLinkedChannels)
```
JulienTant, line 24: offtopic: Every time we added a command we had to update a string with the new command, which is not ideal, but as a bonus we had to do it twice 😅 The other part of this change is marked by a ⭐!

server/command.go
```diff
@@ -159,10 +172,16 @@ func (p *Plugin) ExecuteCommand(_ *plugin.Context, args *model.CommandArgs) (*mo
...
+		return p.executeNotificationsCommand(args, parameters)
 	}
-	return p.cmdError(args, "Unknown command. Valid options: connect, connect-bot, status, disconnect, disconnect-bot and promote.")
+
+	const separator = ", "
+	list := strings.Join(p.subCommands, separator)
+	if i := strings.LastIndex(list, separator); i != -1 {
+		list = list[:i] + " and " + list[i+2:]
+	}
+	return p.cmdError(args, "Unknown command. Valid options: "+list)
```
JulienTant, line 184: ⭐

server/command.go
```diff
@@ -21,14 +21,19 @@ func (p *Plugin) createCommand(syncLinkedChannels bool) *model.Command {
 		p.API.LogWarn("Unable to get the MS Teams icon for the slash command")
 	}
 
+	autoCompleteData := getAutocompleteData(syncLinkedChannels)
+	for i := range autoCompleteData.SubCommands {
+		p.subCommands = append(p.subCommands, autoCompleteData.SubCommands[i].Trigger)
```
lieut-data, line 26: To confirm, do we only ever call `createCommand ` once? Otherwise, this write to `p.subCommands` might be racy, and I wonder if we should just resolve it on demand vs trying to cache?

server/store/storemodels/storemodels.go
```diff
@@ -9,6 +9,11 @@ const (
 	PreferenceNamePlatform         = "platform"
 	PreferenceValuePlatformMM      = "mattermost"
 	PreferenceValuePlatformMSTeams = "msteams"
+
+	// Notification preferences
+	PreferenceNameNotification     = "notifications"
+	PreferenceValueNotificationOn  = "1"
```
lieut-data, line 15: nit: why `1` and `0` instead of `on` and `off`?

lieut-data (changes requested):
Thanks, @JulienTant! Just a few minor comments.

server/command.go
```diff
@@ -159,10 +172,16 @@ func (p *Plugin) ExecuteCommand(_ *plugin.Context, args *model.CommandArgs) (*mo
...
+	if action == "notifications" {
+		return p.executeNotificationsCommand(args, parameters)
 	}
-	return p.cmdError(args, "Unknown command. Valid options: connect, connect-bot, status, disconnect, disconnect-bot and promote.")
+
+	const separator = ", "
+	list := strings.Join(p.subCommands, separator)
+	if i := strings.LastIndex(list, separator); i != -1 {
+		list = list[:i] + " and " + list[i+2:]
+	}
```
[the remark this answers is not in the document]
JulienTant: We call it on onActivate but also when the configuration changes. I can add a lock tho, better be safe than sorry.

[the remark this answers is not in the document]
JulienTant: changed it!

server/command.go
JulienTant: We call it on onActivate but also when the configuration changes. I can add a lock tho, better be safe than sorry.

server/preferences.go
```diff
@@ -28,19 +32,39 @@ func (p *Plugin) setPrimaryPlatform(userID string, primaryPlatform string) error
...
+	return pref.Value == storemodels.PreferenceValueNotificationOn
+}
+
+func (p *Plugin) setNotificationPreference(userID string, enable bool) error {
+	value := storemodels.PreferenceValueNotificationOff
+	if enable {
+		value = storemodels.PreferenceValueNotificationOn
+	}
+
+	err := p.updatePreferenceForUser(userID, storemodels.PreferenceNameNotification, value)
```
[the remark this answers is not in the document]
JulienTant: changed to appErr. 

And I agree, `Preferences` are missimg from the `pluginapi`! I actually wondered why we used this rather than the KV store, but I suppose we get access to preferences from redux so it's easier for frontend components.

server/store/storemodels/storemodels.go
JulienTant: I swear I saw some preference using `1`/`0`  but can't find them anymore haha, will switch to `on`/`off` which is what I had initially.

lieut-data (approved):

sbishel (approved):
Minor comment, question

server/store/storemodels/storemodels.go
```diff
@@ -9,6 +9,11 @@ const (
...
+	PreferenceValueNotificationOff = "0"
```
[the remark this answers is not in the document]
sbishel: Ha...no longer valid...

PR #657, opened 2024-05-18
[MM-58290] Add UI to select notification preference.

JulienTant:
#### Summary
This PR:

- Adds in the redux store info about the user current connection status.
- Adds the notifications preference in the user settings.
- Use the new user current connection status to dynamically enable/disable the user preference options.
- change some of the `connect` api logic to allow the connect button in the preference to work!

What is not in this PR be worth considering for future PRs?
- Enabling/Disabling each configuration based on the current feature set of the plugin. Example: if selective sync is disabled, do we want to also disable (or not show) platform selection?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58290


JulienTant:
@lieut-data Before working on the UI for the notification preference, I worked on having the connection status of the user in redux. Can you take a look and let me know if you see something I should fix at this stage?

webapp/src/actions.ts
```diff
@@ -0,0 +1,9 @@
+import * as Actions from './types/actions';
+
+export const userHasConnected = (): Actions.UserHasConnected => ({
+    type: Actions.USER_HAS_CONNECTED,
```
lieut-data, line 4: nit: the websocket event is `user_connected` but here we model with `has` -- might we converge on the same phrase just so a case insensitive grep is likely to find all related matches?

lieut-data (approved):
Wow, this is phenomenal! Nice work, @JulienTant -- just one thought below, but non-blocking.

webapp/src/actions.ts
JulienTant: Yeah, I'll drop the `has` everywhere it's actually bothering me to have it!

server/api.go
```diff
@@ -347,10 +349,16 @@ func (a *API) connect(w http.ResponseWriter, r *http.Request) {
 		userID = a.p.GetBotUserID()
 	}
 
+	stateSuffix := ""
```
JulienTant, line 352: The code prevented the "connect" button from the preference page to function, because it required a bot message. I changed the code to allow for both.

server/api.go
```diff
@@ -547,6 +599,10 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	a.p.API.PublishWebSocketEvent(WSEventUserConnected, map[string]any{}, &model.WebsocketBroadcast{
```
JulienTant, line 602: except for this line, all the changes in this method are related to the change allowing the "connect" button to work 😅

webapp/src/index.tsx
```diff
@@ -1,85 +1,57 @@
...
-
+import reducer from './reducer';
 // eslint-disable-next-line import/no-unresolved
 import {PluginRegistry} from './types/mattermost-webapp';
 import {getServerRoute} from './selectors';
 
 const MINUTE = 60 * 1000;
 const randomInt = (max: number) => Math.floor(Math.random() * max);
 
-function getSettings(serverRoute: string, disabled: boolean) {
```
JulienTant, line 19: just moved this to its own file.

lieut-data (approved):
Nice! This looks great -- thank you again, @JulienTant :)

sbishel (approved):
Sorry, I thought I had already approved this. Looks good.

PR #659, opened 2024-05-21
[MM-58298] Add notification welcome message

JulienTant:
#### Summary
Add Notification welcome message.

I also added a better confirmation page for the case we don't go to platform selection:

image

Better than the plain black text on the top left telling you to close the window 🤣 Also applies to connecting a bot

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58298


server/main_test.go
```diff
@@ -135,6 +135,7 @@ func setupServer(mt *mainT) error {
 	config.LocalizationSettings.SetDefaults()
 	config.SqlSettings = *storetest.MakeSqlSettings("postgres", false)
 	config.ServiceSettings.SiteURL = model.NewString("http://example.com/")
+	config.ServiceSettings.AllowedUntrustedInternalConnections = model.NewString("localhost")
```
JulienTant, line 138: This allows MM image proxy to pull the gif from the plugin

lieut-data:
This is awesome, thank you @JulienTant! In regards to the welcome page, do you want to see if we can lift the styling proposed by https://mattermost.atlassian.net/browse/MM-58345? (The desktop link might be something to defer to a separate ticket, as I'm unsure how that would work.)

server/api.go
```diff
@@ -621,15 +653,27 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
...
+	switch {
+	case a.p.getConfiguration().SyncNotifications:
+		a.handleSyncNotificationsWelcomeMessage(originInfo[0], mmUserID, postID)
+		http.Redirect(w, r, a.p.GetURL()+"/account-connected", http.StatusSeeOther)
+		return
+	default:
+		a.handleDefaultWelcomeMessage(originInfo[0], mmUserID, postID, channelID)
+		http.Redirect(w, r, a.p.GetURL()+"/primary-platform", http.StatusSeeOther)
+		return
+	}
```
sbishel, line 671: Not, necessarily as part of this PR, but should we have checks here for "RemoteOnlySync" and NOT "SelectiveSync" since those won't require selecting a PrimaryPlatform.

sbishel (approved):
Looks good. One comment on potential future change.

JulienTant:
@lieut-data I'll steal the background from the figma and leave the Close button as-is for now, so we can check later how the "Open in mattermost" link should work.

server/api.go
JulienTant: I will redirect to `/primary-platforn` only if SelectiveSync is true!

server/api.go
```diff
@@ -404,6 +406,36 @@ func (a *API) connectionStatus(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (a *API) accountConnectedPage(w http.ResponseWriter, r *http.Request) {
+	message := "You are now now connected."
+	query := r.URL.Query()
+	if query.Has("isBot") {
+		message = "The bot account's is now connected."
```
lieut-data, line 413:
```suggestion
		message = "The bot account is now connected."
```

server/bot_messages.go
```diff
@@ -93,3 +99,42 @@ func (p *Plugin) notifyUserTeamsPrimary(userID string) {
...
+
+	return &model.Post{
+		Message: strings.Join(msg, "\n\n"),
+		Props: model.StringInterface{
+			"attachments": []*model.SlackAttachment{
+				{
+					Actions: []*model.PostAction{
+						{
+							Integration: &model.PostActionIntegration{
+								URL: fmt.Sprintf("%s/enable-notifications", p.GetURL()),
```
lieut-data, line 129: Can we avoid the fully qualified URL (and the image proxy issues) by just using a relative URL instead? (Taking into account any subpath, of course.)

server/main_test.go
lieut-data: See above, but if true this might be an issue for installations where the URL changes as well, so we should use a relative path instead.

lieut-data (changes requested):
Thanks, @JulienTant! A few comments below.

server/bot_messages.go
JulienTant: I applied some changes to use the relative URL, seems to be working fine for both the gif and the button action. I was afraid I would get rejected because not starting with `http{s}://`.

off-topic, but I wish we had a "hooky" way to deal with this, so I could declare my 

```gp
post := &model.Post{
		Message: strings.Join(msg, "\n\n"),
		Props: model.StringInterface{
			"attachments": []*model.SlackAttachment{
				{
					Actions: []*model.PostAction{
						{
							ID: "enable_notifications", // <-- change here
							Integration: {Plugin: pluginID}, // <-- change here
							Name:  "Enable Notifications",
							Style: "primary",
							Type:  model.PostActionTypeButton,
						},
					},
				},
			},
		},
	}
}
```

and in my plugin:

```go
OnPostAction(id string, data model.PostActionRequest) {
	switch id {
	// ....
	}
}
```

So many ideas, so little time to do it all right now :D

lieut-data (approved):
Nice!

PR #668, opened 2024-05-23
{chore) Update testcontainers

JulienTant:
#### Summary
Dependabot made 3 PRs to update each testcontainers dependencies we have, but they all break because of a breaking change introduced in `0.31.0`

This PR aims to upgrade testcontainers and fix the BC.

#### Ticket Link
N/A


lieut-data (approved):
Thank you!

PR #673, opened 2024-05-27
rm -fr Teams Primary preference

lieut-data:
#### Summary
Rip out all things re: Teams Primary & automute semantics. Our test coverage made me feel much more comfortable doing this, with smoke testing confirming general functionality remaining intact.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-58439


sbishel (commented):
Don't we still need this when SelectiveSync = true and SyncRemoteOnly = false. I thought we wanted to keep this functionality, which is why we added SyncRemoteOnly as a config setting.

sbishel (approved):
> Selective sync as a concept still makes sense without this feature, but I suspect we can simplify SyncRemoteOnly and SelectiveSync further to remove even more code.

I don't think SyncRemoteOnly is even necessary any longer.

JulienTant (approved):
LGTM!
0/5: making a cleanup migration to remove the preferences set in the DB. Would be nice if the plugin was in production but probably not a big issue right now.

Also not sure if we should also remove the  SyncRemoteOnly because that's all we do now anyway?

*edit: oh I see there's an active discussion about that right now :D*

sbishel:
> @sbishel, let's double check with @esethna, but maybe we rebrand `SelectiveSync` as what `SyncRemoteOnly` means (eliminationg the flag, but not the functionality)? I'm unsure though :)

Yeah...that's what I meant... SelectiveSync == SyncRemoteOnly.

lieut-data:
> making a cleanup migration to remove the preferences set in the DB. Would be nice if the plugin was in production but probably not a big issue right now.

@JulienTant, I did this manually on msteams-sync-test, and will probably suggest doing the same for customers migrating to this version. In principle, we /could/ add a migration to achieve this, but given we're still pre-production, I'm inclined to avoid the extra churn. Thoughts?

JulienTant:
@lieut-data 

>  I'm inclined to avoid the extra churn. Thoughts?

I agree that's why I was 0/5; I thought I'd mention it just in case it's something you wanted to add but I also don't think it's necessary right now.

esethna:
> maybe we rebrand SelectiveSync as what SyncRemoteOnly means

> Yeah...that's what I meant... SelectiveSync == SyncRemoteOnly.

+1 to this.

PR #675, opened 2024-05-28
MM-56924: Connection invite metrics

calebroseland:
#### Summary
- Adds plugins metric sources:
  - Current pending invites
  - Pending invites limit
  - Current whitelist count 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56924


webapp/src/index.tsx
```diff
@@ -72,6 +72,18 @@ export default class Plugin {
...
+                        id: 'msteams_invited_users',
+                        icon: 'fa-users', // font-awesome-4.7.0 handler
+                        value: siteStats?.pending_invited_users || 0,
+                    },
+                    msteams_whitelisted_users: {
+                        name: 'MS Teams: Whitelisted Users',
+                        id: 'msteams_whitelisted_users',
+                        icon: 'fa-users', // font-awesome-4.7.0 handler
+                        value: siteStats?.current_whitelist_users || 0,
+                    },
```
lieut-data, line 86: I'm wondering if we should consider keeping this as a metric / Grafana dashboard only for now, vs. exposing to the enduser? It's the kind of data that's transiently useful (vs. connected count, which shows volume / adoption).

server/metrics/metrics.go
```diff
@@ -363,6 +363,45 @@ func NewMetrics(info InstanceInfo) Metrics {
 	})
 	m.registry.MustRegister(m.connectedUsers)
 
+	m.connectedUsersLimit = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   MetricsNamespace,
+		Subsystem:   MetricsSubsystemApp,
+		Name:        "whitelist_limit", // really: connected_users_limit
```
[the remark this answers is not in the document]
lieut-data: Feel free to rename! Easy to update dashboards :)

server/metrics/metrics.go
```diff
@@ -354,6 +353,7 @@ func NewMetrics(info InstanceInfo) Metrics {
 	}, []string{"action"})
 	m.registry.MustRegister(m.syncMsgFileDelayTime)
 
+	// connected, invited, whitelisted
```
lieut-data, line 356: Does this mean `connected + invited + whitelisted`? I'd like to make sure we can still measure actual connections separately.

lieut-data (changes requested):
Looks good! Just a few comments for discussion before approving.

server/metrics/metrics.go
calebroseland: No, that's just a section header (ending on `:403`). Each metric is still separate.

webapp/src/index.tsx
calebroseland: I'm perfectly happy to remove them, but I do think they provide some value in monitoring or any potential troubleshooting during the initial rollout. 1/5

calebroseland: We could leave them as "extra" stats in the api request, but then just not show them in the site statistics page.

lieut-data (approved):

PR #688, opened 2024-06-04
MM-58438: simplify settings

lieut-data:
#### Summary
We're simplifying the plugin configuration to focus users towards the GA functionality for v10:
* Remove the untested `enabledTeams` setting altogether
* Remove `syncReactions` and always sync reactions (as appropriate)
* Remove `syncFileAttachments` and always sync file attachments (as appropriate)
* Consolidate `syncDirectMessages` and `syncGroupMessages` under a new `experimentalSyncChat` 
* Rename `selectiveSync` as `experimentalSelectiveSync`
* * Rename `certificatepublic` as `experimentalcertificatepublic` and `certificatekey` as `experimentalcertificatekey`.
* Rename `disableSyncMsg` as `experimentalUseSharedChannels` so that it defaults off.
* Rename `disableCheckCredentials` as `internalDisableCheckCredentials`
* Remove all internal and experimental options from `plugin.json` to hide them from the system console.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-58438


JulienTant (approved):
LGTM, I'm hesitant to say it's a good idea to re-merge DM and GM tho, I think our DM experience at this point is better than experimental but our GMs are not really at the same level.

lieut-data:
> LGTM, I'm hesitant to say it's a good idea to re-merge DM and GM tho, I think our DM experience at this point is better than experimental but our GMs are not really at the same level.

I agree that GMs introduce the most experimental level of functionality, but the overhead to enable one without the other is non-trivial. Syncing with @esethna, we're aiming to deprioritize chat sync altogether, with no plan to "sell" only DMs to any customer.

server/configuration.go
```diff
@@ -46,8 +42,8 @@ type configuration struct {
 	SyntheticUserAuthService           string `json:"syntheticUserAuthService"`
 	SyntheticUserAuthData              string `json:"syntheticUserAuthData"`
 	AutomaticallyPromoteSyntheticUsers bool   `json:"automaticallyPromoteSyntheticUsers"`
-	DisableSyncMsg                     bool   `json:"disableSyncMsg"`
-	DisableCheckCredentials            bool   `json:"disableCheckCredentials"`
+	UseSharedChannels                  bool   `json:"experimentalUseSharedChannels"`
```
sbishel, line 45: Curious as to the word choice "Use". `UseSharedChannels` makes me think this is related to the SharedChannels architecture. (`EnableSharedChannels`). `ShareChannels` or `AllowSharedChannels` seems more appropriate. I'm 0/5 and may be misunderstanding.

sbishel (approved):
Just a question as to naming.

server/configuration.go
lieut-data: @sbishel, this is indeed referring to the shared channels architecture feature that won't be productized for v10. Agree the wording is confusing -- is there another term that might help clarify? Maybe even explicitly `UseSharedChannelsArchitecture` or `EnableSharedChannelsArchitecture`?

lieut-data: @sbishel, just circling back to our conversation above -- any thoughts after the clarification?

sbishel: If it is referring to the architecture, I think the current naming is fine, no need to add `Architecture`. Its use in `plugin.go` made me think it was more of an on/off flag rather than which architecture to use. My misunderstanding of its full purpose...only looking at the PR.

PR #691, opened 2024-06-05
MM-58480: set user_user_icon with from_webhook

lieut-data:
#### Summary
Turns out that if you set `from_webhook` (which we do to avoid collapsing adjacent messages), you also need `user_user_icon`, otherwise if `Enable integrations to override profile picture icons` is /enabled/ (yes!), it reverts to the default webhook URL.

Basically, change this:

CleanShot 2024-06-05 at 16 45 26@2x

To: 
CleanShot 2024-06-05 at 16 45 36@2x

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-58480


larkox (approved):

hmhealey (approved):
It'd be really nice if we documented some of these props since I did not remember that existed. Looks like it dates back to https://github.com/mattermost/mattermost/pull/7381

PR #692, opened 2024-06-06
MM-58580: show message on welcome bot dismiss

lieut-data:
#### Summary
When dismissing a welcome message from the MS Teams bot re: notifications, show a message explaining how to change those settings instead of just deleting it outright:

CleanShot 2024-06-06 at 12 10 17@2x

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-58580

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -1096,6 +1096,28 @@ func (a *API) enableNotifications(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Verify the post was authored by the bot itself.
+	if post.UserId != a.p.botUserID {
+		a.p.API.LogWarn("Attempt to update post not authored by the bot", "user_id", userID)
+		http.Error(w, "Forbidden", http.StatusForbidden)
+		return
+	}
+
+	// Verify the post is in the direct message channel between the bot and the user.
+	botDMChannel, err := a.p.apiClient.Channel.GetDirect(a.p.botUserID, userID)
+	if err != nil {
+		a.p.API.LogWarn("Unable to get the bot direct channel", "user_id", userID, "error", err.Error())
+		http.Error(w, "failed to authenticate the request", http.StatusInternalServerError)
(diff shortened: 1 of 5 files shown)


server/api.go
```diff
@@ -1147,9 +1171,15 @@ func (a *API) dismissNotifications(w http.ResponseWriter, r *http.Request) {
 
 	// At this point, it might be /another/ post by the bot in the DM with that user, but we
 	// allow this for now.
-	err = a.p.apiClient.Post.DeletePost(actionHandler.PostId)
+
+	post.Message = "To change your notification settings, open your user settings or run `/msteams notifications`"
```
matthewbirtch, line 1175:
```suggestion
	post.Message = "You will not receive notifications from Microsoft Teams. To change your notification settings, open your user settings or run `/msteams notifications`"
```

matthewbirtch (approved):
Minor change proposed to the language.

server/api.go
lieut-data: Today, `Dismiss` just means that: dismiss the message. If, by chance, you had enabled the setting separately and only then went to dismiss this post, it would not be true that you had chosen not to receive notifications.

We could change `Dismiss` to `Disable` to make this explicit, or we could leave the message ambiguous to represent either starting state, or we could even look up the current setting and add two new messages instead.

Thoughts, @matthewbirtch!

matthewbirtch: @lieut-data I think it makes sense to change `Dismiss` to `Disable` to make it explicit.

lieut-data: Sounds good, thank you @matthewbirtch!

matthewbirtch (approved):

JulienTant (approved):
LGTM, thanks @lieut-data

JulienTant (approved):
LGTM, thanks @lieut-data

PR #699, opened 2024-06-12
MM-58702: don't assume JSON endpoint

lieut-data:
#### Summary
Check the response header from the server before trying to interpret the response as JSON. Not all endpoints reply with JSON. And now those that should now do so explicitly.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-58702

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -96,6 +96,17 @@ func NewAPI(p *Plugin, store store.Store) *API {
 	return api
 }
 
+// returnJSON writes the given data as json with the provided httpStatus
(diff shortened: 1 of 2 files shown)


calebroseland (approved):

hmhealey (approved):

hmhealey:
It's interesting to compare this to how the main server and app work. We tend to send `{"status": "ok"}` as a response to anything that isn't responding with data, and it never occurred to me that that was because `Client4` and its predecessors always expected to receive JSON data

PR #708, opened 2024-07-11
MM-59369: reduce surface area

lieut-data:
#### Summary
This is an invasive change: we're essentially ripping out everything unrelated to shipping one-way sync of chat notifications as part of pre-packaging this plugin for Mattermost v10. This includes removing channel sync, synthetic users, and even the embedded experience. 

The rationale is to expedite security review of the remaining code and narrow our support radius to just the feature set we intend to support going forward. 

Should we want to restore any of these features, we'll be able to return to the previous tags and build out the feature set using that as a reference as needed. In the case of embedding Mattermost in Teams, we'd likely pull that out into a separate plugin altogether.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-59464
Fixes: https://mattermost.atlassian.net/browse/MM-55512
Fixes: https://mattermost.atlassian.net/browse/MM-59261


sbishel (approved):

JulienTant (approved):
Ok reviewing this was daunting, yet quite satisfying haha.

Thanks for the cleanup!

PR #711, opened 2024-07-15
MM-59466 remove unused migrations

lieut-data:
#### Summary
This pull request strips out unneeded migrations (keeping their keys as legacy strings so we avoid accidentally reusing them.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-59467


JulienTant (approved):
Looks good, just wondering what kind of impact this would have on a big deployment like SNOW?

sbishel (approved):
Instead of looping and calling the pluginAPI for `DeleteUser`, which could be a performance issue, why not just handle via SQL Queries.

I understand that if these were "normal" users why you would do this. But all of the additional things that this function handles, such as invalidating caches and sending web socket events, revoking sessions, calling other plugins, are performance hits that I don't believe are necessary for these users. 

Unless, I'm missing or forgetting something.

lieut-data:
> Instead of looping and calling the pluginAPI for `DeleteUser`, which could be a performance issue, why not just handle via SQL Queries.
> 
> I understand that if these were "normal" users why you would do this. But all of the additional things that this function handles, such as invalidating caches and sending web socket events, revoking sessions, calling other plugins, are performance hits that I don't believe are necessary for these users.

I do agree we might be able to get away with a more manual deletion, but I'm uncertain of the enduser side effects. For example, these users might be in client-side caches, but won't get "refreshed" without the websocket events. I'm hoping that if we just do this in Cloud during downtime, we won't need to deal with unexpected fallout of direct database manipulation.

lieut-data:
@sbishel & @JulienTant, with the plan to manually transition Cloud customers, I'm leaning towards possibly just removing unneeded migrations and documenting how to achieve this same effect instead of running a migration at all. Thoughts?

sbishel:
> @sbishel & @JulienTant, with the plan to manually transition Cloud customers, I'm leaning towards possibly just removing unneeded migrations and documenting how to achieve this same effect instead of running a migration at all. Thoughts?

Yeah. Might be better, allows cleaning up the database without installing the plugin. We don't have too many that we know about either.

PR #713, opened 2024-07-16
default notifications on

lieut-data:
#### Summary
For new plugin installs, default the notifications experience on at the admin level (individual users still need to enable for themselves).

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-59509

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -64,7 +64,7 @@
         "display_name": "Sync notifications",
         "type": "bool",
         "help_text": "Sync notifications of chat messages for any connected user that enables the feature.",
-        "default": false
+        "default": true
       },
       {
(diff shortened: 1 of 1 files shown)


JulienTant (approved):

JulienTant:
Just wondering, what other uses does the plugin have if this is turned off?

sbishel (approved):

lieut-data:
> Just wondering, what other uses does the plugin have if this is turned off? Isn't it the same as disabling the plugin?

You're right, @JulienTant at the moment it's somewhat redundant to have this setting at all, hence at least defaulting it on. Given the plan to merge more functionality from other plugins, it will be useful to keep the configuration discrete.

PR #714, opened 2024-07-18
Monitor permissions

lieut-data:
#### Summary
To aid in debugging deployments, let's periodically check the application permissions and log those missing or redundant.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-59549

### DIFF
--- a/server/credentials.go
+++ b/server/credentials.go
@@ -1,13 +1,101 @@
 package main
 
 import (
+	"fmt"
 	"runtime/debug"
 	"sort"
 	"strings"
 	"time"
 
 	"github.com/mattermost/mattermost-plugin-msteams/server/metrics"
+	"github.com/mattermost/mattermost-plugin-msteams/server/msteams/clientmodels"
 )
+
+const (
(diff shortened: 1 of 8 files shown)


server/credentials.go
```diff
@@ -1,14 +1,95 @@
...
+	case ResourceAccessTypeScope:
+		return "Scope (Delegated)"
+	default:
+		return resourceAccess.Type
+	}
+}
+
+// getExpectedPermissions returns the set of expected permissions, keyed by the
+// name the enduser would expect to see in the Azure tenant.
+func getExpectedPermissions() []expectedPermission {
```
lieut-data, line 46: This is hard-coded for now, but I anticipate the permission set changing as we add functionality, some of which may be optional.

enahum (approved):
Lgtm but I think as is related to some permissions and logging perhaps it makes sense to include a security review?

esarafianou:
@lieut-data yes that works considering we'll start the review in the beginning of August

mgdelacroix (approved):
LGTM! :+1:

PR #716, opened 2024-07-25
Patches for v2

lieut-data:
#### Summary
A handful of fixes to reported and observed issues after the bug bash:
* The site statistics on the system console weren't updating, and also hadn't been changed to reflect the one-way sync.
* During testing, it was observed that if `OnConfigurationChange` fails, it often clears out otherwise good values for various fields. This is super painful for the admin, so just reset to sensible values instead.
* We were would always renew subscriptions from MS Teams lifecycle events, even if we didn't rely on them anymore. Stop doing that.
* Watch out for panics in the metrics job (none found, but preemptive change), and fix the issue that would reset metrics on plugin upgrade.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-59876
Fixes: https://mattermost.atlassian.net/browse/MM-59862
Fixes: https://mattermost.atlassian.net/browse/MM-59860
Fixes: https://mattermost.atlassian.net/browse/MM-59859
Fixes: https://mattermost.atlassian.net/browse/MM-59861


server/api.go
```diff
@@ -86,8 +86,8 @@ func NewAPI(p *Plugin, store store.Store) *API {
 
 // returnJSON writes the given data as json with the provided httpStatus
 func (a *API) returnJSON(w http.ResponseWriter, data any) {
-	w.WriteHeader(http.StatusOK)
```
lieut-data, line 89: From https://pkg.go.dev/net/http:

>  Changing the header map after a call to [ResponseWriter.WriteHeader] has no effect unless the HTTP status code was of the 1xx class or the modified headers are trailers.

I think I broke this a while back when fixing a different endpoint.

server/api.go
```diff
@@ -879,31 +879,23 @@ func (a *API) siteStats(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "unable to get whitelisted users count", http.StatusInternalServerError)
 		return
 	}
-	receiving, err := a.p.store.GetActiveUsersReceivingCount(metricsActiveUsersRange)
```
lieut-data, line 882: Just simplifying.

server/api_test.go
```diff
@@ -222,7 +221,7 @@ func TestProcessLifecycle(t *testing.T) {
 				Resource:       "mockResource",
 				ChangeType:     "mockChangeType",
 				ClientState:    "mockClientState",
-				LifecycleEvent: "mockLifecycleEvent",
+				LifecycleEvent: "reauthorizationRequired",
```
lieut-data, line 224: Use real events, since we now check and discard unrecognized ones.

server/configuration.go
```diff
@@ -40,6 +40,12 @@ func (c *configuration) ProcessConfiguration() {
 	c.ClientSecret = strings.TrimSpace(c.ClientSecret)
 	c.EncryptionKey = strings.TrimSpace(c.EncryptionKey)
 	c.WebhookSecret = strings.TrimSpace(c.WebhookSecret)
+	if c.MaxSizeForCompleteDownload < 0 {
```
lieut-data, line 43: There's a bug with the system console where values can sometimes not get saved, making it brutal to keep losing the various secrets. If this happens for these values, just use sensible values.

server/handlers.go
```diff
@@ -141,18 +142,33 @@ func (ah *ActivityHandler) Handle(activity msteams.Activity) error {
...
 
 func (ah *ActivityHandler) HandleLifecycleEvent(event msteams.Activity) {
-	if event.LifecycleEvent == "reauthorizationRequired" {
-		expiresOn, err := ah.plugin.GetClientForApp().RefreshSubscription(event.SubscriptionID)
-		if err != nil {
-			ah.plugin.GetAPI().LogWarn("Unable to refresh the subscription", "error", err.Error())
-			ah.plugin.GetMetrics().ObserveLifecycleEvent(event.LifecycleEvent, metrics.DiscardedReasonFailedToRefresh)
-			return
-		}
+	if event.LifecycleEvent != "reauthorizationRequired" {
```
lieut-data, line 145: Check the event type upfront.

server/handlers.go
```diff
@@ -141,18 +142,33 @@ func (ah *ActivityHandler) Handle(activity msteams.Activity) error {
...
+		ah.plugin.GetMetrics().ObserveLifecycleEvent(event.LifecycleEvent, metrics.DiscardedReasonUnknownLifecycleEvent)
+		return
+	}
 
-		ah.plugin.GetMetrics().ObserveSubscription(metrics.SubscriptionRefreshed)
-		if err = ah.plugin.GetStore().UpdateSubscriptionExpiresOn(event.SubscriptionID, *expiresOn); err != nil {
-			ah.plugin.GetAPI().LogWarn("Unable to store the subscription new expiry date", "subscription_id", event.SubscriptionID, "error", err.Error())
-		}
+	// Ignore subscriptions we aren't tracking locally. For now, that's just the single global chats subscription.
```
lieut-data, line 150: We would renew everything, even if unused -- lots of time on msteams-sync-test doing this until I upgraded to a tes build.

server/metrics/metrics.go
```diff
@@ -50,7 +52,8 @@ const (
 
 	WorkerMonitor          = "monitor"
 	WorkerActivityHandler  = "activity_handler"
-	WorkerCheckCredentials = "check_credentials" //#nosec G101 -- This is a false positive
+	WorkerCheckCredentials = "check_credentials" //#nosec G101 -- This is a false positiveo
```
lieut-data, line 55:
```suggestion
	WorkerCheckCredentials = "check_credentials" //#nosec G101 -- This is a false positive
```

server/plugin.go
```diff
@@ -256,6 +256,9 @@ func (p *Plugin) start(isRestart bool) {
 		if err != nil {
 			p.API.LogError("failed to start metrics job", "error", err)
 		}
+
+		// Run the job above right away so we immediately populate metrics.
+		go p.updateMetrics()
```
lieut-data, line 261: We'd always get dips otherwise: 

CleanShot 2024-07-25 at 17 30 45@2x

server/plugin.go
```diff
@@ -294,7 +297,7 @@ func (p *Plugin) start(isRestart bool) {
 		p.checkCredentialsJob = checkCredentialsJob
 
 		// Run the job above right away so we immediately populate metrics.
-		p.checkCredentials()
+		go p.checkCredentials()
```
lieut-data, line 300: Faster, Spock!

server/plugin.go
```diff
@@ -540,6 +543,16 @@ func (p *Plugin) GetRemoteID() string {
 }
 
 func (p *Plugin) updateMetrics() {
+	defer func() {
```
lieut-data, line 546: This was run in an implicit goroutine already, so just tracking it like a proper worker.

server/store/sqlstore/store.go
```diff
@@ -829,33 +829,17 @@ func (s *SQLStore) getConnectedUsersCount(db sq.BaseRunner) (connectedUsers int6
...
-	return activeUsersSending, err
-}
-
-//db:withReplica
-func (s *SQLStore) getActiveUsersReceivingCount(db sq.BaseRunner, dur time.Duration) (activeUsersReceiving int64, err error) {
-	now := time.Now()
-
-	err = s.getQueryBuilder(db).
-		Select("count(*)").
+		Select("count(mmUserID)").
```
lieut-data, line 836: Stop `COUNT(*)` which requires a table scan.

server/store/sqlstore/store.go
```diff
@@ -829,33 +829,17 @@ func (s *SQLStore) getConnectedUsersCount(db sq.BaseRunner) (connectedUsers int6
...
 		From(usersTableName).
 		Where(sq.GtOrEq{"LastChatReceivedAt": now.Add(-dur).UnixMicro()}).
-		Where(sq.LtOrEq{"LastChatReceivedAt": now.UnixMicro()}).
```
lieut-data, line 854: This is redundant.

server/configuration.go
enahum: So what are we actually fixing here, some default values that were missing? Anyone looking at the bug you mention?

enahum (approved):
Fantastic

server/configuration.go
lieut-data: I'm trying to work around https://community.mattermost.com/core/pl/o78pcxfj87f4frt3qhrqxw1x1y, specifically.

server/connect.go
```diff
@@ -86,7 +86,7 @@ func (p *Plugin) SendInviteMessage(user *model.User, pendingSince time.Time, cur
 		return errors.Wrapf(err, "failed to get bot DM channel with user_id %s", user.Id)
 	}
 
-	message := fmt.Sprintf("@%s, you’re invited to use the Microsoft Teams connected experience for Mattermost. ", user.Username)
+	message := fmt.Sprintf("@%s, you're been invited by your administrator to connect your Mattermost account with Microsoft Teams.", user.Username)
```
calebroseland, line 89:
```suggestion
	message := fmt.Sprintf("@%s, you've been invited by your administrator to connect your Mattermost account with Microsoft Teams.", user.Username)
```

server/connect.go
```diff
@@ -97,7 +97,7 @@ func (p *Plugin) SendInviteMessage(user *model.User, pendingSince time.Time, cur
 	}
 
 	connectURL := fmt.Sprintf(p.GetURL()+"/connect?post_id=%s&channel_id=%s", invitePost.Id, channel.Id)
-	invitePost.Message = fmt.Sprintf("%s [Click here to activate the integration in a minute or less](%s).", invitePost.Message, connectURL)
+	invitePost.Message = fmt.Sprintf("%s [Click here to connect your account](%s).", invitePost.Message, connectURL)
```
calebroseland, line 100: Much more succinct

calebroseland (approved):
LTGM, just that one text bit that needs changing

server/connect.go
lieut-data: Brilliant, thank you!

PR #727, opened 2024-08-05
MM-60005: dont notify if present

lieut-data:
#### Summary
If a user gets a Teams notification in Mattermost and then hops over to Teams, they continue to get “spammed” by messages in the ensuing chat. Let’s detect user presence and only notify users if they aren’t available in Teams.

This also means that while a user is "Available" in Teams, they won't get notifications in Mattermost. By default, this will be the state for 5 minutes after a user was last active. Note that if a user has their mobile device connected, but in the background, it won't count against the user being seen by Mattermost as "Available" in Teams.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-60005

### DIFF
--- a/server/credentials.go
+++ b/server/credentials.go
@@ -87,6 +87,13 @@ func getExpectedPermissions() []expectedPermission {
 				Type: "Role",
 			},
 		},
+		{
+			Name: "https://graph.microsoft.com/Presence.Read.All",
+			ResourceAccess: clientmodels.ResourceAccess{
+				ID:   "a70e0c2d-e793-494c-94c4-118fa0a67f42",
+				Type: "Role",
+			},
+		},
 	}
 }
 
--- a/server/msteams/client_timerlayer/timerlayer.go
+++ b/server/msteams/client_timerlayer/timerlayer.go
@@ -412,6 +412,28 @@ func (c *ClientTimerLayer) GetMyID() (string, error) {
 	return result, err
 }
 
+func (c *ClientTimerLayer) GetPresencesForUsers(userIDs []string) (map[string]*clientmodels.Presence, error) {
+	statusCode := "2XX"
+	success := "true"
+	start := time.Now()
+
+	result, err := c.Client.GetPresencesForUsers(userIDs)
+
+	elapsed := float64(time.Since(start)) / float64(time.Second)
+
+	if err != nil {
+		success = "false"
+		statusCode = "0"
+		var apiErr *msteams.GraphAPIError
+		if errors.As(err, &apiErr) {
+			statusCode = strconv.Itoa(apiErr.StatusCode)
+		}
+	}
+
+	c.metrics.ObserveMSGraphClientMethodDuration("Client.GetPresencesForUsers", success, statusCode, elapsed)
+	return result, err
+}
+
 func (c *ClientTimerLayer) GetReply(teamID string, channelID string, messageID string, replyID string) (*clientmodels.Message, error) {
 	statusCode := "2XX"
 	success := "true"
--- a/server/notifications.go
+++ b/server/notifications.go
@@ -15,6 +15,22 @@ func (ah *ActivityHandler) handleCreatedActivityNotification(msg *clientmodels.M
 		return metrics.DiscardedReasonChannelNotificationsUnsupported
 	}
 
+	// Get the presence status for each chat member to decide if we should relay into Mattermost.
+	userIDs := make([]string, 0, len(chat.Members))
+	for _, member := range chat.Members {
+		// We won't notify senders about their own posts, so we don't need their presence.
+		if member.UserID == msg.UserID {
+			continue
+		}
+
+		userIDs = append(userIDs, member.UserID)
+	}
+
+	presences, err := ah.plugin.GetClientForApp().GetPresencesForUsers(userIDs)
+	if err != nil {
+		ah.plugin.GetAPI().LogWarn("Failed to fetch presence information for chat members", "chat_id", chat.ID, "message_id", msg.ID, "error", err)
+	}
+
 	botUserID := ah.plugin.GetBotUserID()
 
 	chatLink := fmt.Sprintf("https://teams.microsoft.com/l/message/%s/%s?tenantId=%s&context={\"contextType\":\"chat\"}", chat.ID, msg.ID, ah.plugin.GetTenantID())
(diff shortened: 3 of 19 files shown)


server/presence.go
```diff
@@ -0,0 +1,50 @@
...
+// userPresenceIsActive returns true if the user is considered online in Teams.
+func userPresenceIsActive(presence *clientmodels.Presence) bool {
+	// If we're missing presence, default to the user being inactive.
+	if presence == nil {
+		return false
+	}
+
+	// Explicitly handle known activity states for being inactive or away.
+	switch presence.Activity {
+	case PresenceActivityOffline, PresenceActivityInactive, PresenceAvailabilityAway:
```
lieut-data, line 44: For now, I'm proposing we key off organic states -- if a user explicitly sets one of the other configurable states in Teams, we'll treat them as online and managing their notifications there. Experimentation will be required to determine if this matches users behaviour, but the happy path is covered.

server/subscriptions.go
```diff
@@ -139,8 +140,7 @@ func (m *Monitor) checkGlobalChatsSubscription(remoteSubscription *clientmodels.
 func (m *Monitor) getMSTeamsSubscriptionsMap() (msteamsSubscriptionsMap map[string]*clientmodels.Subscription, allChatsSubscription *clientmodels.Subscription, err error) {
 	msteamsSubscriptions, err := m.client.ListSubscriptions()
 	if err != nil {
-		m.api.LogError("Unable to list MS Teams subscriptions", "error", err.Error())
-		return nil, nil, err
+		return nil, nil, errors.Wrap(err, "failed to list subscriptions")
```
lieut-data, line 143: We already log this error in the caller, so save a duplicate log.

server/notifications.go
```diff
@@ -29,22 +47,48 @@ func (ah *ActivityHandler) handleCreatedActivityNotification(msg *clientmodels.M
...
 		if err != nil {
 			ah.plugin.GetAPI().LogWarn("Failed to get bot DM channel with user", "user_id", mattermostUserID, "teams_user_id", member.UserID, "error", err)
+			ah.plugin.metricsService.ObserveNotification(isGroupChat, hasFilesUnknown, metrics.DiscardedReasonInternalError)
 			continue
 		}
 
 		post, skippedFileAttachments, _ := ah.msgToPost(channel.Id, botUserID, msg, chat, []string{})
 
-		ah.plugin.metricsService.ObserveNotification(len(chat.Members) >= 3, len(post.FileIds) > 0)
+		hasFiles := len(post.FileIds) > 0
```
lieut-data, line 90: Technically, this should be the same for all users, but we only know it after we process the message and upload files on a per user basis.

lieut-data:
@crspeller, wondering if I can get an "out of band" set of eyes on this change? Interested not only in the review itself, but any feedback you have from looking at this and related code with fresh eyes :)

server/credentials.go
```diff
@@ -87,6 +87,13 @@ func getExpectedPermissions() []expectedPermission {
 				Type: "Role",
 			},
 		},
+		{
```
crspeller, line 90: I am guessing this means the plugin will break until this permission is added? Unless the admin gave extra permissions?

crspeller (changes requested):
Not a lot to comment on. Mostly just observations.

server/presence.go
```diff
@@ -0,0 +1,50 @@
+package main
```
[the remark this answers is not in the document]
lieut-data: Good feedback -- your intuition is right on the boundary between `msteams` and `main`. For the most part, we're trying to keep business logic out of `msteams` and otherwise flatten most everything else into `main`.

We went through a cleanup in the past few weeks eliminating layers of interfaces and packages in favour of something simpler to maintain, but might split out in the future if this project has legs.

server/credentials.go
lieut-data: Good question: no, it won't "break", but it also won't behave to spec. Without the permission:
* The logs will log errors daily
* The presence semantics won't apply, and we'll always notify users (also clearly noted in logs)

server/msteams/client_timerlayer/timerlayer.go
```diff
@@ -412,6 +412,28 @@ func (c *ClientTimerLayer) GetMyID() (string, error) {
 	return result, err
 }
 
+func (c *ClientTimerLayer) GetPresencesForUsers(userIDs []string) (map[string]*clientmodels.Presence, error) {
```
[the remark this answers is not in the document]
lieut-data: It's an interesting idea -- right now, we're copying what the server does, but if you can point me at an example of this live elsewhere I'd be open to mimicking!

server/notifications.go
```diff
@@ -29,22 +47,48 @@ func (ah *ActivityHandler) handleCreatedActivityNotification(msg *clientmodels.M
 			continue
 		} else if err != nil {
 			ah.plugin.GetAPI().LogWarn("Failed to map Teams user to Mattermost user", "teams_user_id", member.UserID, "error", err)
+			ah.plugin.metricsService.ObserveNotification(isGroupChat, hasFilesUnknown, metrics.DiscardedReasonInternalError)
 			continue
 		}
 
 		if !ah.plugin.getNotificationPreference(mattermostUserID) {
```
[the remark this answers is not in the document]
crspeller: I meant for `getNotificationPrefrence` the Not notification preference made me go chek what that function was doing. Maybe `userEnabledNotifications` would be more clear. 0/5 (yes it's unrelated to this PR)

crspeller (approved):

sbishel (approved):
Code looks good. I don't understand the Teams side - Activity vs Availability, user actions, etc. 

You already know this, but I can see this causing confusion as to "When do I get notifications?" Similar but not as complicated as selective sync.

lieut-data:
> I don't understand the Teams side - Activity vs Availability, user actions, etc.

Indeed, this is somewhat blurry for me as well @sbishel. The combination I've picked above was derived partly from documentation at https://learn.microsoft.com/en-us/microsoftteams/presence-admins and partly from manual testing.

> You already know this, but I can see this causing confusion as to "When do I get notifications?"

Also agreed. It's no consolation to the enduser, but we'll be explicitly in logs when this happens. And while this swings us neatly away from the problem being solved, we could tweak it back a bit later by monitoring chats that don't get marked as read and "catch you up" in Mattermost.

PR #728, opened 2024-08-08
Remove unused code

mgdelacroix:


esarafianou (approved):

server/plugin.go
```diff
@@ -504,7 +504,7 @@ func generateSecret() (string, error) {
 }
 
 func (p *Plugin) GenerateRandomPassword() string {
-	lowerCharSet := "abcdedfghijklmnopqrst"
+	lowerCharSet := "abcdefghijklmnopqrstuvwxyz"
```
lieut-data, line 507: Should we just remove this function altogether? It looks to be unused now.

lieut-data (approved):

server/plugin.go
mgdelacroix: 0/5 really, if we keep it there, it should be fixed, but it does look like dead code by now. If we can't think of any other use for this coming in the near future, I'm happy to delete it

lieut-data: Yeah, let's just delete it. Easy to bring back if/when we need to: the goal for v10 is to reduce the surface area for @esarafianou and team and streamline security review.

lieut-data (approved):

esarafianou (approved):

jespino (approved):
LGTM

PR #747, opened 2024-09-18
Improved logging

lieut-data:
#### Summary
Improved logging:
* consistently log on connect/disconnect
* log whenever we send a bot message to a user, now including invite flow
* log on notification delivery
* fix various issues logging around chat notifications, including empty messages

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-60060
Fixes: https://mattermost.atlassian.net/browse/MM-60061

### DIFF
--- a/server/notifications.go
+++ b/server/notifications.go
@@ -3,6 +3,7 @@ package main
 import (
 	"database/sql"
 	"fmt"
+	"strings"
 
 	"github.com/mattermost/mattermost-plugin-msteams/server/metrics"
 	"github.com/mattermost/mattermost-plugin-msteams/server/msteams/clientmodels"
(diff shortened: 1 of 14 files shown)


server/notifications.go
```diff
@@ -99,13 +117,46 @@ func (ah *ActivityHandler) handleCreatedActivityNotification(msg *clientmodels.M
...
+		}
+
+		ah.plugin.GetAPI().LogInfo(
+			"Delivered notification for chat member away from Teams",
+			"user_id", mattermostUserID,
+			"teams_user_id", member.UserID,
+			"chat_id", chat.ID,
+			"message_id", msg.ID,
+			"activity", presences[member.UserID].Activity,
+			"availability", presences[member.UserID].Availability,
```
gabrieljackson, line 142: Seeing fields needing to be defined for every log message like this really makes me appreciate how loggers like logrus allow you to slowly build a custom logger for each flow that contains important field values. Would we ever consider making important field keys constants and/or making field helper functions like userIDField(mattermostUserID) to help eliminate typos or allow for renaming a field key? Not sure I even think that would be a good idea, but figured I would float it by you.

gabrieljackson (approved):
Looks great! Just left some rambling thoughts and questions below, but 0/5 on changing anything.

calebroseland (approved):

server/notifications.go
lieut-data: > Seeing fields needing to be defined for every log message like this really makes me appreciate how loggers like logrus allow you to slowly build a custom logger for each flow that contains important field values. 

Yes, I'm in full alignment. I "stuck to the basics" for this PR, but got fed up trying to replicate the fields over at https://github.com/mattermost/mattermost-plugin-msteams/pull/749/files and ended up just importing logrus there.

> Would we ever consider making important field keys constants and/or making field helper functions...

Oh wow, yes! I spent a bunch of clean up early on in this project fixing `user_id` and `userID` to make grepping logs more useful. Formalizing a `withUser([string|model.User])` would be even better!

I've added this idea to the pool over at https://mattermost.atlassian.net/wiki/spaces/Toolkit/pages/3044966406/Toolkit+Improvements+Q4+2024.

PR #748, opened 2024-09-18
MM-60576: enable notifications by default

lieut-data:
#### Summary
Enable notifications by default, immediately after connecting. This can still be disabled via settings, but no longer requires an onboarding flow. Various text updates are included:

CleanShot 2024-09-18 at 13 28 54@2x

CleanShot 2024-09-18 at 13 30 29@2x

CleanShot 2024-09-18 at 13 30 18@2x

CleanShot 2024-09-18 at 13 30 41@2x

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-60576


lieut-data:
@enahum & @Willyfrog, wondering if you'd be open to reviewing a minor change requested by @esethna & @asaadmahmood?

enahum (approved):

server/api.go
```diff
@@ -503,6 +501,13 @@ func (a *API) oauthRedirectHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	err = a.p.setNotificationPreference(mmUserID, true)
+	if err != nil {
+		a.p.API.LogWarn("Failed to update user preferences on connect", "user_id", mmUserID, "error", err.Error())
+		http.Error(w, "error updating the preferences", http.StatusInternalServerError)
```
Willyfrog, line 507: should there be a reference to teams in the error logs/messages? Same way as you have when it is successful. Or are admins/users able to tell this is related to the plugin?

lieut-data: Ooh, good question! In this case, we get the benefit of the server automatically tagging all logs with the `plugin_id`, e.g.:

```json
{"timestamp":"2024-09-24 15:10:08.987 Z","level":"warn","msg":"Failed to update user preferences on connect","plugin_id":"com.mattermost.msteams-sync","user_id":"rcnuja8z8p8mb811p989kpaosh"}
```

PR #784, opened 2025-02-26
Restore iframe support

lieut-data:
#### Summary
Bring back the ability to embed Mattermost inside Microsoft Teams. For now, we're exposing just the endpoint to support embedding Mattermost as an iframe, and handling the manifest generation elsewhere.

#### Ticket Link
N/A


wiggin77 (approved):

server/iframe.go
```diff
@@ -0,0 +1,54 @@
...
+	siteURL := *config.ServiceSettings.SiteURL
+	if siteURL == "" {
+		a.p.API.LogError("SiteURL cannot be empty for MS Teams iFrame")
+		http.Error(w, "SiteURL is empty", http.StatusInternalServerError)
+		return
+	}
+
+	html := strings.ReplaceAll(iFrameHTML, "{{SITE_URL}}", siteURL)
+
+	w.Header().Set("Content-Type", "text/html")
```
enzowritescode, line 24: We should set some additional security headers here but we will want to test how it behaves in Teams. After some back and forth with Aider, here is some sample code (NOTE: it uses `parsedURL` from the validation code above):

```
// Get the origin (scheme + host) for CSP
origin := parsedURL.Scheme + "://" + parsedURL.Host

// Set a minimal CSP for the wrapper page
cspDirectives := []string{
    "default-src 'none'",           // Block all resources by default
    "frame-src " + origin,          // Only allow iframe to load content from Mattermost origin
    "style-src 'unsafe-inline'",    // Allow inline styles for the iframe positioning
}
w.Header().Set("Content-Security-Policy", strings.Join(cspDirectives, "; "))

// Set additional security headers w.Header().Set("X-Content-Type-Options", "nosniff")
w.Header().Set("X-Frame-Options", "SAMEORIGIN") // This is for the outer page, not the iframe content
w.Header().Set("Referrer-Policy", "strict-origin-when-cross-origin")  
```

server/iframe.go
```diff
@@ -0,0 +1,54 @@
...
+
+var iFrameHTML = `<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="UTF-8">
+  <title>Mattermost</title>
+  <meta name="viewport" content="width=device-width, height=device-height, initial-scale=1.0">
+</head>
+<body>
+	<iframe
```
enzowritescode, line 49: I double checked that our content security policy allows us to be embedded in an iFrame within MS Teams (currently it has `frame-ancestors 'self' teams.microsoft.com;` which is good - although we can probably have it be `https://teams.microsoft.com` since I doubt Microsoft teams is ever over `http`. That code lives in `mattermost/server/channels/web/handlers.go`

enzowritescode (changes requested):

server/iframe.go
```diff
@@ -0,0 +1,54 @@
...
+	_ "embed"
+	"net/http"
+	"strings"
+)
+
+// iFrame returns the iFrame HTML needed to host Mattermost within a MS Teams app.
+func (a *API) iFrame(w http.ResponseWriter, _ *http.Request) {
+	config := a.p.API.GetConfig()
+	siteURL := *config.ServiceSettings.SiteURL
+	if siteURL == "" {
```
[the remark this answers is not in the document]
lieut-data: :+1: for checking a valid SiteURL, but any objections on skipping the explicit SSL check? (I'll need to rewrite the testing framework to allow us to unit test this the way things are currently setup.)

server/iframe.go
lieut-data: The `X-Frame-Options` header failed to load the iFrame, I think because the outer isn't running on the same origin (it's on teams.microsoft.com?)

CleanShot 2025-02-26 at 18 29 51@2x

Removing it and the others worked.

server/iframe.go
lieut-data: I won't be able to fix this as part of this PR, but it's something we can patch the core app with.

server/iframe.go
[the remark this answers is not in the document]
enzowritescode: Ah okay that's fine

server/iframe.go
enzowritescode: Good to remove since it's covered in the CSP via frame-src (just repeating what I DMd you lol)

server/iframe.go
enzowritescode: Yup just wanted to put that on your radar

enzowritescode (approved):

PR #802, opened 2025-03-13
feat: sso support

fmartingr:
#### Summary

- Adds a route that users coming from SSO in an embedded experience get redirected to (see https://github.com/mattermost/mattermost/pull/30408)
- Adds JS to the iframe to handle the SSO logic


lieut-data:
Superseded by https://github.com/mattermost/mattermost-plugin-msteams/pull/803.

PR #803, opened 2025-03-17
Support implicit OAuth via MS Teams

lieut-data:
#### Summary
This pull request enables automatic login to the embedded iFrame experience based on matching the user principal name against the corresponding Mattermost user, assuming a valid Teams JWT for the configured tenant.

Here's a flowchart and diagram of how things currently work:

image

#### Ticket Link
None.


server/iframe.go
```diff
@@ -46,30 +53,175 @@ func (a *API) iFrame(w http.ResponseWriter, _ *http.Request) {
...
+		logger.Info("Skipping authentication, user already logged in")
+		http.Redirect(w, r, "/", http.StatusSeeOther)
+		return
+	}
+
+	config := a.p.apiClient.Configuration.GetConfig()
+
+	enableDeveloper := config.ServiceSettings.EnableDeveloper
+
+	// Ideally we'd accept the token via an Authorization header, but for now get it from the query sring.
```
wiggin77, line 90:
```suggestion
	// Ideally we'd accept the token via an Authorization header, but for now get it from the query string.
```

wiggin77: I wonder if having the token in a query string will pass security.  The token will appear in proxy logs etc.  What is the life-time of a typical token here?

lieut-data: @fmartingr's original approach (before we nested twice) was to exchange this via `window.postMessage`. We can certainly go back to that if it's a blocker.

server/iframe.go
```diff
@@ -46,30 +53,175 @@ func (a *API) iFrame(w http.ResponseWriter, _ *http.Request) {
...
+	secure := true
+	maxAgeSeconds := *config.ServiceSettings.SessionLengthWebInHours * 60 * 60
+	domain := getCookieDomain(config)
+	subpath, _ := utils.GetSubpathFromConfig(config)
+
+	expiresAt := time.Unix(model.GetMillis()/1000+int64(maxAgeSeconds), 0)
+
+	session, err := a.p.apiClient.Session.Create(&model.Session{
+		UserId:   mmUser.Id,
+		DeviceId: model.NewId(),
```
wiggin77, line 168: `Session.DeviceId` is used to indicate the session is on a mobile device.  Why do we want to set that in this case?

lieut-data: Ooh, that's a good catch! No, we shouldn't do that.

server/iframe.go
```diff
@@ -46,30 +53,175 @@ func (a *API) iFrame(w http.ResponseWriter, _ *http.Request) {
...
+		// TODO, should we allow this to be configurable?
+		ExpiresAt: model.GetMillis() + (1000 * 60 * 60 * 24 * 1), // 1 day
```
wiggin77, line 169: Since it is not exposed via a plugin API, can we just duplicate the code here: https://github.com/mattermost/mattermost/blob/c95968c380be927de4db562ba2e10d26cb40a8f9/server/channels/app/platform/session.go#L224

server/iframe.go
```diff
@@ -46,30 +53,174 @@ func (a *API) iFrame(w http.ResponseWriter, _ *http.Request) {
...
+	maxAgeSeconds := *config.ServiceSettings.SessionLengthWebInHours * 60 * 60
+	domain := getCookieDomain(config)
+	subpath, _ := utils.GetSubpathFromConfig(config)
+
+	expiresAt := time.Unix(model.GetMillis()/1000+int64(maxAgeSeconds), 0)
+
+	session, err := a.p.apiClient.Session.Create(&model.Session{
+		UserId: mmUser.Id,
+		// TODO, should we allow this to be configurable?
+		ExpiresAt: model.GetMillis() + (1000 * 60 * 60 * 24 * 1), // 1 day
```
enzowritescode, line 169: There might be issues if the JWT expires before the cookie - ex. their Teams session expires but they can still access Mattermost

Let's make them the same before we consider configuring this value.

enzowritescode (changes requested):

server/iframe.go
wiggin77: If I'm reading this right, if their session expires after a day, it simply means the user go through the Entra validation code path here again, and that would happen at most once per day (provided the user stayed logged into Entra).  Based on that, ignore my previous comment.

server/plugin.go
```diff
@@ -310,6 +315,13 @@ func (p *Plugin) start(isRestart bool) {
 	if err = p.API.RegisterCommand(p.createCommand()); err != nil {
 		p.API.LogError("Failed to register command", "error", err)
 	}
+
+	p.cancelKeyFuncLock.Lock()
+	if !isRestart && p.cancelKeyFunc == nil {
+		p.tabAppJWTKeyFunc, p.cancelKeyFunc = setupJWKSet()
+	}
+	p.cancelKeyFuncLock.Unlock()
```
wiggin77, line 323: There is a fair amount of code behind the `setupJWKSet` call, that goes into some third party libs (e.g. https://github.com/MicahParks/jwkset).  If there is a panic the HTTP handler will recover but this mutex won't release and further requests will block.  

```suggestion
	defer p.cancelKeyFuncLock.Unlock()
```

wiggin77 (approved):
LGTM :+1:  One change requested (mutex unlock).

server/iframe.html
```diff
@@ -0,0 +1,65 @@
...
+    microsoftTeams.app.getContext().then((context) => {
+      let tenantId;
+      if (context && context.user && context.user.tenant && context.user.tenant.id) {
+        tenantId = context.user.tenant.id;
+      }
+
+      switch (tenantId) {
+        case 'd2888234-d303-4c94-8f45-c7348f089048':
+          microsoftTeams.authentication.getAuthToken().then((token) => {
+            iframe.src = 'https://msteams-sync-test.test.mattermost.cloud/plugins/com.mattermost.msteams-sync/iframe/authenticate?token=' + token;
```
enzowritescode, line 36: All spots where we append token to the URL we should `encodeURIComponent()` to ensure there is no unexpected behavior with the URL.

enzowritescode (changes requested):

server/iframe.html
fmartingr: Would using `URLSearchParams().toString()` suffice?

enzowritescode: That works too but it's more commonly used when you're working with multiple query params. In this case since you only have one param my suggestion also works.

fmartingr: I'm going ahead adding my suggestion since a follow up PR will add more query parameters.

server/iframe.go
```diff
@@ -46,30 +53,174 @@ func (a *API) iFrame(w http.ResponseWriter, _ *http.Request) {
...
+func (a *API) formatTemplate(template string) (string, error) {
+	config := a.p.API.GetConfig()
+	siteURL := *config.ServiceSettings.SiteURL
+	if siteURL == "" {
+		return "", fmt.Errorf("ServiceSettings.SiteURL cannot be empty for MS Teams iFrame")
+	}
+
+	html := strings.ReplaceAll(template, "{{SITE_URL}}", siteURL)
+	html = strings.ReplaceAll(html, "{{PLUGIN_ID}}", url.PathEscape(manifest.Id))
+	html = strings.ReplaceAll(html, "{{TENANT_ID}}", url.PathEscape(a.p.getConfiguration().TenantID))
```
enzowritescode, line 66: I don't think `{{TENANT_ID}}` needs `url.PathEscape()` since it's only used in the case

enzowritescode (changes requested):

server/plugin.go
lieut-data: I think if `setupJWKSet` panics, the plugin itself will shutdown since `start` and in turn `OnActivate()` / `OnConfigurationChange` will panic and kill the process.

server/iframe.go
lieut-data: Sounds good!

enzowritescode (approved):

PR #804, opened 2025-03-18
feat: send user activity notifications to teams users when mentioned in Mattermost

fmartingr:
#### Summary
- Sends a user activity when a post contains a mention to the `@`username.
- Notification ends up in the activity on the teams application
- [x] Clicking on the activity in the list shows up Mattermost highlighting the message that triggered the activity.
- [x] Clicking on the desktop notification opens the tab application highlighting the message that triggered the activity.
- [x] Customize the notification (ex: `{user} mentioned you in Mattermost`)
- [x] Use a background runner to send the notifications
- [x] Icon in the activity list.

#### How this works

First we retrieve (along other params) the `app_id` in the authentication flow and store it in the user, since we need it to deeplink the "action" of the user activity notification (so we send the user to the tabApp or show the tabApp in the activities list, see screenshots).

The flow works this way:
- Using `MessageHasBeenPosted`, we check for the `@mention` in the message. **This don't work for keywords**, we would need to duplicate more code from server to make that work.
- If there's a mention and we have the attributes from the authentication stored for that user, we create a new UserActivity for the user in the Teams side, with a `post_(postid)` in the `context.subEntityId`. This context is sent to the tab application and we can retrieve it using `microsoftTeams.app.getContext()` (under `context.page.subPageId`).
- When the user lands in the `/iframe/authenticate`, we check if that value is set and get the necessary information to generate a `/(team)/pl/(post_id)` link to use as redirect url at the end of this handler. 
  - If the parameter is not set we redirect to `/` as usual.

##### System Notification (browser and app, when allowed)

Screenshot 2025-03-18 at 13 00 06

##### Teams notification (app, web)

Screenshot 2025-03-19 at 16 49 46

##### Video

https://github.com/user-attachments/assets/fea67412-3134-4707-98e0-af829f91eebc


server/iframe.go
```diff
@@ -72,16 +76,49 @@ func (a *API) authenticate(w http.ResponseWriter, r *http.Request) {
...
 	if r.Header.Get("Mattermost-User-ID") != "" {
 		logger = logger.WithField("user_id", r.Header.Get("Mattermost-User-ID"))
 		logger.Info("Skipping authentication, user already logged in")
-		http.Redirect(w, r, "/", http.StatusSeeOther)
+		http.Redirect(w, r, redirectPath, http.StatusSeeOther)
 		return
 	}
 
+	appID := r.URL.Query().Get("app_id")
+	if appID == "" {
```
wiggin77, line 118: I think this can be closer to where it is used.  That will help remove any confusion about why this error is logged but no action taken.

server/iframe.go
```diff
@@ -72,13 +76,41 @@ func (a *API) authenticate(w http.ResponseWriter, r *http.Request) {
...
+	// Check if we have a subEntityID coming from the Microsoft Teams SDK to redirect the user to the correct URL.
+	// We use this from the Team's notifications to redirect the user to what triggered the notification, in this case,
+	// a post.
+	subEntityID := r.URL.Query().Get("sub_entity_id")
+	if subEntityID != "" {
+		if strings.HasPrefix(subEntityID, "post_") {
+			postID := strings.TrimPrefix(subEntityID, "post_")
+			post, err := a.p.API.GetPost(postID)
+			if err != nil {
+				logger.WithError(err).Error("Failed to get post to generate redirect path from subEntityId")
```
lieut-data, line 90: Already commented in other PR, but nil dereference potential here unless we break out.

server/iframe.go
```diff
@@ -222,14 +261,98 @@ func (a *API) authenticate(w http.ResponseWriter, r *http.Request) {
...
 
 	// Redirect to the home page
-	http.Redirect(w, r, "/", http.StatusSeeOther)
+	http.Redirect(w, r, redirectPath, http.StatusSeeOther)
+}
+
+// MessageHasBeenPosted is called when a message is posted in Mattermost. We rely on it to send a user activity notification
+// to Microsoft Teams when a user is mentioned in a message.
+// This is called in a controller Goroutine in the server side so there's no need to worry about concurrency here.
+func (p *Plugin) MessageHasBeenPosted(c *plugin.Context, post *model.Post) {
```
lieut-data, line 270: `MessageHasBeenPosted` tends to be expensive -- can we preflight looking up the user's MS Teams ID as early as possible and avoid doing any other work if we can abort early?

server/iframe.go
```diff
@@ -222,14 +261,98 @@ func (a *API) authenticate(w http.ResponseWriter, r *http.Request) {
...
+	p.API.LogError("Message has been posted", "post_id", post.Id, "post_participants", post.Participants)
```
lieut-data, line 271: I'm assuming this was for debugging, but definitely don't want to log on each call of this method.

server/iframe.go
```diff
@@ -222,14 +261,98 @@ func (a *API) authenticate(w http.ResponseWriter, r *http.Request) {
...
+		if !exists {
+			p.API.LogError("MSTeams user ID is empty. Not sending notification.")
+			continue
+		}
+
+		appID, exists := u.GetProp(getUserPropKey("app_id"))
+		if !exists {
+			p.API.LogError("MSTeams app ID is empty. Not sending notification.")
+			continue
+		}
```
lieut-data, line 303: Would we consider maintaining this mapping in our own table (or even just the kv store for the plugin) instead of putting it inside the user prop?

lieut-data (changes requested):
Just one high level thought before we merge: should we use our kv store instead of the user properties?

server/iframe.go
fmartingr: Fixed this on #806 where the redirect logic was isolated into it's own function: fc423b288e6a9ff21ebd382d96577e12959d66a8

server/iframe.go
fmartingr: I think we already do it as early as possible, right? We extract the mentions and check each user for the properly.

server/iframe.go
fmartingr: @wiggin77 and I talked about this at some point and decided on the user properties since with one DB Call we had all the information for the user. This can even be improved to use `ListByUsernames` to get all the users at once.

wiggin77: There is also the issue of cleanup when users are removed.

server/iframe.go
wiggin77: The author of the post may not have O365 id stored, but we still need to send notifications for the users mentioned.

server/iframe.go
```diff
@@ -222,14 +261,96 @@ func (a *API) authenticate(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	urlParams := url.Values{}
+	urlParams.Set("context", string(jsonContext))
+
+	for _, mention := range extractMentionsFromPost(post) {
+		u, err := p.apiClient.User.GetByUsername(mention)
+		if err != nil {
+			p.API.LogError("Failed to get user", "error", err.Error())
```
wiggin77, line 287: I don't think this is an error.  Can this be a debug log?  It's valid for a post to have an `@` then some text that is not a username.  e.g.  giving someone's Instagram handle.

server/iframe.go
lieut-data: I think I was just looking at the log, and `context`/`jsonContext`/`urlParams` initialization -- admittedly none of that really exercises the pluginAPI (besides the log) so it's not a big deal.

But I'm curious, could we use the `NotificationWillBePushed` hook instead?

(Unblocking my requested changes as we can iterate on this.)

lieut-data (approved):

wiggin77 (approved):

PR #806, opened 2025-03-20
feat: improve mention retrieval from posts, allows to disable notifications

fmartingr:
#### Summary
mproves and refactors logic that extracts `@mentions` from the post and checks if an user should be mentioned from the received message.

Currently handling:
- DMs
- GMs
- `@user`
- `@group`
- `@channel` / `@all`
- `@here`

Missing:
- User custom keywords
- Check for user notification preferences

Requires: https://github.com/mattermost/mattermost-plugin-msteams/pull/804

### DIFF
--- a/server/mentions.go
+++ b/server/mentions.go
@@ -4,672 +4,319 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
+	"net/url"
+	"regexp"
 	"strings"
-	"unicode"
-	"unicode/utf8"
 
+	"github.com/mattermost/mattermost-plugin-msteams/server/msteams"
 	"github.com/mattermost/mattermost/server/public/model"
-	"github.com/mattermost/mattermost/server/public/shared/markdown"
-	"github.com/mattermost/mattermost/server/public/shared/mlog"
+	"github.com/mattermost/mattermost/server/public/plugin"
 )
 
-const (
-	// Different types of mentions ordered by their priority from lowest to highest
-
-	// A placeholder that should never be used in practice
-	NoMention MentionType = iota
-
-	// The post is in a GM
-	GMMention
-
-	// The post is in a thread that the user has commented on
-	ThreadMention
-
-	// The post is a comment on a thread started by the user
-	CommentMention
-
-	// The post contains an at-channel, at-all, or at-here
-	ChannelMention
-
-	// The post is a DM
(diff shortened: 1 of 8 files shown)


server/mentions.go
```diff
@@ -0,0 +1,295 @@
...
+		switch mention {
+		case "@here":
+			fallthrough
+		case "@channel":
+			fallthrough
+		case "@all":
+
+		default:
+			user := p.isUserMention(strings.TrimPrefix(mention, "@"))
+			group := p.isGroupMention(strings.TrimPrefix(mention, "@"))
```
wiggin77, line 63: Since these calls hit the database, and a mention can be only one of user or group here, can we avoid the `isGroupMention` call when it is a user mention?

server/mentions.go
```diff
@@ -0,0 +1,295 @@
...
+	if channel.Type == model.ChannelTypeDirect || channel.Type == model.ChannelTypeGroup {
+		p.Notifications = append(p.Notifications, &UserNotification{
+			Trigger: post.Message,
+			Post:    post,
+			Channel: channel,
+		})
+	}
+
+	for _, notification := range p.Notifications {
+		p.PAPI.LogInfo("Processed mention", "notification", *notification)
```
wiggin77, line 83: Should this be debug?

wiggin77 (changes requested):
Good refactor.  2 minor requests.

server/mentions.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	if notification.Group != nil {
+		return p.sendGroupNotification(notification)
+	}
+
+	switch notification.Trigger {
+	case "@here":
+		return p.sendChannelNotification(notification, true)
+	case "@channel":
```
wiggin77, line 144: You can have multiple expressions in a single case.  `fallthrough` is not needed here.

wiggin77 (approved):
LGTM :+1:   One minor nit.

wiggin77:
@lieut-data can you have a look at this or reassign?  Would like to get notifications into community.

server/mentions.go
wiggin77: @fmartingr it needs to be `case "@channel", "@all":`

lieut-data (approved):

PR #809, opened 2025-03-28
Update .gitignore for Aider

wiggin77:
#### Summary
Add Aider files to .gitignore.

#### Ticket Link
NONE

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ server/manifest.go
 
 # notice
 .notice-work
+
+# Aider
+.aider*


fmartingr (approved):

.gitignore
```diff
@@ -14,3 +14,6 @@ server/manifest.go
 
 # notice
 .notice-work
+
+# Aider
+.aider*
```
lieut-data, line 19: 0/5, but it feels like a lot of effort to have to propagate all these things across our many repos. Would you consider adding the following to your `~/.gitconfig`:

```
[core]
	excludesfile = ~/.gitignore_global
```

and then adding this things there instead?

lieut-data (approved):

PR #811, opened 2025-03-28
Add noroute URL param for iframe which skips the routing logic

wiggin77:
#### Summary
We are creating a multi-tab app for internal use that has hub, community and corpus as tabs.  These tabs are across the top (see screenshot below).  This will require a manifest with three tabs defined, one of which will point at Community but the current routing rules will redirect that to hub.  This PR allows for a query string params that bypasses the routing and forces the redirect to stay on Community for any tenant.  

Screenshot from 2025-03-28 11-14-07

#### Ticket Link
NONE

### DIFF
--- a/server/iframe.html
+++ b/server/iframe.html
@@ -30,13 +30,6 @@
     const urlParams = new URLSearchParams(window.location.search);
     const noRoute = urlParams.has('noroute');
 
-    // bypass routing if `noroute` parameter is present in the URL
-    if (noRoute) {
-      console.log('noroute parameter is present, bypassing routing.');
-      iframe.src = '{{.SiteURL}}';
-      return;
-    }
-
     // Define a map of tenant IDs to their corresponding domain roots
     const tenantMap = {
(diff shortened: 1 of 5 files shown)


lieut-data (approved):

server/iframe.html
```diff
@@ -27,6 +27,16 @@
...
     });
 
+    const urlParams = new URLSearchParams(window.location.search);
+    const noRoute = urlParams.has('noroute');
+
+    // bypass routing if `noroute` parameter is present in the URL
+    if (noRoute) {
+      console.log('noroute parameter is present, bypassing routing.');
+      iframe.src = '{{.SiteURL}}';
+      return;
```
fmartingr, line 37: nit: How correct is this return considering is not inside a function context?

fmartingr (approved):

wiggin77:
@fmartingr I updated/simplified the logic to simply select the correct URL and try to authenticate.

PR #812, opened 2025-03-28
Disable sync job via plugin config

wiggin77:
#### Summary
Provides a way to disable the sync job thus avoiding error logs and unneeded Graph API calls.

To disable, set the `Enable sync job` to false via plugin config in the system console. 

#### Ticket Link
NONE

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -20,6 +20,13 @@
     "footer": "",
     "settings": [
       {
+        "key": "enableSyncJob",
+        "display_name": "Enable sync job",
+        "type": "bool",
+        "help_text": "When false, disables the sync job that creates and maintains chat and channel subscriptions",
+        "default": true
+      },
+      {
         "key": "tenantId",
         "display_name": "Tenant ID",
         "type": "text",
--- a/plugin.json
+++ b/plugin.json
@@ -20,11 +20,11 @@
     "footer": "",
     "settings": [
       {
-        "key": "enableSyncJob",
-        "display_name": "Enable sync job",
+        "key": "disableSyncJob",
+        "display_name": "Disable sync job",
         "type": "bool",
-        "help_text": "When false, disables the sync job that creates and maintains chat and channel subscriptions",
-        "default": true
+        "help_text": "When true, disables the sync job that creates and maintains chat and channel subscriptions",
+        "default": false
       },
       {
         "key": "tenantId",
--- a/server/monitor.go
+++ b/server/monitor.go
@@ -5,9 +5,7 @@ package main
 
 import (
 	"fmt"
-	"os"
 	"runtime/debug"
-	"strings"
 	"time"
 
 	"github.com/mattermost/mattermost-plugin-msteams/server/metrics"
@@ -56,11 +54,6 @@ func (m *Monitor) Start() error {
 	// Close the previous background job if exists.
 	m.Stop()
 
-	if !m.IsEnabled() {
-		m.api.LogInfo("Monitoring system job is disabled via env var.")
-		return nil
-	}
-
 	job, jobErr := cluster.Schedule(
 		m.api,
 		monitoringSystemJobName,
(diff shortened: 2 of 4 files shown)


fmartingr (approved):

server/monitor.go
```diff
@@ -78,6 +85,16 @@ func (m *Monitor) Stop() {
...
 
+func (m *Monitor) IsEnabled() bool {
+	// The job can be disabled via an environment variable
+	val := os.Getenv("MM_TEAMS_SYNC_MONITORING_SYSTEM_DISABLED")
+	val = strings.ToLower(strings.TrimSpace(val))
+	if val == "true" || val == "1" {
+		return false
+	}
+	return true
+}
```
lieut-data, line 96: Can we just add this as a proper configuration value? (This works fine for Cloud, but isn't an ideal setup if anyone else deploys.)

lieut-data (changes requested):

server/monitor.go
wiggin77: I can.  This was meant as a temporary nice-to-have for developers while we move this code into a new plugin.  I can't imagine anyone else deploying this plugin.

lieut-data: If we're not shipping this to production, maybe we can just live with the pain of the log spam for now? (Or are there other symptoms?)

wiggin77: I have made this a plugin config setting.

plugin.json
```diff
@@ -19,6 +19,13 @@
     "header": "",
     "footer": "",
     "settings": [
+      {
+        "key": "enableSyncJob",
+        "display_name": "Enable sync job",
+        "type": "bool",
+        "help_text": "When false, disables the sync job that creates and maintains chat and channel subscriptions",
+        "default": true
```
lieut-data, line 27: Could we make this match the `disableCheckCredentials` pattern, and default to `false` instead?

server/plugin.go
```diff
@@ -282,16 +282,20 @@ func (p *Plugin) start(isRestart bool) {
...
+	} else {
+		p.API.LogInfo("Sync job disabled via config")
 	}
 
 	ctx, stop := context.WithCancel(context.Background())
 	p.stopSubscriptions = stop
 	p.stopContext = ctx
 
-	if !p.getConfiguration().DisableCheckCredentials {
+	if !p.getConfiguration().DisableCheckCredentials && p.configuration.EnableSyncJob {
```
lieut-data, line 298: Why do we need two configurations to disable the credentials job?

lieut-data (changes requested):

server/plugin.go
wiggin77: There is no UI that I can find to `DisableCheckCredentials`   It is only set by unit test helpers.

plugin.json
wiggin77: The current behaviour is that by default the sync job is enabled.  I don't want to change the default least it mess up something's existing instance by turning it off after an upgrade.

lieut-data: Right, that's why I'm proposing `disableSyncJob` with a default of `false`. (We had to do this for `disableCheckCredentials` since I couldn't rely on the `plugin.json`'s defaults /actually/ taking effect.)

server/plugin.go
lieut-data: Indeed: should this setting be the same? Disabling this functionality isn't something someone should do for the production version of this plugin. Developers can manipulate this via config.json.

wiggin77: I don't think there is a production version of this plugin - at least not one where someone is using the sync feature.  In a few weeks we'll have the requisite code in a separate plugin and I don't think this one will be deployed anywhere.

plugin.json
wiggin77: The setting is flipped.

server/plugin.go
lieut-data: I'm actively using the one-way sync feature of this plugin on Hub. We've prepackaged the functionality and shipped it in the last ESR. And a customer at CKO mentioned using it (although they balked at the cost, so stopped). I can't say with confidence that no one is using this.

lieut-data (approved):

server/plugin.go
wiggin77: > Disabling this functionality isn't something someone should do for the production version of this plugin

This would have been a nicely hidden setting as a env var not meant for production use.  That was my original plan.

PR #815, opened 2025-04-01
fix: sso audience and other minor fixes

fmartingr:
#### Summary
- [fix: redirect iframe to site url if there's an error](https://github.com/mattermost/mattermost-plugin-msteams/commit/4d4b49da5cc02a8871500c14d7132cc2f12770af): We weren't catching errors on the `getContext` call, leaving the frame in `about:blank`.
- [fix: retrieve the correct property from the user](https://github.com/mattermost/mattermost-plugin-msteams/commit/a83e5e5fe37711df3fe31b25ddada58b9e90ccea): We were storing the object id as `oid` but trying to use it as `user_id`, breaking the login flow.
  - @wiggin77 you were hitting this, that's why SSO wasn't working on your app
- [feat: store user metadata in the plugin kvstore](https://github.com/mattermost/mattermost-plugin-msteams/pull/815/commits/46d3b7569c6e82a5f3816d508f66a43690d3db7f)

### DIFF
--- a/server/jwt.go
+++ b/server/jwt.go
@@ -5,7 +5,9 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"net/url"
 
 	"github.com/google/uuid"
 
@@ -17,7 +19,7 @@ import (
 
 const (
 	MicrosoftOnlineJWKSURL = "https://login.microsoftonline.com/common/discovery/v2.0/keys"
-	ExpectedAudience       = "api://community.mattermost.com/4ef56ea2-4a2f-4817-a6e0-a7cd760e2034"
+	ExpectedAudienceFmt    = "api://%s/%s"
 )
 
 type validationError struct {
@@ -43,7 +45,7 @@ func setupJWKSet() (keyfunc.Keyfunc, context.CancelFunc) {
 	return k, cancelCtx
 }
 
-func validateToken(jwtKeyFunc keyfunc.Keyfunc, token string, expectedTenantIDs []string, enableDeveloper bool) (jwt.MapClaims, *validationError) {
+func validateToken(jwtKeyFunc keyfunc.Keyfunc, token string, expectedTenantIDs []string, enableDeveloper bool, siteURL, clientID string) (jwt.MapClaims, *validationError) {
 	if token == "" && enableDeveloper {
 		logrus.Warn("Skipping token validation check for empty token since developer mode enabled")
 		return nil, nil
@@ -82,11 +84,20 @@ func validateToken(jwtKeyFunc keyfunc.Keyfunc, token string, expectedTenantIDs [
 		// There's no WithNotBefore() helper, but the library always verifies if the claim is present.
 	}
 
+	mmServerURL, err := url.Parse(siteURL)
+	if err != nil {
+		logrus.WithError(err).WithField("site_url", siteURL).Warn("Failed to parse site URL, check your system console")
+		return nil, &validationError{
+			StatusCode: http.StatusUnauthorized,
+			Message:    "Failed to authenticate due to Mattermost server missconfiguration. Contact your system administrator.",
+		}
+	}
+
 	// Verify that this token was signed for the expected app, unless developer mode is enabled.
 	if enableDeveloper {
 		logrus.Warn("Skipping aud claim check for token since developer mode enabled")
 	} else {
-		options = append(options, jwt.WithAudience(ExpectedAudience))
+		options = append(options, jwt.WithAudience(fmt.Sprintf(ExpectedAudienceFmt, mmServerURL.Host, clientID)))
 	}
 
 	parsed, err := jwt.Parse(
@@ -161,6 +172,7 @@ func validateToken(jwtKeyFunc keyfunc.Keyfunc, token string, expectedTenantIDs [
 	for _, expectedTenantID := range expectedTenantIDs {
 		if claims["tid"] == expectedTenantID {
 			logger.Info("Validated token, and authorized request from matching tenant")
+			logger.Info("Claim tenant id: ", claims["tid"])
 			return claims, nil
 		} else if enableDeveloper && expectedTenantID == "*" {
 			logger.Warn("Validated token, but authorized request from wildcard tenant since developer mode enabled")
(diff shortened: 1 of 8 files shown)


wiggin77 (approved):
LGTM.  Non-blocking comment...  can we store the app id and object id in the plugin KV store or plugin config instead of in each user's properties?  I think we should do this sooner than later so we don't pollute user records with unnecessary props.

Also, what is the value of non validating auth tokens when in dev mode?

server/jwt.go
```diff
@@ -17,7 +19,7 @@ import (
 
 const (
 	MicrosoftOnlineJWKSURL = "https://login.microsoftonline.com/common/discovery/v2.0/keys"
-	ExpectedAudience       = "api://community.mattermost.com/4ef56ea2-4a2f-4817-a6e0-a7cd760e2034"
```
lieut-data, line 20: > The audience validation was fixed to community, let's format the URI so it can match wathever instance this plugin is installed into.

This is by design, as in a production deployment we expect this to match the Teams app we've deployed and no other. There is a case for it being paratemerizable, but only for customers who deploy this plugin and want to use the app they've deployed for their org. When we support that flow, we can make this whole thing an explicit configuration, but for now it has to be constant.

(Note that in developer mode, this audience check is skipped, so it's still possible to iterate without running the full suite of checks.)

server/jwt.go
```diff
@@ -161,6 +172,7 @@ func validateToken(jwtKeyFunc keyfunc.Keyfunc, token string, expectedTenantIDs [
 	for _, expectedTenantID := range expectedTenantIDs {
 		if claims["tid"] == expectedTenantID {
 			logger.Info("Validated token, and authorized request from matching tenant")
+			logger.Info("Claim tenant id: ", claims["tid"])
```
lieut-data, line 175: Can we just add the `tid` to the logger context early on so all logs end up with this being annotated?

lieut-data (changes requested):

server/jwt.go
fmartingr: Yeah, the `tid` is already in the logger context, this may have slipped while debugging.

server/jwt.go
fmartingr: But for the deployed community plugin, this value will match the configuration, so it doesn't need to be fixed in the code, right?

fmartingr:
> LGTM. Non-blocking comment... can we store the app id and object id in the plugin KV store or plugin config instead of in each user's properties? I think we should do this sooner than later so we don't pollute user records with unnecessary props.

https://github.com/mattermost/mattermost-plugin-msteams/pull/815/commits/2e11ec59db0811225f2b85f9f782fba46de0dc5e

> Also, what is the value of not validating auth tokens when in dev mode?

I suppose faster iteration locally? cc: @lieut-data

server/jwt.go
lieut-data: No, community is just passing the JWT to the downstream server which runs this code, so the SiteURL and ClientId wouldn't match. This has to be a fixed constant.

lieut-data:
> Also, what is the value of not validating auth tokens when in dev mode?

For the Playbooks product, this allows the webpage to load "outside" of Teams, when there is no JWT available. Not sure if that's all that interesting for this app.

wiggin77 (approved):

server/jwt.go
fmartingr: Then maybe a setting in the plugin to enable this "routing behaviour" and check against that, but allow this for people that want to install their own manifest? cc: @wiggin77

wiggin77: The [devsecops](https://github.com/mattermost/mattermost-plugin-msteams-devsecops) variant of this plugin will handle the non-routing customers, and help customers generate a manifest that can be installed. 

The [msteams](https://github.com/mattermost/mattermost-plugin-msteams) variant will (temporarily) handle the customers that need routing.  We will move the routing code to something like an Azure Function (lambda) once we have the mechanics nailed down.

fmartingr: 👍 

[feat: store user metadata in the plugin kvstore](https://github.com/mattermost/mattermost-plugin-msteams/pull/815/commits/46d3b7569c6e82a5f3816d508f66a43690d3db7f)

fmartingr:
Hey @lieut-data, can you take a look at this? I reverted the audience change for this one.

server/mentions.go
```diff
@@ -241,10 +244,10 @@ func (p *NotificationsParser) sendChannelNotification(un *UserNotification, onli
 }
 
 func (p *NotificationsParser) sendUserActivity(userActivity *UserActivity) error {
-	sender, err := p.PAPI.GetUser(userActivity.UserNotification.Post.UserId)
-	if err != nil {
-		p.PAPI.LogError("Failed to get sender", "error", err.Error())
-		return err
+	sender, appErr := p.PAPI.GetUser(userActivity.UserNotification.Post.UserId)
```
lieut-data, line 247: nit: can use the plugin api wrapper exposed on `p` to get real `error` values and avoid the `appErr` juggling.

server/store/pluginstore/store.go
```diff
@@ -0,0 +1,98 @@
...
+
+func NewUser(mattermostUserID, teamsObjectID, teamsSSOUsername string) *User {
+	return &User{
+		MattermostUserID: mattermostUserID,
+		TeamsObjectID:    teamsObjectID,
+		TeamsSSOUsername: teamsSSOUsername,
+	}
+}
+
+type Store interface {
```
lieut-data, line 27: Do we need a `Store` for this? The plugin API wrapper already has a `Get(key string, o any)` interface that does the JSON unmarshalling (similar for `Set`).

lieut-data (approved):

server/store/pluginstore/store.go
fmartingr: I made it a separate file so it was easier to move to the devsecops application, the other store seemed more focused into direct SQL storage.

lieut-data: Yeah, totally fine to keep them separate, I mean just use the pluginapi directly without any abstraction layer, but 0/5 given this is moving.

PR #826, opened 2025-04-04
feat: check set up audience if noroute is set

fmartingr:
#### Summary
Use the `noroute` param in the backend to determine if we need to validate the audience against community or to the set up parameters in the plugin configuration.

### DIFF
--- a/server/jwt.go
+++ b/server/jwt.go
@@ -5,7 +5,9 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"net/url"
 
 	"github.com/google/uuid"
 
@@ -18,6 +20,7 @@ import (
 const (
 	MicrosoftOnlineJWKSURL = "https://login.microsoftonline.com/common/discovery/v2.0/keys"
 	ExpectedAudience       = "api://community.mattermost.com/4ef56ea2-4a2f-4817-a6e0-a7cd760e2034"
+	ExpectedAudienceFmt    = "api://%s/%s"
 )
 
 type validationError struct {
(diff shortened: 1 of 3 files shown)


server/jwt.go
```diff
@@ -82,9 +85,22 @@ func validateToken(jwtKeyFunc keyfunc.Keyfunc, token string, expectedTenantIDs [
 		// There's no WithNotBefore() helper, but the library always verifies if the claim is present.
 	}
 
+	mmServerURL, err := url.Parse(siteURL)
```
wiggin77, line 88: I don't think we can add a param to the actual site URL of the MM instance.  I was thinking this code path would be called in response to a HTTP request to fetch the iFrame and the request would have the param.  The request param would be passed here via context or a flag added to this method that determines if community audience should be used.

wiggin77: I'll update the PR.

wiggin77:
@fmartingr I updated the PR to use the iFrame query string to determine if `noroute` is present.   PHAL.

lieut-data (changes requested):
Can we schedule a call re: the dynamic audience? My understanding back from the security review for Playbooks was that this had to match the static registration we published to the Teams App Store.

I'd love to get up to speed on what I'm misunderstanding that makes this change necessary, and help avoid any breach.

fmartingr:
> Can we schedule a call re: the dynamic audience? My understanding back from the security review for Playbooks was that this had to match the static registration we published to the Teams App Store.
> 
> I'd love to get up to speed on what I'm misunderstanding that makes this change necessary, and help avoid any breach.

In order to match not only the community msteams tab application but any other apps we generate from this plugin we need to be able to tell the plugin to validate against the community msteams plugin (when we are relying on routing) and to the final msteams plugin (installed on destination mattermost instances). Happy to hop on a call to call to discuss about this.

lieut-data:
> In order to match not only the community msteams tab application but any other apps we generate from this plugin we need to be able to tell the plugin to validate against the community msteams plugin (when we are relying on routing) and to the final msteams plugin (installed on destination mattermost instances). Happy to hop on a call to call to discuss about this.

Yep, that makes sense. In that case, this needs to be an explicit `ExpectedAudience` configuration setting on the plugin, and shouldn't try to copy from `SiteURL` or `ClientID` directly.

lieut-data (approved):
Let's get this merged -- reviewed synchronously :)

PR #828, opened 2025-04-07
MM-63675: Fix nil pointer dereference in GetFileSizeAndDownloadURL

lieut-data:
## Summary
- Fixed nil pointer dereference in GetFileSizeAndDownloadURL when processing file attachments
- Added proper error handling for nil additionalData
- Added safer type assertion for downloadURL that handles both string and *string types 

## Jira Ticket
https://mattermost.atlassian.net/browse/MM-63675

## Related PRs
N/A


agnivade (approved):
Not sure if it's possible to add a unit test for this?

PR #835, opened 2025-04-19
Remove hard-coded app id.

wiggin77:
#### Summary
Removes the hard-coded app id that was added while trouble-shooting notifications.

#### Ticket Link
NONE

### DIFF
--- a/server/store/pluginstore/store.go
+++ b/server/store/pluginstore/store.go
@@ -81,18 +81,16 @@ func (s *PluginStore) StoreAppID(appID string) error {
 }
 
 func (s *PluginStore) GetAppID() (string, error) {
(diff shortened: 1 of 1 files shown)


fmartingr (approved):

lieut-data (approved):

lieut-data:
@wiggin77 @ @fmartingr, is there a plan to address the stomping and "global variable"? (Even without multiple apps in play, this remains an issue.)

wiggin77:
> @wiggin77 @ @fmartingr, is there a plan to address the stomping and "global variable"? (Even without multiple apps in play, this remains an issue.)

@lieut-data Yes, I will address that on Monday.

PR #868, opened 2025-07-04
https://mattermost.atlassian.net/browse/MM-64369

agnivade:


server/api.go
```diff
@@ -268,7 +268,15 @@ func (a *API) connect(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	state := fmt.Sprintf("%s_%s_%s", model.NewId(), userID, stateSuffix)
+	stateKey := model.NewId()
+	stateMetadataKey := fmt.Sprintf("oauth_state_%s_%s", userID, stateKey)
+	if appErr := a.p.API.KVSet(stateMetadataKey, []byte(stateSuffix)); appErr != nil {
+		a.p.API.LogWarn("Error in storing the OAuth state metadata", "error", appErr.Message)
+		http.Error(w, "Error in trying to connect the account, please try again.", http.StatusInternalServerError)
+		return
+	}
```
lieut-data, line 277: Can we set with expiry so that we don't accumulate these keys indefinitely if they don't get used correctly? (24hr ought to suffice)

lieut-data (approved):
Looks good, just one thought below.

